PR #86, opened 2014-02-15
add better android app launch

atomicptr:
A small change: I've "improved" (removed the Actionbar) the look of the app while launching (example below).

So the app launch now looks somewhat more like this:

Some further improvements would be the ability to create an own theme which could be used for a splash screen or like above to color the whole screen red ;).

Also done some refactoring on the manifest (now it looks pretty!)

### DIFF
--- a/templates/android/template/AndroidManifest.xml
+++ b/templates/android/template/AndroidManifest.xml
@@ -1,26 +1,56 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="::APP_PACKAGE::" android:versionCode="::APP_BUILD_NUMBER::" android:versionName="::APP_VERSION::" android:installLocation="::ANDROID_INSTALL_LOCATION::">
-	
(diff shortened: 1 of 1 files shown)


restorer:
For anyone who want to make proper fix:

You should not use `android:theme="@android:style/Theme.Holo.NoActionBar"` inside `AndroidManifest.xml`.
Use you own custom theme instead - `android:theme="@style/AppTheme"` (`AppTheme` is just for example, actually you can use any other name).

Than create `res/values/styles.xml` (for pre-HONEYCOMB):
```xml
<resources xmlns:android="http://schemas.android.com/apk/res/android">
    <style name="AppTheme" parent="@android:style/Theme.NoTitleBar.Fullscreen" />
</resources>
```
And  `res/values-v11/` (for HONEYCOMB and above):
```xml
<resources xmlns:android="http://schemas.android.com/apk/res/android">
    <style name="AppTheme" parent="@android:style/Theme.Holo.NoActionBar" />
</resources>
```

There is also other features that can be tweaked by custom app theme (use no background for speedup, for example), but it is out of scope of this PR.

Issue #300, opened 2014-12-31
haxedef HXCPP_STACK_* has no effect on iOS release build


cbatson:
On an iOS release build, these have no effect:

``` xml
<haxedef name="HXCPP_STACK_TRACE" />
<haxedef name="HXCPP_STACK_LINE" />
```

The definitions never make it to the compiler command line. However, I've confirmed that adding this to my Main class does result in stack traces on a release build:

``` haxe
@:buildXml("
	<files id='haxe'>
		<compilerflag value='-DHXCPP_STACK_LINE'/>
	</files>
	<files id='runtime'>
		<compilerflag value='-DHXCPP_STACK_LINE'/>
	</files>
")
```

I'll have a pull request for this shortly.

PR #301, opened 2014-12-31
for iOS, pass HXCPP_STACK_TRACE and HXCPP_STACK_LINE to hxcpp

cbatson:
Resolves https://github.com/openfl/lime/issues/300.

### DIFF
--- a/legacy/templates/iphone/PROJ/haxe/makefile
+++ b/legacy/templates/iphone/PROJ/haxe/makefile
@@ -39,7 +39,7 @@ ifeq ("$(HAXE_OS)","iphonesimulator")
 endif
 CONFIG := $(CONFIG)-$(HAXE_OS)
 
-HXCPP_DEFS := ::if DEFINE_HXCPP_STACK_TRACE::-DHXCPP_STACK_TRACE::end:: ::if DEFINE_HXCPP_STACK_LINE::-DHXCPP_STACK_LINE::end::
+HXCPP_DEFS := ::foreach HAXE_DEFS:: -D::__current__::::end::
 
 debug_print:
 	@echo "Make $(HAXE_BUILDS)"
--- a/legacy/templates/iphone/PROJ/haxe/makefile
+++ b/legacy/templates/iphone/PROJ/haxe/makefile
@@ -39,6 +39,8 @@ ifeq ("$(HAXE_OS)","iphonesimulator")
 endif
 CONFIG := $(CONFIG)-$(HAXE_OS)
 
+HXCPP_DEFS := ::if DEFINE_HXCPP_STACK_TRACE::-DHXCPP_STACK_TRACE::end:: ::if DEFINE_HXCPP_STACK_LINE::-DHXCPP_STACK_LINE::end::
+
 debug_print:
 	@echo "Make $(HAXE_BUILDS)"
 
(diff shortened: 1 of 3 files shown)


cbatson:
When done for all haxedefs, these additional flags are passed to hxcpp (as an example):

```
HXCPP_DEFS :=  -Dtools=2.0.3 -Dlime-legacy -Dno-compilation -DHXCPP_STACK_LINE -Dnative -Dopenfl-native -Dlime-native
```

It's not clear to me what, if any, effect the additional flags might have on the hxcpp build. In a quick trial with my local project, there was no immediately apparent ill effect.

If you think the extra flags are ok, we can go with the generalized approach.

jgranick:
Well, I'm just thinking, we pass all haxedef values to HXCPP, I know this affects can affect the C++ cache, needlessly, but I'm not sure we can assume what defines will affect the HXCPP build, and which will not. At least, if we did make this assumption, we would definitely want to do it higher up in the Lime tools, not at a template level. I'm just thinking, the next time someone adds a C++ define in the toolchain XML, we'll have the same problem if we don't do something more general

jgranick:
I could pull as-is and it would be a definite step forward, I'm just trying to think ahead. Up to you :)

cbatson:
I don't feel quite qualified to make a call about a generalized approach, because I have low familiarity with what's happening under the hood.

A 3rd option, which I haven't explored yet, is to forward only those haxedefs that are explicitly defined in the application.xml. That might be a good middle path.  But it will add a little extra bookkeeping to HXProject specifically for this purpose.

So the options before us are:

1. Pull request as it is; a baby step of forwarding the 2 specific flags, and figure out something more general later.

2. Forward **all** haxedefs are described in the earlier comment. I already have this implemented and can commit an update to the pull request.

3. Forward only those haxedefs that appear in the application.xml. I *think* it shouldn't be too hard to do this...

Personally, I will be happy with any of these.  :-)

cbatson:
In other words, the reason I started with specifically the 2 flags instead of generally forwarding all haxedefs, is because it was the most targeted change I could make with confidence I wouldn't inadvertently introduce unexpected issues.

Forwarding all haxedefs is completely fine with me -- just that at my personal level of familiarity at this time with the build process I do not have the same level of confidence that issues won't be introduced as a result of the extra flags.

Hope that made sense.  :-)

jgranick:
When you use Haxe to compile for C++, usually, it calls HXCPP automatically. HXCPP knows what Haxe defines you used by an "Options.txt" file that Haxe leaves in the same directory as the Build.xml file. In the Lime tools, I manually call HXCPP instead of allowing Haxe to call it automatically (giving us some further control over it), but read in those same options:

https://github.com/openfl/lime/blob/master/tools/helpers/CPPHelper.hx#L20

Last I checked, I think it sends all the -D defines into that file, so -Dhello -Dworld as well as -DHXCPP_STACK_TRACE. To be honest, I thought iOS was doing this already, like the other targets do. For now, I think that's the most consistent fix, but later, I think we might consider omitting some defines when calling HXCPP, so as to try and preserve the C++ code cache more often :)

cbatson:
I think iOS is a bit different because of the use of a makefile -- to do things in an Xcode-acceptable way.

Here's what my Options.txt looks like:

```
 -DHXCPP_STACK_LINE="1" -Dactuate="1.7.5" -Dhaxe3="1" -Dhaxe_ver="3.2" -Dhxcpp="3.1.0" -Dhxcpp_api_level="312" -Dios="1" -Diphone="1" -Dlime-legacy="1" -Dlime-native="1" -Dlime_legacy="1" -Dlime_native="1" -Dmobile="1" -Dnative="1" -Dno-compilation="1" -Dno_compilation="1" -Dopenfl="2.1.8" -Dopenfl-native="1" -Dopenfl_native="1" -Dsimulator="1" -Dtools="2.0.3"
/Users/chuck/fundo/projects/repos/haxe/hxcpp/
```

Here's what I get in the makefile after forwarding all haxedefs:

HXCPP_DEFS :=  -Dtools=2.0.3 -Dlime-legacy -Dno-compilation -DHXCPP_STACK_LINE -Dnative -Dopenfl-native -Dlime-native

There isn't a perfect 1-to-1 correlation:
 - the default value is not "1"
 - there are no underscore variants for defs with hyphens
 - some are just missing

Will push the updated commit shortly; at least it's a start.  =^D

jgranick:
Thank you! I just used your code to make a change that's a little less far-reaching -- specific to the iOS target. I believe this should also resolve the underscore, the quotes and other differences, hopefully (I think?) this should be consistent with other targets now. Thanks again for noticing this, I'm not sure how I never saw that the iOS builds were not quite the same in this regard

cbatson:
Ah! Good move putting it into IOSPlatform.hx, that's a better place.  I'm not very familiar with the build code, but you have helped me learn something today.  :-)

Issue #270, opened 2014-11-22
getPixels() fills the blue channel of non transparent images with 255 (Neko/mac/android)


mbalestrini:
After calling BitmapData.getPixels() of an image with no alpha channel, that area of the image gets modified. Every pixel get a blue value of 255

Here's the code to reproduce the bug:
```
var sourceBmd = Assets.getBitmapData("assets/someimage2.jpg");
sourceBmd.getPixels(new Rectangle(50,50,100,100));
var anotherBitmapData = new BitmapData(sourceBmd.width, sourceBmd.height, true, 0);
var bmp = new Bitmap(anotherBitmapData);
addChild(bmp);
anotherBitmapData.draw(sourceBmd);
```

I tried to find the place where that error could happen. I don't get all the complex structure of lime/openfl/nme, but after a while of surfing through the different sources/libraries, I think I found it.
The problem seems to be here in /legacy/project/src/common/Surface.cpp

```
void SimpleSurface::getPixels(const Rect &inRect,uint32 *outPixels,bool inIgnoreOrder, bool inLittleEndian)
```
...
```
if (!(mPixelFormat & pfHasAlpha))
 {
    for(int x=0;x<r.w;x++)
    {
       *a = 255;
       a+=4;
    }
 }
```

Hope it helps

gunnbr:
Sure enough--that looks wrong to me. The only thing I can think of is that someone was using that as a temporary test. I can't see when it got added though.

Taking that out does fix the problem. I also added a bunch of unit tests for other cases and am trying to get them to work on all targets before I can get the fix checked in though.

And note that it's actually pretty easy to get lime and openfl development environments setup, so you can try out the fix yourself as well!

mbalestrini:
I didn't want to change this particular code because I don't know how lime handles image formats. I could've fixed this particular issue, but without knowing if I was doing it the wrong way.
I thought it may have something to do with some change in the order of the ARGB channels, but I have no idea when or why they did it.

gunnbr:
Okay--the fix is in. After the next release comes out, give it a try!

PR #306, opened 2015-01-04
Fixed BitmapData.getPixels() to not change the image to turn it blue.

gunnbr:
This only happens when the BitmapData does not have an alpha component, such as from a loaded JPG.

As far as I can tell, this code should never have been here as it causes getPixels() to modify the BitmapData its getting the pixels from. It looks like either:

1. Someone was using it for testing to visually verify it was grabbing pixels from where they expected.
2. It was trying to get the alpha value in the returned pixels to 0xFF, but doing it wrong. And, that's not even necessary because without this, it already returns an alpha of 0xFF.

I also have a bunch of unit tests to go along with this change, but I can't submit those yet because they don't work on -next as that appears to use RGBA instead of ARGB like legacy does.

### DIFF
--- a/legacy/project/src/common/Surface.cpp
+++ b/legacy/project/src/common/Surface.cpp
@@ -1145,15 +1145,6 @@ void SimpleSurface::getPixels(const Rect &inRect,uint32 *outPixels,bool inIgnore
             src+=4;
          }
          outPixels += r.w;
-
-         if (!(mPixelFormat & pfHasAlpha))
-         {
-            for(int x=0;x<r.w;x++)
-            {
-               *a = 255;
-               a+=4;
-            }
-         }
       }
    }
 }


jgranick:
The latest release should have switch `getPixels` on Lime to use ARGB instead of RGBA, let me know if this does not seem to be the case

gunnbr:
Yeah, it works. Thought I had the latest, but I didn't. I do still see another difference between legacy and next, but I'm gathering information and will post something about that shortly.

jgranick:
I am not sure if I recall where it crept -- it could have been in the image loading code (which might be fixed now, as we use newer lib versions) or it might have been something internal that set surface pixels (such as `bitmapData.draw`) or some combination of the `getPixel`/`setPixel` APIs, I forget

I recall a problem where an image that was supposed to not have alpha was returning alpha values other than 0xFF, for example, if you created a new BitmapData and set the `transparent` value to false, but then used `setPixels` with alpha values, this would be stored, and returned when you called `getPixels`

The masking could be done on the set call instead of the get, but perhaps get was the safest way to ensure it never, ever had this data when it was not supposed to

gunnbr:
Okay. I just put up a pull request for some unit tests for all this. They don't currently test the case where you write different alpha values to a BitmapData with the `transparent` value set to false. I can add that in really quick though.

gunnbr:
Oh, and I just realized I didn't say above, but this pull request fixes issue #270.

PR #309, opened 2015-01-07
[legacy] Decomposing and recomposing of Matrix like AS3

dimanux:


jgranick:
I was meaning to try and find a good test case, so that I could be sure that this was an improvement, and would not cause any regressions. A change at this level makes me feel scared that we'll take a step backward regarding functionality that I knew already worked, but I think that Matrix adjustments could very well be the right thing for making things better. If you had a test case that enabled the "before" and "after" of this fix, that would be super helpful as well

dimanux:
Hi Joshua,
Sorry, it was closed because of my brach managment.
I opened another pull request (https://github.com/openfl/lime/pull/315). It has a reference to test case of this issue (https://github.com/dimanux/MatrixTest).

PR #311, opened 2015-01-09
Adds all launch image and icon sizes wanted by XCode

grapefrukt:
This fixes https://github.com/openfl/openfl/issues/463 and as a bonus enables apps to run at the native resolution on iPhone 6 and iPhone 6 Plus, as proper launch images are required for that.

### DIFF
--- a/legacy/templates/iphone/PROJ/Images.xcassets/AppIcon.appiconset/Contents.json
+++ b/legacy/templates/iphone/PROJ/Images.xcassets/AppIcon.appiconset/Contents.json
@@ -1,9 +1,22 @@
 {
   "images" : [
     {
+      "size" : "29x29",
       "idiom" : "iphone",
-      "scale" : "1x",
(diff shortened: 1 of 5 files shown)


grapefrukt:
After fiddling a bit more with this I noticed that XCode will display a warning unless you're explicitly targeting iOS 5 or above, I can't imagine this being a problem?

jgranick:
I'm pretty sure we don't work for a version below iOS 5, so perhaps we should go ahead and do something to get rid of the warning? I don't want to lock it into being 5, though, if you're targeting something newer

grapefrukt:
You can set the deployment version via the nmml, so it's in there already:
```
<ios deployment="6.0" />
```
As far as I can tell Lime defaults to 5 if nothing else is set, but I don't think it's properly output in the XCode project or something. Either way, I'll try and change that default to 6 and see how that goes.

PR #330, opened 2015-01-24
[legacy] Set TextField's caret color to match textColor.

LaughingLeader:
Input text with a dark background and light text color should be more readable now.

The "text cursor" that appears when hovering over selectable text needs to reflect this next (the cursor currently becomes black).

This is in the Windows build.

### DIFF
--- a/legacy/project/src/common/TextField.cpp
+++ b/legacy/project/src/common/TextField.cpp
@@ -1512,7 +1512,7 @@ void TextField::Render( const RenderTarget &inTarget, const RenderState &inState
             double height = mLines[line].mMetrics.height;
             if (pos.y+height <= fieldHeight-GAP+1)
             {
-               mCaretGfx->lineStyle(1, 0x000000 ,1.0, false, ssmOpenGL  );
+               mCaretGfx->lineStyle(1, textColor ,1.0, false, ssmOpenGL  );
                mCaretGfx->moveTo(pos.x,pos.y);
                mCaretGfx->lineTo(pos.x,pos.y+height);
             }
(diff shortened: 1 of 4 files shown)


ubald:
Maybe this could be merged as a default value into https://github.com/openfl/lime/pull/320 it already adds support to customize caret and selection color.

LaughingLeader:
Aha! Custom caret colors are even better. 

The "TextCursor" cursor color will definitely need to be addressed, as currently it defaults to black, making it hard to see with certain text/layouts colors.

ubald:
yeah that one I ended up just doing my own openfl subclass, hiding the cursor and replacing it with one of my own as if I recall correctly it uses SDL cursors which are either white or black.

LaughingLeader:
Nice. That's a good idea. I'll have to do that in the meantime.

eminfedar:
Hey, anyone here? openfl.text.textField's caret isn't change his color when I set textColor to 0xFFFFFF.

Any solution?

legacy/project/src/common/TextField.cpp
```diff
@@ -1512,7 +1517,11 @@ void TextField::Render( const RenderTarget &inTarget, const RenderState &inState
             double height = mLines[line].mMetrics.height;
             if (pos.y+height <= fieldHeight-GAP+1)
             {
-               mCaretGfx->lineStyle(1, 0x000000 ,1.0, false, ssmOpenGL  );
+				if(caretIndex == 0 && (mCharGroups.empty() || (mCharGroups.size() == 1 && mCharGroups[0]->Chars() == 0)) && getTextFormat()->align == tfaCenter)
```
Gama11: Looks like you're using tabs instead of spaces here.

PR #332, opened 2015-01-26
Update extension template Build.xml

DanielUranga:
Previous to this commit the extension template project does not puts the correct extension to generated file (.dylib instead of .ndll).
I left the previous "mac" in order not to break something that maybe was working with previous versions of Hxcpp (Im using Hxcpp git).


jgranick:
Was this caused by not using `lime rebuild <path to or haxelib name of extension> mac`?

DanielUranga:
Im using "haxelib run hxcpp Build.xml -DHXCPP_M64" inside the project directory to build

DanielUranga:
Well, "lime rebuild <path to or haxelib name of extension> mac" works ok, so I will close this.

PR #334, opened 2015-01-27
Adding text input event

MattTuttle:
This change uses the SDL_TEXTINPUT event to capture text input from the user. It's useful for text fields and should support characters that require multiple key presses.

Thoughts:
* Should this accumulate a text string and output it or just return individual characters?
* Flash requires a dummy text field to work properly. Is forcing focus the right way to handle text input?


jgranick:
Thanks again, as I think you know, this has already been integrated in :smile:

PR #342, opened 2015-02-16
Added fallback to filename when font name couldn't be parsed

kiroukou:
Just added a simple fallback in case the font name could not be parsed properly (we have such font here, homemade).
Haven't been able to debug the lime code directly, so meanwhile I though that fallback could be helpful


kiroukou:
The same thing is required on HxProject ! 

I just discovered that later today...

PR #347, opened 2015-02-23
Support key modifier on HTML5

misohena:


jgranick:
Thank you, I really appreciate your help. This inspired me to improve the "KeyModifier" type further -- I made it abstract (so you can use regular integers or the static types) and also added boolean values to make it easy to check `modifier.altKey` and `modifier.ctrlKey` (etc) as well :smile:

I also found that you can override "onKeyDown" or "onKeyUp" using either the abstract KeyMode/KeyModifier types or just using Int, whichever is easiest for people, which is a neat bonus :smile:

misohena:
Thank you for improving the KeyModifier type.
It's very ease of use!
I have to study more Haxe.

misohena:
public static inline var ALT:KeyModifier = (0x0100 | 0x0200);
private inline function set_altKey (value:Bool):Bool {
 this = value ? (this | ALT) : (this & 0xFFFFFFF - (1 << ALT));
 return value;
}

It looks like this part is not correct > (this & 0xFFFFFFF - (1 << ALT))

jgranick:
Thanks, you're right, it should have been `this & (0xFFFFFFF - ALT)` in order to work properly, but I changed it to `this &= 0xFFFFFFF - ALT` to maybe make a little more sense.

Abstracts are a fairly new feature to Haxe, but quite powerful -- you can use this one just like an Int, but you get the added benefits of a "smart" type. You can do it over Strings and any other type, too. It's fun :smile:

PR #379, opened 2015-03-22
`cvs` removed from built-in exclude filter

eliasku:
I have a lot of files with `cvs` sequence inside file name. For example `docvslayer.png`, abbreviation `dcvs_click.mp3`, `cvsync.atlas`, `cvs_log.json`, etc. Now these files is ignored and not copying into build folders, I maintain these specific situations by gradle task now, but this is just workaround solution.

So I propose to remove CVS exclude filter from build-in exclude list. I think it would be much better to set `cvs` exclude filter only by hands in xml project file.

### DIFF
--- a/lime/project/HXProject.hx
+++ b/lime/project/HXProject.hx
@@ -556,7 +556,7 @@ class HXProject {
 			
 		}
 		
-		exclude = exclude.concat ([ ".*", "cvs", "thumbs.db", "desktop.ini", "*.hash" ]);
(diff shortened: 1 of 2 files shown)


jgranick:
Hmm, I think the real issue is that filter is not supposed to act like a wildcard -- it is supposed to match exactly "cvs" and not "*cvs*", so I think we'll have other unexpected behavior if that is not resolved :frowning:

jgranick:
Since the wildcard filter was fixed in the Lime tools, I think we should be safe here. Thanks again for your help :smile:

PR #384, opened 2015-03-25
Fixed incorrect alignment behavior.

player-03:
I know it's only legacy code, but it's still obviously incorrect.


player-03:
Wait, never mind, all the other alignments (besides the default) are broken too. This won't fix anything.

PR #386, opened 2015-03-26
Adding mouse functionality for locking/warping in native

sepharoth213:
This pull request contains 3 things:
 - `lime.Mouse.lock`, a boolean that hides the cursor and prevents it from leaving the window. AFAIK there isn't a way to keep the cursor from being hidden using `SDL_SetRelativeMouseMode`
- `lime.Mouse.globalWarp`, a function that warps the cursor anywhere in screen space
- `window.warpMouse`, a function that warps the cursor anywhere in window space

Not having relative mouse motion events is mitigated by having window.warpMouse, as users can log the position of the mouse after it moves and then warp it back to the center. It is worth noting, however, that both of the warp functions will fire a `mouseMove` event.

### DIFF
--- a/lime/_backend/flash/FlashMouse.hx
+++ b/lime/_backend/flash/FlashMouse.hx
@@ -38,6 +38,10 @@ class FlashMouse {
 		
 	}
 	
+	public static function setRelative (value:Bool):Int{
+
+		return -1;
+	}
 	
 	
 	
--- a/lime/_backend/flash/FlashMouse.hx
+++ b/lime/_backend/flash/FlashMouse.hx
@@ -38,16 +38,27 @@ class FlashMouse {
 		
 	}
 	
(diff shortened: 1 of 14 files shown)


jgranick:
Thank you :smile:

I made some modifications to help integrate this in, and few fixes. I am still using `Mouse.lock` but I combined both of the warp methods into `Mouse.warp`

`Mouse.warp` has an optional third argument for a window, so it can be used for global or window-relative warping, where supported.

I decided to add `window.onMouseMoveRelative` for relative mouse events. I'm still debating it a little, but for now, it should provide a clean (but separate) path for relative mouse lock events

sepharoth213:
Brilliant!
I looked the changes and I really like them all. I do agree that at some point the button argument should be dropped from `onMouseMove`.

PR #387, opened 2015-03-26
Fix for lime deploy (zip stuff)

larsiusprime:


jgranick:
Just pulled it in. Thanks!

PR #388, opened 2015-03-26
fixing segfault on Gamepad.name and Gamepad.guid

whilesoftware:
This issue wasn't obvious until I started testing disconnection/reconnection and multiple gamepads. The code seemed to work without modification with only one gamepad connected, but that was just luck.


jgranick:
Thanks! Just added :smile:

PR #389, opened 2015-03-27
Add help info for -web flag

T1mL3arn:
And additional questions - for what purpose exists opera and chrome templates in the templates directory? These are used with flags ```-opera``` and ```-chrome``` ? I tried to do ```lime test flash -opera``` but the result is steel the same as just ```lime test flash```.


ibilon:
Thanks.

Looks like the opera and chrome flags were removed.

PR #414, opened 2015-04-22
BitmapData: added getColorBoundsRect implementation for next

larsiusprime:
I've tested this on a wide range of bitmap images (including randomly generated ones) and it is confirmed accurate. I tried a few more "optimized" versions than this but the gains were marginal, the code was much more complex, and failed on edge cases. Best to keep it simple -- if this needs to be sped up, it could be dropped into C++ via CFFI.

### DIFF
--- a/lime/graphics/Image.hx
+++ b/lime/graphics/Image.hx
@@ -460,6 +460,31 @@ class Image {
 		
 	}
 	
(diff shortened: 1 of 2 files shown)


larsiusprime:
Hang on, I broke flash compiling, I'm pushing a quick fix for that too.

PR #416, opened 2015-04-23
Fixes copyPixels mergeAlpha

mrcdk:
Fixes issue https://github.com/openfl/openfl/issues/618

### DIFF
--- a/lime/graphics/utils/ImageDataUtil.hx
+++ b/lime/graphics/utils/ImageDataUtil.hx
@@ -240,10 +240,10 @@ class ImageDataUtil {
 						sourceAlpha = sourceData[sourceOffset + 3] / 255;
 						oneMinusSourceAlpha = (1 - sourceAlpha);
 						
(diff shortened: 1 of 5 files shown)


mrcdk:
The only problem is that the cpp code doesn't work on windows. Haxe code does.

larsiusprime:
Confirmed this fixes next/neko. Testing windows/cpp now, if it's confirmed broken as mrcdk says I'll look into a patch.

larsiusprime:
Did some digging:
- ImageDataUtil.hx: before the CFFI call, sourceImage.transparent is true
- ImageDataUtil.cpp: sourceImage->transparent is now false

Something in the conversion process is setting transparent from true to false, which makes the cpp part of the code use the first if block (don't merge) instead of the else block, b/c it thinks the image is opaque.

mrcdk:
@jgranick can you check that I haven't messed up anything with the last change?

PR #419, opened 2015-04-26
Font improvements

vroad:
- Font::RenderGlyphs uses anonymous objects again instead of ByteArrays
 - Add Font.RenderGlyphs in haxe for repetitive rendering
 - Disable FreeType's auto hinting

### DIFF
--- a/lime/text/Font.hx
+++ b/lime/text/Font.hx
@@ -139,33 +139,65 @@ class Font {
 		#if (cpp || neko || nodejs)
 		
 		lime_font_set_size (src, fontSize);
+		var ret = lime_font_render_glyph (src, glyph);
 		
-		var bytes = new ByteArray ();
-		bytes.endian = "littleEndian";
-		
-		if (lime_font_render_glyph (src, glyph, bytes)) {
+		if (ret != null) {
 			
-			bytes.position = 0;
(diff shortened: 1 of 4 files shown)


jgranick:
I moved from using an anonymous object to a byte stream in order to improve performance. We could test either method, but I believe that writing into a byte stream (for potentially thousands of glyphs) may give us better results than allocating that many fields

We should probably do only an array or a map for renderGlyphs, I thought a Map might be more useful, but perhaps if you are iterating through the glyphs it would not be.

Have you noticed the difference between auto-hinting enabled and not? I thought I had seen better results with it on

Thank you!

vroad:
Then, I posted the version with "multiple ByteArrays" before, which would not incur unnecessary memory copy.

If you iterate over thousands of glyphs, Map might not be good for performance. RenderGlyphs will also search for duplicate glyphs, but as I said before, most users have list of rendered glyphs, I think.

FT_LOAD_FORCE_AUTOHINT disables fon'ts native hinting and use FreeType's hinter instead. Well, I saw difference with smaller font sizes, but was auto-hinted one good? I think default one is better but others may say not.

I think we should use Font's native hinter by default, unless users want.

PR #422, opened 2015-04-27
Let's make a float division and not an integer one

mrcdk:
Fixes https://github.com/openfl/openfl/issues/631


larsiusprime:
Oooooh, subtle! Because on haxe it auto-casts that to a Float, but on cpp it doesn't! I never would have seen it.

PR #424, opened 2015-04-30
Remove ia32-libs-multiarch installation

ibilon:
The `ia32-libs-multiarch` was removed since ubuntu 13.10,
asking to install it will make `openfl setup linux` fail.

See https://github.com/openfl/openfl/issues/627

Though maybe we'd need to install something else instead,
I'll set up an ubuntu 13.10 64 vm to see if `gcc-multilib g++-multilib` are enough.


raould:
I am on Ubuntu 14.04 x86 (32bit) and ran into this / something like this today (fresh Lime install):
$ haxelib run lime setup linux
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package ia32-libs-multiarch
Called from ? line 1
Called from CommandLineTools.hx line 1359
Called from CommandLineTools.hx line 25
Called from CommandLineTools.hx line 94
Called from CommandLineTools.hx line 1374
Called from utils/PlatformSetup.hx line 406
Called from utils/PlatformSetup.hx line 442
Called from utils/PlatformSetup.hx line 1769
Called from lime/tools/helpers/ProcessHelper.hx line 172
Called from lime/tools/helpers/ProcessHelper.hx line 242
Uncaught exception - Error running: sudo apt-get install ia32-libs-multiarch gcc-multilib g++-multilib []

marcelofrau:
I am having the same problem. There is something we can do to the "openfl setup linux" executes correctly, or there is an alternative to resolve this?

PR #434, opened 2015-05-13
setPixels amend

peteshand:
This brings the lime setPixels method more closely in line with the way the Flash bitmapData.setPixels method work.

### DIFF
--- a/lime/graphics/utils/ImageDataUtil.hx
+++ b/lime/graphics/utils/ImageDataUtil.hx
@@ -1114,6 +1114,17 @@ class ImageDataUtil {
 			var width = image.buffer.width;
 			var color;
 			
+			if (byteArray.bytesAvailable < len * 4) {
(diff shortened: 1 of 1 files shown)


jgranick:
Should it stop writing if it is going to overflow, or should it resize the target ByteArray in order to be large enough?

peteshand:
Hey @jgranick. Yeah actually you're right there. If the byteArray is not long enough extra 0x0 data should be added to fill the rest of the bitmapdata. I've amended my branch

peteshand:
any update on merging this?

jgranick:
I just gave this a try, on Flash Player, it throws an "End of file" error if the `ByteArray` is not large enough. I had a solution worked up, but since Flash is throwing an error, I think we'll leave it as-is.

Please let me know if I'm missing a test case where Flash does exhibit this behavior. Thank you :smile:

PR #443, opened 2015-05-17
basic html5 favicon support

JeriXNet:


jgranick:
What image format are favicons expected to be in? Perhaps we should do a conversion (to a specific size or format) rather than copying all icons?

JeriXNet:
I tested on png and ico. My intention was use svg for this, so it will not need sizes etc, but svg icon support on modern browsers still coming.
I was planning to use it like this:
```
<icon path="icon.svg" unless="html5" />
<icon path="icon.ico" if="html5" />
```

jgranick:
Take a look at the Windows or BlackBerry target platforms, in many targets, we spit out PNG files of the appropriate size (using source JPG, PNG or SVG images) and in Windows, we generate an icon file. We could generate a PNG icon (if we knew a good size) or perhaps could generate an ICO

JeriXNet:
I think there is point in generating ico from svg if there is svg present, because ico supported by all modern browsers and it's actually container, that can contain all needed sizes. I'll look through code for windows and try to implement it as soon as I have time.

jgranick:
This is now supported :smile:

PR #459, opened 2015-06-01
Added meta-package define to lime create project

danielmar:
Added ability to use -Dmeta-package=com.foo.bar with lime create project


jgranick:
Thank you!

This inspired me to give a try with the standard overrides system in the tools, it seems to be working, but I'd love your feedback on how it works, how much is changeable through it, and if it works for you.

The tools support overrides such as `--meta-package-name=com.hello.world` or `--meta-version=1.2.0`

Theoretically, the title, package name, version, company and file name should be above to be overridden now :smile:

PR #464, opened 2015-06-05
Corrected parameter types of two GL calls.

mauvecow:
Function parameter types for glClearDepth and glDepthRange were incorrect, causing invalid data to be sent, breaking some 3D contexts.


mauvecow:
Away3D displays a black screen in all of its examples, currently. That is how I found it in the first place, I ran apitrace through on the currently released versions of OpneFL and received some rather strange results on Linux, namely oddities like: glClearDepth(depth = 0.007812501848093234) When depth is 1 all throughout the haxe portion.

delahee:
Well done ! TBH I searched this one for a long time before abandonning. You rock very much !

PR #473, opened 2015-06-11
fixed: webgl errors involving null/undefined arguments break the whole process

ProPuke:
`glFunctionArgToString` in webgl-debug.js was broken for null and undefined arguments.

In fact the whole thing looks pretty old. We should maybe consider updating to the [newer version by Kronos](https://www.khronos.org/registry/webgl/sdk/devtools/src/debug/). *Buut* haven't tested that in replace, so just patched this simple bug for now :P.

### DIFF
--- a/dependencies/webgl-debug.js
+++ b/dependencies/webgl-debug.js
@@ -170,7 +170,13 @@ function glFunctionArgToString(functionName, argumentIndex, value) {
       return glEnumToString(value);
     }
   }
-  return value.toString();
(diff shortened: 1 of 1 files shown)


jgranick:
Thanks!

I'll try Khronos' version in right now, and see if there are any issues. If not, we'll move forward with it in the next release, otherwise, I pulled your fix, so we can roll it back to your patched version if it does not work out :smile:

PR #474, opened 2015-06-11
add the "headerOnly" parameter to getImage. Only works with PNG's at …

larsiusprime:
…the moment.

### DIFF
--- a/lime/Assets.hx
+++ b/lime/Assets.hx
@@ -254,7 +254,9 @@ class Assets {
 	 * @param	useCache		(Optional) Whether to use BitmapData from the cache(Default: true)
 	 * @return		A new BitmapData object
 	 */
-	public static function getImage (id:String, useCache:Bool = true):Image {
+	public static function getImage (id:String, useCache:Bool = true, headerOnly:Bool = false):Image {
+		
+		trace("getImage() headerOnly = " + headerOnly);
 		
 		initialize ();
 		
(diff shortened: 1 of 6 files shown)


jgranick:
After fussing with this a bit, I took your work and adapted it into a new commit here :smile:

https://github.com/openfl/lime/commit/0090c953b71514afdd8917c33b9472b504d501df

Both should return `null` if the file is not JPEG, PNG, etc

```haxe
var path = Assets.path ("image.png");
var image = PNG.decodeFile (path, false);
trace (image.width);
trace (image.height);
```

larsiusprime:
The only problem with this is that you have to know the exact file extension of your image, and I could have situations where the image path is passed in via data as either as a .png or a .jpg, and sprinkling a bunch of string checks and ifs all around user code seems messy.... is there a way to add a light wrapper that falls through to PNG.decodeFile or JPG.decodeFile?

larsiusprime:
(good work, btw!)

jgranick:
It should work like this:

```haxe
var image = PNG.decodeFile (path, false);
if (!image) image = JPEG.decodeFile (path, false);
if (image) trace (image.width);
```

I'm hesitant to burden the `fromBytes` or `fromFile` methods with this parameter, due to the onload parameter in each. Perhaps if we handled those without an `onload` or ... I'm not sure.

larsiusprime:
Yeah I suppose that works. I'll just make my own wrapper for now.

Unrelated: does ```!image``` work? I thought Haxe didn't auto-cast null to false.

jgranick:
that's what I get for coming off some C++ code, fixed :wink:

PR #475, opened 2015-06-12
Gamepad changes to add custom bindings

wselken:
Added Some minor gamepad changes.

### DIFF
--- a/project/src/backend/sdl/SDLApplication.cpp
+++ b/project/src/backend/sdl/SDLApplication.cpp
@@ -18,7 +18,6 @@ namespace lime {
 	
 	
 	SDLApplication::SDLApplication () {
-		
 		if (SDL_Init (SDL_INIT_VIDEO | SDL_INIT_GAMECONTROLLER | SDL_INIT_TIMER) != 0) {
 			
 			printf ("Could not initialize SDL: %s.\n", SDL_GetError ());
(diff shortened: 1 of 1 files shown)


larsiusprime:
Um, wow. This is pretty cool.

I think I roughly get what this does, but could you explain it a bit? I maintain HaxeFlixel's gamepad API and so I'd like to know exactly what this adds to the equation on the lime level.

project/src/backend/sdl/SDLApplication.cpp
```diff
@@ -131,7 +234,8 @@ namespace lime {
 			case SDL_CONTROLLERBUTTONUP:
 			case SDL_CONTROLLERDEVICEADDED:
 			case SDL_CONTROLLERDEVICEREMOVED:
-				
+			case SDL_JOYDEVICEADDED:
+				printf("game Event %i\n" ,event->type);
```
Gama11: Leftover from debugging?

jgranick:
Could you guys try the latest Lime (after this commit: https://github.com/openfl/lime/commit/5c5430ea0b24a422efd3b91b6164b1fb11d85e91) and see if it helps?

I'm wondering if the JOYSTICK_ADDED is needed. Is it going to add a regular joystick as if it were a gamepad if we do that? Should we be concerned about that problem?

wselken:
it works with your current code, expect the JOYSTICK_ADDED needs to call ProcessGamePadEvent, 
I'm don't see why this would be a problem, since ProcessGamePadEvent checks to make sure the joystick is gamepad before adding it. if its not a gamepad, it will just do nothing, which is the current use case for it now. (from the looks of the code.)

PR #476, opened 2015-06-13
fixed: html5 async asset loading

ProPuke:
`Assets.loadImage`, `loadText` and `loadBytes` were all sync and didn't support non-embedded assets.

This fixes https://github.com/openfl/openfl/issues/369.


jgranick:
Thanks!

PR #477, opened 2015-06-14
Fix legacy Surface.BlitChannel source rect

fserb:
When you see it... :)

(I don't see any equivalent on V2, but if there was any copy&paste maybe we should double check that too).


jgranick:
Thanks :)

Issue #479, opened 2015-06-17
Android touch events return float values


PVince81:
On Android the touch event values appear as values between 0.1 and 1.0 instead of being stage coordinates.

Maybe something is wrong in the code that translates the touch events from Android.

Please note this is non-legacy mode (lime legacy works fine)

I'll try and debug this later.

PVince81:
Some log info in project/src/backend/sdl/SDLApplication.cpp
```
D/LOG_TAG ( 5004): ProcessTouchEvent TOUCH_START: x=0.669314, y=0.291667, id=0
```

I noticed that `onNativeTouch` is called from SDLActivity.java, that seems to be where the magic happens. And onNativeTouch passes the event to SDL which produces the message above.

PR #481, opened 2015-06-18
Fix touch coords in Android SDLActivity

PVince81:
The coordinates are already relative to the display so they should be
kept as such when passing them to SDL native.


PVince81:
Fixes https://github.com/openfl/lime/issues/479

PR #485, opened 2015-06-21
Support for high DPI on next

fserb:
Ok. So here's the change.
It's basically using the proper getsize function and then everything else magically works (we create the rendercontext of the right size, pass the right size along, etc, etc).

It works much better than legacy, because on legacy we create the Lib.stage based on the project options (instead of the window sizes).

I've tested on both retina and non-retina Mac. I've also tested on regular monitors on retina macs.

One thing that is not supported is to have non-retina apps on retina display. This would require a bit of magic. Here is the reasoning (as per SDL/Mac logic):

SDL_GetWindowSize : returns the size in "points", this unit which is not necessarily a pixel, which is sort a windows space metric (in non-retina it's like pixels, in retina is twice that).
SDL_GL_GetDrawableSize : return the actual size in pixels for that window.
There seems to be a weird bug on SDL size that GetDrawableSize returns the full retina size even when we are not on highdpi. I've seen some comments about this [on this SDL bug](https://bugzilla.libsdl.org/show_bug.cgi?id=2716).

We could add a flag right now to disable highdpi, that would not only remove the sdlFlag, but also call the getWindowSize instead of GetDrawableSize.

Also, I haven't done many tests on fullscreen support yet. But it should be fine.

Feel free to hold this pull until I've added the flag. But let me know what you think.


fserb:
Added highdpi config parameter.
There are still [some](https://bugzilla.libsdl.org/show_bug.cgi?id=2215) [issues](https://bugzilla.libsdl.org/show_bug.cgi?id=2716) on the SDL side regarding high DPI.
This changes tries to get around those issues too.
First, we call the correct size function if we are requesting high/non-high DPI.
Second, we don't allow high DPI on fullscreen (you must request higher resolution if you want high dpi on fullscreen). That's the current SDL behaviour, so it's best to stick to it for now.

Tested on mac retina/non-retina, external monitor/retina display, windowed/fullscreen.

lime/app/Config.hx
```diff
@@ -27,5 +27,6 @@ typedef Config = {
 	@:optional var version:String;
 	@:optional var vsync:Bool;
 	@:optional var width:Int;
+	@:optional var highdpi:Bool;
```
Gama11: Shouldn't this be camel case? (`highDpi`)

fserb: I felt a bit weird, since DPI is an acronym, if I do highDPI (Which I did), then the internal flag becomes HIGH_D_P_I, which sucks. And we have "vsync" which is also not capitalized, so.

Gama11: Can't you have `highDPI` and `HIGH_DPI`?

jgranick:
Recently fixed as `allowHighDPI` and `--window-allow-high-dpi` :smile:

Issue #461, opened 2015-06-03
Implement Capabilities.screenResolutions for Blackberry


DanielUranga:
Use https://github.com/openfl/openfl-native/issues/114 as a basis

PR #487, opened 2015-06-23
Hardware screen resolution for Blackberry

DanielUranga:


DanielUranga:
Fixes https://github.com/openfl/lime/issues/461

PR #499, opened 2015-07-14
added legacy support for wiimote gamepad detection

larsiusprime:


larsiusprime:
Can we pull this in soon? It's quite minimal and Flixel-dev has already pulled in changes that are waiting on it.

jgranick:
Okay, but I want to get out of this game of adding new IDs for each exact String match in the future. Lime 2 has a proper API for getting a controller name, next step, I think we should patch something into legacy to get the same if we need it. For example, on connected, it could dispatch the device ID, then we could have a static call to get the controller GUID based on the ID

larsiusprime:
Ah, sounds good! Next time I issue one of these I'll rig something up like that.

PR #506, opened 2015-07-16
fixed: repeat calls to DefaultAssetLibrary load* may call the wrong handlers

### LINKED ISSUE #505: Multiple DefaultAssetLibrary load* only call the last handler [native]
So we have proper async loading on native targets now, yay! There is a seemingly tiny bug, though.

Example code: (note the identical filenames)

    openfl.Assets.loadText('somefile.txt', function(data){
        trace('call 1 completed!');
    });
    openfl.Assets.loadText('somefile.txt', function(data){
        trace('call 2 completed!');
    });

Expected output:

    call 1 completed!
    call 2 completed!

Resultant output:

    call 2 completed!
    call 2 completed!

Because they both have the same id only the last assigned handler always gets called, both times :(

*"Well don't do that, then!"*

I know, it does seem a stupid thing to do. But, sometimes your code does weird things and causes weird scenarios like this that bug in just slightly subtle ways.

ProPuke:
Fixes #505.

Since repeat load* calls to the same asset have the same id, they'd all be assigned the same handler if called multiple times before the queue was executed.

Now passes the handler via an anon struct in an inline function, rather than an id'd Map. Feel free to reject and implement your own way if this isn't to your fancy.


jgranick:
Thank you so much for catching this, should be fixed now! :smile:

ProPuke:
Glad to be of help!

(Now I look at your amend I am puzzled as to why I chose to nest the inline functions. Yours looks much better. I need to cut back on the smack :P)

jgranick:
Well, I debated. I even thought about a class instance, but then I realized that the handler, getMethod and result were all Dynamic. Might as well make the whole thing Dynamic :wink: Thanks again :smile:

PR #513, opened 2015-07-21
guard against division by 0

Hasufel:


Hasufel:
fixed in https://github.com/openfl/lime/commit/e545023

PR #514, opened 2015-07-21
build fix

Hasufel:


jgranick:
Thanks, I just threw in an `(int)` :smile:

PR #522, opened 2015-07-27
Fixed CPP target embedded assets

Matan:
Embedding fonts for Android isn't currently working as the assets for CPP targets aren't embedded anywhere. NOTE: I've only tested for Android with Fonts. To recreate this issue, create the TextMetrics sample project. Set the assets to embed in the project.xml and try to compile.

Edit: Added failsafe to Preloader for small AIR binaries. Sometimes the the on complete handler would miss the dispatch, which causes the app to be stuck on the preloader screen. I've added a simple check in the enterframe loop to prevent the app from getting stuck.

### DIFF
--- a/lime/app/Preloader.hx
+++ b/lime/app/Preloader.hx
@@ -286,7 +286,14 @@ class Preloader #if flash extends Sprite #end {
 	
 	#if flash
 	private function current_onEnter (event:Event):Void {
-		
+
+		// Failsafe for small binary missing the 'onComplete' handler
+		if(!complete && Lib.current.loaderInfo.bytesLoaded == Lib.current.loaderInfo.bytesTotal)
+		{
+			complete = true;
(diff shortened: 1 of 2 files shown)


jgranick:
Thanks!

Do you think we might have a problem with slow Flash Player loads reporting bytesTotal as 0 at first, and assuming (prematurely) that it's finished? This was a problem a long time ago, but I'm not sure if Flash Player still does this at times

Matan:
No problem! I thought about that, but deemed it unlikely (no proof). The ADL actually caught the complete event whereas the actual device didn't. I think it's due to nature of the application load procedure on the device. I'll do some more testing and fix up where needed.

Please make sure I did the correct thing with the embedding assets. All I know it fixes the compile and the font renderings, but like I mentioned, I don't know if it's correct for the other type assets.

PR #529, opened 2015-07-31
settable lime.ui.Window.title, SDL only

jarnik:
I needed to change native window title in runtime.
SDL only, not implemented for HTML5, nor Flash.
My first lime contribution, works for me like that, should be reviewed.


jgranick:
This is awesome, thank you! :smile:

PR #530, opened 2015-07-31
displays: added atWindow and intersectingWindow functions

larsiusprime:
Happy to change names/organization, but it seems we'll need functions like these (for instance to drive openfl's Capabilities.screenResolutionX/Y), and it looks like we have enough information to process them now.

### DIFF
--- a/lime/system/Display.hx
+++ b/lime/system/Display.hx
@@ -51,7 +51,7 @@ class Display {
 			
 		}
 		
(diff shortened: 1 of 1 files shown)


jgranick:
Thank you, this is a great idea!

This is implemented now as `window.display`, it will return the display showing the center point of the window :smile:

PR #536, opened 2015-08-10
Process all touches instead of only the first touch.

xastor:
One touch event can contain updates to more than one touch. I just added a loop over the changedTouches property, so all changes to all touches are properly dispatched.


jgranick:
Sorry for the delay in closing this, this was merged and resolved quite some time ago :smile:

PR #543, opened 2015-08-15
Update neko linux

ibilon:
Same as https://github.com/openfl/lime/pull/391 (glibc 2.13 and -rpath) but this time with libneko.so

Worth waiting a couple of days to merge,
we'll need to update our version when this is merged https://github.com/HaxeFoundation/neko/pull/86 since it's a really important fix


kulick:
Woot, good to hear.  ;)

Let me know if any questions/issues come up.  Thanks!

PR #546, opened 2015-08-16
Additional key mappings for Flash and HTML5 targets

Rezmason:
Extending key mappings in Flash and HTML5 targets, based on the constants listed in the [AS3 Keyboard class reference] (http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/ui/Keyboard.html).

Both targets tested in Safari.


jgranick:
Thank you! :smile:

PR #553, opened 2015-08-27
fixed: legacy word-wrapped TextFields would sometimes leave an extra empty line below

ProPuke:
disclaimer: I just pushed buttons and this worked (fixed my test cases, and did not break any text not displaying the bug). I actually have no idea what I'm doing.


jgranick:
Thanks! Just cherry-picked the commit :smile:

ProPuke:
xD (I don't know how to make git not messy! D:)

jgranick:
This might help :smile:

http://stackoverflow.com/questions/8134960/how-to-revert-master-branch-to-upstream

ProPuke:
Cheers :]

So.. Turns out this "fix" is actually wrong. It was a higher-level problem with another library and this only masked the problem :( This change is actually incorrect (hey, I did include a disclaimer!)

Could this be reverted? Sorry! (now I've made more mess)

PR #567, opened 2015-09-11
[legacy] handle non-US layout keyboards in input textfields

DanielUranga:
I had proposed this change a time ago, but dont know what happened.
Trying again. We need this, since on legacy, TextInput fields aren't working right for non-US keyboards. We're testing with latest lime and Spanish keyboard and its not possible to write "@", "|", "ñ", etc.
This change fixes it.


nescribal:
Thanks! I need this change please merge!

PR #571, opened 2015-09-17
Added support for cpp cross-compile from Linux-64bit to Windows-32bit

jobs-git:
Added support to cpp cross-compile from Linux-64bit to Windows-32bit, recompiling of lime tools is required to take effect


larsiusprime:
wait, really??? What black magic is this?

jobs-git:
yhea, it can be done now, using mingw of course :)

PR #581, opened 2015-09-22
enable finalizer for lime.graphics.cairo.Cairo

james4k:


vroad:
It's disabled because of crash issues on nodejs. Some of needed objects were already garbage collected when fnalizer was invoked by V8 JavaScript engine.

PR #586, opened 2015-09-24
Fixed issue on DLL not copied during cross-compile

jobs-git:
The suggested fix recovers the ability of lime to automatically copy DLL files to bin directory during Mingw cross compilation from Linux to windows.


ibilon:
Wouldn't that prevent linux users to make mac neko cross-build?

jgranick:
Thanks!

I think this should do it: https://github.com/openfl/lime/commit/e0cbcb78101526a490789ad424e59cb1667b4eda

I want to be able to allow `-neko` to still force Neko cross-desktop building :)

jobs-git:
Alright, Rock on! :)

PR #589, opened 2015-09-26
Fixed Linux cross compilation

jobs-git:
Fixed Linux cross compilation using Mingw.

It is interesting to note that Fedora and Ubuntu could cross compile Windows 32bit application using Mingw.

Resulting binaries have been found out to be fully functional in Wine Abstraction Layer, Windows 7 and Windows 10.

### DIFF
--- a/lime/tools/platforms/WindowsPlatform.hx
+++ b/lime/tools/platforms/WindowsPlatform.hx
@@ -38,7 +38,7 @@ class WindowsPlatform extends PlatformTarget {
 		
 		super (command, _project, targetFlags);
 		
-		if (project.targetFlags.exists ("neko") ) {
+		if (project.targetFlags.exists ("neko")) {
 			
 			targetType = "neko";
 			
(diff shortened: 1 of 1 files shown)


lime/tools/platforms/WindowsPlatform.hx
```diff
@@ -141,12 +155,21 @@ class WindowsPlatform extends PlatformTarget {
 				
 			}
 			
+			if (enableLinuxMingw) {
```
jobs-git: These are compile flags needed when compiling Win32 application from Linux (Fedora and Ubuntu).

larsiusprime:
@jgranick Does this mean that an automated build server such as the one we've got on TeamCity could use this to compile for linux as well as windows?

jobs-git:
Hi, sorry to close this, but the recent merge on Hxcpp could now allow openfl and lime to do cross desktop compile naturally, the only thing left is for Lime to support cross-compile from linux to windows. present build supports neko as the default. but when this is updated, Lime and OpenFl could do CPP cross-desktop compile.

https://github.com/HaxeFoundation/hxcpp/pull/309

PR #590, opened 2015-10-01
Fix crash after reading file on Android

vroad:
PiratePig runs now with this fix, though only white background and text shows up on the screen.

### DIFF
--- a/project/src/backend/sdl/SDLSystem.cpp
+++ b/project/src/backend/sdl/SDLSystem.cpp
@@ -401,7 +401,8 @@ namespace lime {
 		
 		#ifndef HX_WINDOWS
 		
-		return ((SDL_RWops*)handle)->type == SDL_RWOPS_STDFILE;
+		int type = ((SDL_RWops*)handle)->type;
+		return type == SDL_RWOPS_STDFILE || type == SDL_RWOPS_JNIFILE;
 		
 		#else
 		


jgranick:
Unfortunately, Android JNI_FILE cannot be treated as a "real" file.

Unless I am mistaken, you cannot read until EOF on these virtual files. As a result, when all is well, it may work properly, but when trying to read a corrupted file, or in some cases when using something like OGG Vorbis decoding, you can end up skipping into the contents of another file within the APK and really causing trouble :smile:

Does PiratePig work for you in the main-line versions of OpenFL and Lime? Perhaps this is caused by some difference in fork?

vroad:
On non-windows platforms, lime uses SDL_RWOps abstraction layer for all file operations, doesn't it?
So it doesn't matter if it's a virtual file or not, though I don't know what happens if you try to load the same file again in Bytes constructor, without closing previous handle.

vroad:
Well, that was not a problem... 
The problem is that second file read returns corrupted data, at least on some systems.
Also, "file->getFile" returns valid handle that libraries like Ogg can use, but I don't know how it works.
Is SDL extracting files in apk to somewhere?

vroad:
I couldn't install lime 2.6.7 from haxelib (It still tries to install 2.6.6), so I downloaded zip manually and installed with 
`haxe local lime-2.6.7.zip`.
2.6.7 (official) doesn't crash anymore, then I tried running my build with same source in git.
I got crash in OAL_OggBufferRead again.
I'm not sure why I got crash with the same source as official version. I made sure that submodules are up to date.

SDL source says that assets files are read with standard *nix calls, so I think that asset files can be read by other libraries.
https://github.com/spurious/SDL-mirror/blob/30f533b54870c22e6af6b6e822f72c736a580f5b/src/core/android/SDL_android.c#L730

vroad:
Anyway here is what I got without this PR. Still does not work even without changes from my repo.
What is still different between my build and your build??
https://gist.github.com/vroad/735e053cbc8090b7d8c1

PR #595, opened 2015-10-03
Minor fixes for tvOS + ARC

tanis2000:
I've fixed a few bugs in the tvOS implementation and force the iOS target to use ARC as SDL takes it for granted and some of the files of the latest revisions won't compile if it's not enabled for iOS/tvOS targets.

### DIFF
--- a/lime/tools/platforms/IOSPlatform.hx
+++ b/lime/tools/platforms/IOSPlatform.hx
@@ -224,7 +224,6 @@ class IOSPlatform extends PlatformTarget {
 			context.OBJC_ARC = true;
 			
 		}
-		LogHelper.info("OBJC_ARC:" + context.OBJC_ARC);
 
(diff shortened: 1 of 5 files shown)


tanis2000:
Latest commit adds the definitions for iOS/tvOS MFI game controllers

larsiusprime:
If you can give me the numerical ID's for each button, we can add native definitions for them over in HaxeFlixel's new fancy Gamepad API.

tanis2000:
Here you can find the mapping that I added to Lime: https://github.com/tanis2000/lime/commit/8a765e45d3f58da0503f569c16ed77ec41344116

It should be fine as it is the same I'm using with plain SDL.
Nonetheless, wihout touching the current HaxeFlixel code, I do get the axis but not the buttons. If you could lend me a hand, that would be awesome!

PR #596, opened 2015-10-05
set armv7+x86 as default targets for Android builds

ph0b:
Hi,

They're more and more x86 devices on the market since 2012, such as the Google Nexus Player, Nokia n1, Tesco Hudl2, Asus Zenfones, Samsung Galaxy Tab 3 10.1, Motorola Razr i...
 
Currently full x86 architecture support is there when adding *-emulator* option or specifically adding x86 architecture for android inside *project.xml*, I'm proposing to change it so arm and x86 libraries are included by default in android builds, not only when building for the emulator.

### DIFF
--- a/lime/project/HXProject.hx
+++ b/lime/project/HXProject.hx
@@ -148,15 +148,7 @@ class HXProject {
 					
 				} else if (target == Platform.ANDROID) {
 					
-					if (targetFlags.exists ("simulator") || targetFlags.exists ("emulator")) {
-						
-						architectures = [ Architecture.X86 ];
-						
-					} else {
-						
-						architectures = [ Architecture.ARMV7 ];
-						
-					}
+					architectures = [ Architecture.ARMV7, Architecture.X86 ];
 					
 				} else if (target == Platform.TVOS) {
 					


ibilon:
I looked at my phone spec and the last one indeed had a x86 cpu, and the new one even has a x64.

But I couldn't manage to get android cpu market share anywhere.

By how much does adding the x86 bin increase the apk size?

ph0b:
The market share depends a lot depending on the countries. It's a lot less in the US than in other regions but I don't have the numbers.

The recently announced Chromebook support for Android apps will continue making it grow, since the majority of Chromebooks are x86/x64.

The x86 bin is approximately of the same size than the arm one, so it does add it to the apk size (~10MB with a regular build?). It's possible to split and publish different APKs as described on this page if it becomes a problem: http://tools.android.com/tech-docs/new-build-system/user-guide/apk-splits

jgranick:
I thought x86 Android devices could still run ARM?

http://stackoverflow.com/questions/31047552/android-running-armeabi-only-apps-on-intel-devices

ph0b:
All the Intel based ones do, yes.
The software translation in use on these devices may have a non-negligible impact on launch time, performance, battery consumption. It's working very well these days but not every manufacturer keeps updating the devices with the latest version of the OS and libs.

jgranick:
I feel a little bit torn -- I'm generally a fan of making our default build configuration broader and more compatible, but adding a default architecture burdens the build time and file size required for any project. As you say, you can add an architecture already, I wish we had better statistics on which architectures are in use.

The last problem is using native extensions that may not have x86 binaries built. Some extensions for Android don't support this, and it is not possible to get a new one from the vendor.

vroad:
I guess we should still be able to build for one architecture when debugging. Couldn't we override default CPU architecture instead, either from command line or project.xml?

Hasufel:
Regarding x86 market shares stats:
Intel has canceled 3 months ago all of its publicly announced 14nm x86 smartphone SoCs (Atom Z lineup, powering those old devices you mentionned)
and most of its 14nm tablet SoCs that would have shipped in Android devices -- namely SoFIA 3GX, SoFIA LTE, SoFIA LTE2, and Broxton.
The entire x86 Atom smartphone chip lineup has been cancelled. That is a 0% market share.
Sofia 3GX was supposed to be ready for 2014 and SoFIA LTE for 2015; Broxton was supposed to be released a year ago.
Intel announced in March this year that it lost over USD 10 Billion dollars on mobile chips.
Intel is getting out of the tablet (and mobile) business altogether.
Correlate that with Windows Phone on the verge of total failure (less than 2% of the current mobile market, the entire Nokia division being laid off as we speak) while Intel is in wholesale retreat from the phone and tablet market:
the old Intel/Microsoft combo totally did miss the mobile revolution, and ARM has but won 100% of the market. For now.
I would advise against adding x86 as default. It does not make sense, would double the build time, and you can already set what architecture you want to build to in the project configuration with a single line.

ph0b:
For the reasons you exposed, I agree it doesn't make sense to make a switch to x86 by default right now.

Considering the Chromebooks support and the current market share (it's not because they aren't many x86 phones/tablets sold anymore that the user base isn't sizable and there will be not come back..), I think it's still important to allow users to build projects for ARM+x86 easily if they want to.

Can we set more than one architecture in the project configuration?
If it's the case, then you can close this old PR.
If not, it'd be better if you can allow for setting multiple architectures for android in the project configuration, and add a new parameter `-android-arm-x86` `-android-archs=arm,x86` or whatsoever to allow people to do so. Being able to build arm+x86 apks is important to let people have and distribute just one APK which supports both platforms (and emulators/CI environments) easily.

ibilon:
Adding `<architecture name="x86" />` to the project.xml is enough (since armv7 is already enabled),
if you want even more simply add more `<architecture />` tags.

PR #609, opened 2015-10-16
Fixes for Retina scaling

tanis2000:
This fixes the problem of scaling on Retina displays. 
I'm going to submit a PR to OpenFL as well as it's messing up things a bit. If you can come up with a more elegant solution, please go ahead :)

### DIFF
--- a/lime/_backend/native/NativeApplication.hx
+++ b/lime/_backend/native/NativeApplication.hx
@@ -520,8 +520,8 @@ class NativeApplication {
 				
 				case WINDOW_RESIZE:
 					
-					window.__width = Std.int(windowEventInfo.width * window.scale);
-					window.__height = Std.int(windowEventInfo.height * window.scale);
+					window.__width = windowEventInfo.width;
+					window.__height = windowEventInfo.height;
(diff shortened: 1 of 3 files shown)


kiroukou:
+1 
We had to do that same kind of fix here.

As it is a major problem and still relevant, could it be possible de accept the pull request or apply the fix (since no more sync with develop branch) ?

tanis2000:
@kiroukou is it still relevant? I thought this had been fixed months ago by @jgranick

kiroukou:
Humm.  I don't know how I have have missed the closed status information... is that recent?

With version 2.9 we had the issue, indeed.  We had to scale the coordinates directly on our engine API.
(for the viewport, the touch/clicks coordinates were just fine!)

We are quite good with it so far, simple fix for us, but I think this should be a bit more tested.  (that's why I though the pull request was still open)

tanis2000:
I'm going to close this one as I feel it's no longer relevant. The current Lime code for Retina displays has been tested and in use for quite a few months, so I guess that if no one reported any issue it should be considered solid. Feel free to open a new issue if needed.

PR #614, opened 2015-10-24
FIX linker paths for raspberry pi

gepatto:
Adds linker flags to build.xml to enable compiling SDL with SDL_VIDEO_DRIVER_RPI
on Raspbian Wheezy and Jessie.

No raspberry pi specific changes in sdl-module in this pullrequest just yet.

### DIFF
--- a/project/Build.xml
+++ b/project/Build.xml
@@ -349,11 +349,11 @@
 			
 			<section if="rpi">
 				
-				<!-- <lib name="-lbcm_host" /> -->
-				<lib name="-ldl" />
+				<lib name="-lbcm_host" />
 				<lib name="-lm" />
 				<lib name="-lGLESv2" />
 				<lib name="-lEGL" />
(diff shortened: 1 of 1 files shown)


jgranick:
The Raspberry Pi binaries are built from Ubuntu, so (unfortunately) this will break the build server. I wish we could just do X windows with GPU acceleration :confused:

gepatto:
I'm afraid gles2 in X11 is not going to happen soon :(
So compiling for Rpi on debian will require a manual step of  editing build.xml
That's not too much of a problem.

It's the sdl-submodule that requires the most changes...

jgranick:
Here's an idea:

http://forum.xda-developers.com/showthread.php?t=2522907

It's not pretty, but hacking libEGL, then copying it into the application directory, so it finds its own version instead of the system one. I believe this enables hardware acceleration with X11 + GLES2

gepatto:
Interesting, will look into this!

PR #629, opened 2015-11-17
Make build number file configurable

sebthom:
Currently the build number file .build is "hardcoded" in the app's output directory (bin). 
When the bin directory is deleted the build number information is lost. 
Same is true in case the bin directory is GIT/SVN ignored, on a fresh checkout the build number information is lost too.
Therefore it can be desirable to be able to control, where the build number file is actually stored. 

This patch makes the path to the file holding the last build number configurable via

```
<haxedef name="lime-buildnumber-file" value="[filepath]" />
```

in the project.xml file.

E.g.

```
<haxedef name="lime-buildnumber-file" value="settings/last_build_number" />
```

### DIFF
--- a/tools/CommandLineTools.hx
+++ b/tools/CommandLineTools.hx
@@ -968,7 +968,8 @@ class CommandLineTools {
 		
 		if (project.meta.buildNumber == "1") {
 			
-			var versionFile = PathHelper.combine (project.app.path, ".build");
+			var versionFile = project.haxedefs.get("lime-buildnumber-file");
+			if (versionFile == null) versionFile = PathHelper.combine (project.app.path, ".build");
 			var version = 1;
 			
 			try {


jgranick:
How do you feel about this commit?

https://github.com/openfl/lime/commit/d0f2c6699937190bbc54cb94dfceebe40c536873

I believe the build number is most used when you publish to iOS/Android, but is not so important that it is updated every time you build.

One solution (committed above) is to use GIT commits to mark the build number, but in addition to the default `<meta build-number="git" />` behavior (which will now work by default), I added support for `<meta build-number="git+100" />` (or some other value) to apply an offset.

Does this solve your problem? Are you committing code in GIT, or another version control system?

Eiyeron:
I haven't checked how it works but something to allow different numbering if compiling a debug build or not could be very useful. I usually get into rapid code-test cycles and in one week, I got around 5000 debug builds for only showing up one or two (private) release builds.

jgranick:
The build number is generally invisible to users, and is used for Apple and Google to basically track whether one copy is newer than another -- a pretty version number might be "1.0.0 alpha" or "2.1.3 beta" but internally they use a single Int-based build number to translate that 3014 is newer than 2012

Do you commit to GIT during your development time? If so, using the GIT commit number will create a linear history that should work fine for the build version, as for the user-facing release number, that would always be something you update yourself in `<meta version="" />`, from "1.0.0" to "1.0.1", for example.

sebthom:
Hi Joshua,
I was hoping for a more generic, SCM independent solution. Your proposed patch works with GIT but doesn't help when using other SCMs.

jgranick:
What SCM are you using? Perhaps we can add a command to do something similar with a different source control system

sebthom:
We are for example required to use SVN for some projects.

If something like this would be implemented for SVN it should also support offline builds, i.e. when the SCM is not reachable for whatever reason (e.g. while traveling).

jgranick:
Perhaps we could parse `svn info`, or does that work only online?

jgranick:
SVN auto build number support is in :shipit:

PR #632, opened 2015-11-23
Angle

larsiusprime:
adds the ability for the user to ship two ndll's -- "lime.ndll" and "lime-angle.ndll", and the openfl app will switch between them in response to the presence or absence of the "--angle" command line argument at launch time.

Also adds support for processing window configs via command line arguments at launch time.

This way you can get all these options w/ command line arguments:
- Desktop GL (myapp.exe)
- ANGLE (myapp.exe --angle)
- Software rendering (myapp.exe --window-hardware=false)

Ultimately a plan will have to be made for either providing ANGLE functionality in the base lime.ndll, or adding a build option that supplies both with the final app, for this to more convenient.


larsiusprime:
Obsolete, see this branch for continuing work towards ANGLE support:
https://github.com/larsiusprime/lime/tree/angle2

jobs-git:
So, you will be pushing a new one?

larsiusprime:
see branch "angle2" on larsiusprime/lime

PR #637, opened 2015-11-25
Fix correct path to recover the usual behaviour of Lime during Windows Build

jobs-git:
In Linux, 'lime build linux' automatically lets you see the compiled project by running it after build, in build windows this is not the case. 

Although WindowsPlatform.hx is programmed to do so, it does not happen due to typo.


jgranick:
This depends on the command processor. The default Windows "cmd.exe" processor would not handle "./PiratePig.exe" properly, though "PiratePig.exe" would run the program.

jobs-git:
Closing this now since cmd behavior is different from linux-terminal.

PR #638, opened 2015-11-25
Fill rect alpha (closes #631)

### LINKED ISSUE #631: FillRect() not respecting alpha
Currently the behaviour of `bitmapData.fillRect()` is not aligned with flash. Specifically if `color` contains an alpha value the results are different:

before

I think I have a fix for this, that I'm testing by skipping cffi in neko (https://github.com/azrafe7/lime/blob/c9f8bb99606b2f075465a067c14a5184bc50d37d/lime/graphics/utils/ImageDataUtil.hx#L325), but compiling it targeting cpp results in `Uncaught exception - Could not find NekoAPI interface.`

The fix seems to work fine in neko and html5:

after

How can I make it so that cpp follows the same codepath without erroring (using `-Ddisable-cffi` doesn't seem to work)?

azrafe7:
This PR fixes fillRect() by taking care of the alpha value in the `color` parameter. Specifically:
- clears the rect before drawing it in html5 (otherwise blending would be applied)
- uses multiplyAlpha once on fillColor (which also prevents the overriding of values caused by RGBA.multiply)
- similar code for both cpp and pure hx

see https://github.com/openfl/lime/issues/631

### DIFF
--- a/lime/graphics/utils/ImageCanvasUtil.hx
+++ b/lime/graphics/utils/ImageCanvasUtil.hx
@@ -203,7 +203,7 @@ class ImageCanvasUtil {
 			
 			if (image.transparent && ((color & 0xFF) == 0)) {
 				
-				image.buffer.__srcCanvas.width = image.buffer.width;
+				image.buffer.__srcCanvas.width = image.buffer.width; // force clear canvas
 				return;
 				
 			}
@@ -228,6 +228,8 @@ class ImageCanvasUtil {
 			
 		}
 		
+		// clear first to avoid blending
+		image.buffer.__srcContext.clearRect (rect.x + image.offsetX, rect.y + image.offsetY, rect.width + image.offsetX, rect.height + image.offsetY);
 		image.buffer.__srcContext.fillStyle = 'rgba(' + r + ', ' + g + ', ' + b + ', ' + (a / 255) + ')';
 		image.buffer.__srcContext.fillRect (rect.x + image.offsetX, rect.y + image.offsetY, rect.width + image.offsetX, rect.height + image.offsetY);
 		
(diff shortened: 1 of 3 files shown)


azrafe7:
Tested with https://gist.github.com/azrafe7/8e87f57098bdf062af9a

jgranick:
Wouldn't there be a problem if you were performing a fill over an existing drawing? Clearing first would delete the original contents, even if you were filling only partially visible color, right?

larsiusprime:
If I understand this correctly, clearing first is the intended behavior of the spec; fillRect should literally just fill the entire specified region with the given pixel color; floodFill is for something more like the photoshop fill bucket tool.

jgranick:
Perhaps I thought of graphics.beginFill for a moment, not bitmapData.fillRect, I'll try and take a look at this soon (I bet on HTML5 we can change the blend mode, though, instead of clearing)

azrafe7:
@larsiusprime is correct about the spec (or at least that's how I interpret the flash behaviour).

@jgranick I've tried using blending to avoid the clearRect(), but it seems there's no way to do it when alpha is involved. Please correct if I'm wrong, but fiddling with the example [over here](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation#Examples) with:

```js
ctx.globalCompositeOperation = "source-over";

ctx.fillStyle = "rgba(255,0,0,.5)";
ctx.fillRect(10, 10, 100, 100);

//ctx.clearRect(50, 50, 100, 100); //   <- uncomment this
ctx.fillStyle = "rgba(0,0,255,.5)";
ctx.fillRect(50, 50, 100, 100);
```

I couldn't find any value for `globalCompositeOperation` to get the expected result (although I still wonder why `copy` doesn't do it).

jgranick:
Thanks again for your help, it looks like we were still missing the canvas-based image path. I handled this using a variation of your approach, but doing a clear only if the alpha is less than 100%

PR #639, opened 2015-11-26
Fix copyPixels w/ alpha

azrafe7:
This PR attempts to make copyPixels() behave like flash on all targets (currently 99% similar, due to off-by-one values in some cases).
- [cffi] I've gone with a readable impl., as moving the ifs outside and using different loops didn't seem to bring significant performance improvements (maybe branch prediction at work?!)
- [html] now uses the pure haxe non-cffi copyPixels(), after creating buffers (canvas/data) for source and alpha images

ref: https://github.com/openfl/openfl/issues/910

tested with https://gist.github.com/azrafe7/85e3cf486ffe23b13102

after

Additional notes:
- in my benchmarks flash is still slighly faster than cffi (0.001s vs 0.003s)
- probably html5 could be improved as actually relies on convertToCanvas/createImageData (but it's already around 0.006s)
- although my tests are passing mind that they're programmatically created BitmapDatas, so it might be wise to test further using actual pngs
- also this needs the other PR about fillRect alpha to behave _better_ (you can spot a difference in `neko with transparent ON`)

### DIFF
--- a/lime/graphics/utils/ImageCanvasUtil.hx
+++ b/lime/graphics/utils/ImageCanvasUtil.hx
@@ -92,14 +92,43 @@ class ImageCanvasUtil {
 	
 	public static function copyPixels (image:Image, sourceImage:Image, sourceRect:Rectangle, destPoint:Vector2, alphaImage:Image = null, alphaPoint:Vector2 = null, mergeAlpha:Bool = false):Void {
 		
-		convertToCanvas (sourceImage);
-		createImageData (sourceImage);
-		if (alphaImage != null && alphaImage.transparent) {
-			convertToCanvas (alphaImage);
-			createImageData (alphaImage);
+		if (destPoint == null || destPoint.x >= image.width || destPoint.y >= image.height || sourceRect == null || sourceRect.width < 1 || sourceRect.height < 1) {
+			
+			return;
+			
 		}
 		
-		ImageDataUtil.copyPixels (image, sourceImage, sourceRect, destPoint, alphaImage, alphaPoint, mergeAlpha);
+		if (alphaImage != null && alphaImage.transparent) {
+			
+			if (alphaPoint == null) alphaPoint = new Vector2 ();
+			
+			// TODO: use faster method
+			
+			var tempData = image.clone ();
+			tempData.copyChannel (alphaImage, new Rectangle (alphaPoint.x, alphaPoint.y, sourceRect.width, sourceRect.height), new Vector2 (sourceRect.x, sourceRect.y), ImageChannel.ALPHA, ImageChannel.ALPHA);
+			sourceImage = tempData;
+			
+		}
+		
+		sync (image, true);
+		
+		if (!mergeAlpha) {
+			
+			if (image.transparent && sourceImage.transparent) {
+				
+				image.buffer.__srcContext.clearRect (destPoint.x + image.offsetX, destPoint.y + image.offsetY, sourceRect.width + image.offsetX, sourceRect.height + image.offsetY);
+				
+			}
+			
+		}
+		
+		sync (sourceImage, false);
+		
+		if (sourceImage.buffer.src != null) {
+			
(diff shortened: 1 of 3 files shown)


azrafe7:
Oh... Travis CI is failing because initialization of alphaView is behind a conditional, but it's only used when that specific conditional is true!

I see haxe version used for CI is 3.2.1, it seems to work fine for me with 3.3.0 (git build development @ d99cb4d).

Should I fix that by guarding it with a null check?

larsiusprime:
Wow this is extremely thorough

azrafe7:
Hey @larsiusprime, thanks.

I'm a bit rusty in c++, so would appreciate some help in passing CI and do the correct thing. 
Travis is rightly complaining about `./src/graphics/utils/ImageDataUtil.cpp:163:54: error: taking address of temporary [-fpermissive]`.

How I should go with [this](https://github.com/azrafe7/lime/blob/e0fcac428c2ad2d9e0db857598b9166baf5bf83d/project/src/graphics/utils/ImageDataUtil.cpp#L163)? Just use `new`  and let GC do its work, and/or dipose the mem at the end of the function?

larsiusprime:
I'm not sure that the GC is going to kick in if you're already directly at the C++ level, I think the hxcpp GC only operates on code you write in haxe. I'm by no means a C++ expert either. Maybe ask @mauvecow or @jgranick

mauvecow:
Yeah, Haxe's GC doesn't work on data you allocate manually in C/C++. So that alphaPoint you're allocating is a memory leak. For that, since it's not used anywhere else, you can scope a Vector2 onto that scope and then just reference that instead of allocating anything. eg

    Vector2 temp;
    if (alphaPoint == NULL) alphaPoint = &temp;

alphaView is a little trickier. ImageDataView doesn't appear to allocate anything itself, but it doesn't have a blank constructor and it's being taken out of scope. I think if you added a blank constructor to it, you can just ditch the pointer reference and do 'alphaView = ImageDataView (alphaImage, &alphaRect);' no problem.

If that ends up not being an option, yeah temporarily allocate and delete at function end is fine.

azrafe7:
Thanks @mauvecow. I've gone with a temp vec for alphaPoint, and new/delete for alphaView.

I'm wondering if it would be better to just branch on needsMultiplyAlpha and use local vars instead...

azrafe7:
Don't merge this yet!

html5 is slower than previously stated with this impl. (I was getting wrong timings as getPixel32 triggered a side effect not accounted for in the tests).

azrafe7:
We need more people testing this but, as it is, it seems to work reliably neko/cpp.

I've reverted my changes for js, as invoking convert/create noticeably slowed down the thing. I was also getting different results for canvas/dom/webgl (canvas and dom worked but were too slow, webgl had other _not-directly-related_ problems), and being not familiar with the differences between them I preferred to just leave js out of the PR.

The logic so far _seems_ to be solid (needsMultiplyAlpha/needsBlending), so if anyone with more knowledge about the js target can take a shot at how to correctly implement it would be really nice (...actually I'd be really interested in following that process).

azrafe7:
I was thinking of a faster way to do this in html5 by leveraging `globalCompositeOperation`, but trying to come up with a POF saw myself struggling with `__srcContext/__srcCanvas/etc`.

I still think the idea might be worth persuing. Some pseudocode:

```
  if (needsMultiplyAlpha) {
    sourceImage = sourceImage.clone(); // here we could also just clone the interested area instead of the whole bitmapData
    sourceImage.multiplyAlphaWith(alphaImage);
  }

  if (needsBlending) {
    // ensure to set globalCompositeOperation to 'source-over' so we get blending...
    destImageContext.drawImage(sourceCanvas, properOffsets/RectsArgs);
  } else {
    destImageContext.putImageData(sourceCanvas, properOffsets/RectsArgs);
  }
```

I'll probably try to come up with a pure haxe/html5 impl. to test if this is viable, but any efforts in coding this with openfl directly would be great.

jgranick:
Thank you so much for all of your help. Your work has been extremely thorough, and was directly helpful in the commit I've just made, implementing improvements for better alpha handling in copyPixels :smile:

I just realized that I don't believe that we have 100% accurate behavior in the `ImageCanvasUtil.copyPixels` render path -- your sample hits the data path. I'll have to try and circle back with tests on the canvas path, but this is already much better :smile:

azrafe7:
Great! I knew it's a feature not many use (alphaImage), but glad to know my investigations were useful! :+1: (...at last :stuck_out_tongue_winking_eye:) 

PS: it seems there's a redundant check here https://github.com/openfl/lime/blob/b60f6bd94e44e698f5fce252d32168386523765e/lime/graphics/Image.hx#L362

Thanks, really!

PR #640, opened 2015-11-26
Window can now set resizable & borderless at runtime

larsiusprime:


larsiusprime:
@jgranick do these seem useful to expose? If not I'll just close this issue.

jobs-git:
I think this is useful, please don't close.

PR #641, opened 2015-11-26
Angle support + Software fallback for Windows XP

larsiusprime:
This should give us dang near 100% coverage/compatibility across all of Windows.


larsiusprime:
TODO:

1. Since we default to software on XP, we can probably drop d3dcompiler_43.dll and any associated hints. Not sure if this requires recompiling the ANGLE dll's.
2. My fork of ANGLE is [here](https://github.com/larsiusprime/angle). Not sure if you want to set this up as a lime submodule or whatever or not.

larsiusprime:
Super special thanks to @mauvecow and vroad, whatever the associated github name is.

jobs-git:
Hey, great work!  How do we deal with cases were there were only builtin graphics card like IntelGraphics or IntelHD?

larsiusprime:
@Jobs-git Can you be more specific about what you're talking about?

Here's basic information that might help in the meantime.

This pull makes ANGLE available on all versions of windows except XP. ANGLE runs OpenGL via directx, which should have better compatibility and support across the board than regular OpenGL.

On XP it forces software rendering, which is accurate but slow, and doesn't depend on graphics cards or drivers.

If you are talking about machines that don't use windows xp, but DO use hardware rendering, but still have some problem or other, you can always force software rendering by passing in "--window-hardware=false" on the command line.

jgranick:
https://github.com/openfl/lime/commit/8ef9b2893926c3240ac616417c964acfb2d5d334

jobs-git:
Hi @larsiusprime, Im am trying to integrate your work to the latest lime build I have, 2.7.0, however there is no 'PROJECT' folder existing there, how could I be able able to build an app with ANGLE support?

ibilon:
@jobs-git you need the version from github and not haxelib if you want to rebuild.

Joshua already added angle support so you won't have anything to change to get it.

jobs-git:
I did, but I got problem with ByteArray, guess its not ready yet, so I reverted it back to 2.7.0 then manually included the changes, but I got stucked with the 'Project' folder.

larsiusprime:
The ANGLE support for this has already been merged into the latest lime/openfl from git, by the way, the only thing different now is my changes for forcing software on XP. In any case, if you want to rebuild lime, you will need a current set of full files from github.

jobs-git:
I have solve the ByteArray problem, we should have updated openfl too! :)

PR #647, opened 2015-12-10
updated ANGLE binaries that disable ALT+ENTER fullscreen -- this now …

larsiusprime:
…causes Alt+Enter fullscreen behavior to work properly, as openfl takes care of this instead of ANGLE


larsiusprime:
Source is here, just for reference:
https://github.com/larsiusprime/angle

PR #648, opened 2015-12-10
add "fullscreenHotkey" to project.xml window config

larsiusprime:
default value is true, if true, alt+enter for windows/linux (and ctrl+meta+F for mac) will fullscreen the app. If false, it will not when such keystrokes are pressed.

the config value is "fullscreenHotkey", and the value in xml is "fullscreen-hotkey"

I'm not particularly attached to the names/style, just the functionality.


larsiusprime:
requires:
https://github.com/openfl/openfl/pull/958

jgranick:
This should be possible in Lime, now, with `window.onKeyDown.cancel ()` when appropriate, though I would recommend (if you are using OpenFL) to use `event.stopPropagation ();` on the appropriate keyboard event to cancel it, now :smile:

PR #656, opened 2015-12-29
BitmapData.threshold() massive speedup - 95% time reduction

larsiusprime:


larsiusprime:
https://github.com/openfl/openfl/pull/984

Gama11:
How is this so much faster? Is it the same logic, just in C++?

larsiusprime:
It's the same logic written in a not-terrible way.

The original logic, using the weird Memory.select() method, took about 3000 ms to recolor a Berserker sprite in Defender's Quest (which involves multiple calls to threshold() over a fairly large sprite sheet).

The new logic, written in Haxe (and which will be executed on targets that don't support CFFI), took about 700 ms to do the same task with the same visual results.

Taking the new logic and dumping it into C++, got the task down to about 130 ms, with identical visuals.

Basically, it's faster because the original logic was *incredibly slow* so there was nowhere to go but up, and no task is more prone to being slow than iterating over every pixel and doing something to it. 

Keep in mind this is for non-flash targets; I have no idea how this compares to the native flash implementation, but I bet the CFFI path is just as fast or faster.

Gama11:
I see. Wouldn't have expected Haxe generated C++ code to perform so much worse than regular C++. I wonder if there's something specific about this logic that makes the generated version of it slow.

larsiusprime:
@Gama11 Well, I think the issue here is that we're dealing with a really performance sensitive inner loop where you want to strip away all semblances of safety and just do exactly what you know is right. IE, completely bypass the hxcpp garbage collector and everything else that goes on with haxe generated C++.

More to the point, however, I think it's not so much Haxe/hxcpp vs C++ so much as direct access to low-level data structures.

larsiusprime:
Okay this is indeed obsolete as it looks like it's been manually merged into Lime. Closing.

PR #657, opened 2015-12-30
guard CFFI STATIC_LINK anchor funcs for the optional libs

james4k:


james4k:
Just occurred to me that the commit message no longer makes sense. In the past, the register_prims functions were removed, but that was addressed since this commit was originally made, and the current commit is the result of a merge.

What this now does is make each of the libs optional, which is important for console where curl, openal, and opengl are not used.

PR #668, opened 2016-01-10
Sound cutoff fix

keyreal-code:
Sound would cut off if started with an offset due to incorrect sound length.


KirillPoletaev:
Without this fix the sound's length would not be calculated properly when playing a sound with an offset. Unless I'm missing something, the formula for calculating the sound's length should not subtract or depend in any way on the offset value.

The bug was discovered when using SoundChannel in OpenFL, tracked it down to this lime class.

I think the bug can be reproduced like this:

Take a sound file with length of 100 seconds.
Play the sound with an offset of 49 seconds.
The formula incorrectly calculates the length to be 100-49=61 seconds.
As I see it, the sound plays until it's duration matches the total length, and since we started at 49 second mark - the sound plays for 2 seconds and cuts off.

I might be wrong about this so correct me if I'm wrong, but that's the only way I could explain why the sound was getting cut off.

jgranick:
This should be fixed, thanks!

PR #669, opened 2016-01-12
Expose display device DPI

larsiusprime:
This seems like it would be useful for various platforms.

Unfortunately, it doesn't seem to be aware of display scaling in win8/10 -- it just returns "96" for everything on my system regardless of display scale. But perhaps it will work nicely on other platforms until we can figure out what's going on with Win8/10.

### DIFF
--- a/lime/system/Display.hx
+++ b/lime/system/Display.hx
@@ -20,6 +20,11 @@ class Display {
 	public var id (default, null):Int;
 	
 	/**
+	 * Pixel density of the display
+	 */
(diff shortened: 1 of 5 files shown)


larsiusprime:
Talking to an SDL guy, we determined 96 is the dummy value for when the display is not DPI aware. He passed along an article for maybe adding awareness to Windows:

http://kennykerr.ca/2014/04/07/making-a-window-dpi-aware/

Rezmason:
Oh, excellent!

PR #679, opened 2016-01-21
Renderer.readPixels fix

DanielUranga:
On cpp, readPixels returns an image with alpha=0. Can be tested with:
```haxe
package;

import flash.display.Bitmap;
import flash.display.BitmapData;
import flash.events.MouseEvent;
import flash.Lib;
import lime.app.Application;
import lime.math.ColorMatrix;
import lime.math.Rectangle;
import openfl.display.Sprite;
import openfl.events.Event;

class Main extends Sprite {

	public function new () {
		super ();
		var gfx = this.graphics;
		gfx.beginFill(0xff0000);
		gfx.drawCircle(Lib.current.stage.stageWidth/2, Lib.current.stage.stageHeight/2, 200);
		gfx.endFill();
		addEventListener(MouseEvent.MOUSE_DOWN, readPixels);
		addEventListener(Event.ADDED_TO_STAGE, readPixels);
	}

	function readPixels (_) {
		var img = Application.current.renderer.readPixels();
		var spr = new Sprite();
		spr.scaleX = spr.scaleY = 0.5;
		/*
		var mat = new ColorMatrix();
		mat.alphaOffset = 256.0;
		img.colorTransform(new Rectangle(0, 0, img.width, img.height), mat);
		*/
		spr.addChild(new Bitmap(BitmapData.fromImage(img)));
		addChild(spr);
	}

}
```

Without this fix the commented code is necessary for this test to work correctly.
This code set all alpha bytes of the resulting image to 255.

### DIFF
--- a/project/src/backend/sdl/SDLRenderer.cpp
+++ b/project/src/backend/sdl/SDLRenderer.cpp
@@ -217,6 +217,10 @@ namespace lime {
 			
 			SDL_RenderReadPixels (sdlRenderer, &bounds, SDL_PIXELFORMAT_ABGR8888, buffer->data->Data (), buffer->Stride ());
 			
+			for (unsigned char *it=buffer->data->Data()+3; it<(buffer->data->Data()+buffer->data->Length());it+=4) {
+				*it = 0xff;
+			}
+
 		}
 		
 	}


larsiusprime:
This looks like it will definitely work (set all the pixels to alpha=1 in cpp), I wonder if we could also just set the transparent flag to false in the lime image object we return.

DanielUranga:
I tried that, but the image was still transparent, maybe I was doing something wrong.

DanielUranga:
Also tried various combinations of pixels formats on the cpp side and the Haxe side, but that did not work too.

larsiusprime:
Cool, thanks for ruling those out real quick. I'll kick this up to @jgranick for now then

jgranick:
Could we try different renderers (like Cairo) to try and see if it is GL specific, or happens across the board?

DanielUranga:
The test program works OK on html5 without setting the alpha to opaque.

Also, a little unrelated: Sometimes readPixels returns a black screen (always the first time, maybe because its called as soon the program starts, can be seen on the test), happen with this fix or without it.

DanielUranga:
Works ok on flash too.

jgranick:
I'll pull this for now, if we come up with a more elegant solution, we can do that, otherwise, better to have this feature work :smile:

PR #684, opened 2016-01-26
New IOS "insecure-http-allowed" xml option

aW4KeNiNG:
New option commented [here](http://community.openfl.org/t/modify-info-plist/7270)

It avoids the problem `App Transport Security has blocked a cleartext HTTP (http://) resource load since it is insecure. Temporary exceptions can be configured via your app's Info.plist file.`

You can accept all domains with `<ios insecure-http-allowed="*"/>` or custom domains (subdomains included) with `<ios insecure-http-allowed="www.domain1.com|domain2.org"/>`


tanis2000:
@jgranick this should be merged upstream as it's a needed configuration if you want to deploy an app on the App Store.

tanis2000:
@aW4KeNiNG the new insecure-http-allowed option should be documented somewhere as well IMHO

aW4KeNiNG:
Yeah. But i don't know where i can write the information.

aW4KeNiNG:
I have seen that the PR #752 (already merged) is a similar solution, but it isn't customizable by the user like this one.

ibilon:
This PR does indeed look better, and doesn't activate `NSAllowsArbitraryLoads`,
I'll look into merging it.

aW4KeNiNG:
Synchronized with version 4.

ibilon:
Oh there's no merge conflict anymore,

@jgranick what do you think? Having `NSAllowsArbitraryLoads` as default doesn't seem good.

jgranick:
I apologize for delaying on merging a more complex nuanced solution for this. I've taken inspiration from your original approach, but implemented `<config:ios allow-insecure-http="" />` as an optional value. The tag can be defined multiple times for multiple domains, or once with "false" to disable insecure HTTP. Otherwise, a value of "*" or "true" is handled as the default, which is to allow all HTTP requests :smile:

PR #696, opened 2016-03-01
Conversion issue that prevent openfl webgl rendering

crazyjul:
This MR contains 2 small fixes that prevented correct webgl rendering in openfl 

- The conversion of image to canvas was missing a data copy. Then all canvas rendered images (swf shapes with bitmap fill,...) were empty
- Image.copyChannel was not converting the source to data. In openfl, all transparent images are composed as jpg+ alpha in png. All loaded images would then be opaque, black in transparent area


jgranick:
Thank you again for your help, I forgot to close when we fixed this earlier in the year :smile:

PR #697, opened 2016-03-01
Image: crossorigin same origin fix

Leander1P:
HTML5 Image: don't set crossOrigin to anonymous if we're likely same-origin.

Fixes an issue discussed at

  http://community.openfl.org/t/html5-401-error-when-loading/7367/6

...wherein the addition of .crossOrigin="Anonymous" causes issues with applications hosted behind e.g. BasicAuth, for browsers that follow the W3C HTML5 standard (vs the WHATWG HTML living standard, which allows submission of credentials for same-origin sites even with "anonymous").

This is not a full-fledged same-origin check; it notably ignores document.domain, and checks only for exact match between hostname, port, and protocol on a dummy link.

This also creates a temporary HTML anchor ('a') element rather than parsing out the hostname/port/protocol itself, to avoid other sources of error.

Any input is appreciated.  Notably, I'm not happy with having to create the anchor element and leave it to the GC, but it seemed the safest to have the browser be responsible for parsing out hostname/port/protocol (rather than do it ourselves and risk getting it wrong).

It also seems that the other .crossOrigin="Anonymous" (from base64 data) might not need this fix, but would appreciate a sanity check on that.

### DIFF
--- a/lime/graphics/Image.hx
+++ b/lime/graphics/Image.hx
@@ -1182,6 +1182,7 @@ class Image {
 
 			// Do something like the same-origin test, so we can continue
 			// to use credentials for assets on the same site.
+			// Note this does not take into account document.domain or similar.
 			var a:AnchorElement = Browser.document.createAnchorElement();
 			a.href = path;
 			if ( a.hostname != __origin_hostname || a.port != __origin_port || a.protocol != __origin_protocol )
--- a/lime/graphics/Image.hx
+++ b/lime/graphics/Image.hx
@@ -27,7 +27,6 @@ import js.html.AnchorElement;
 import js.html.CanvasElement;
 import js.html.ImageElement;
 import js.html.Image in JSImage;
-import js.html.Location;
 import js.Browser;
 #elseif flash
 import flash.display.BitmapData;
@@ -73,6 +72,7 @@ class Image {
 	private static var __origin_hostname:String = Browser.location.hostname;
 	private static var __origin_port:String     = Browser.location.port;
 	private static var __origin_protocol:String = Browser.location.protocol;
+	private static var __origin_test_element:AnchorElement = null;
 	#end
 	
 	public var buffer:ImageBuffer;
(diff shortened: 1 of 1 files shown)


jgranick:
Would we get a different hostname value if we check it later (such as when we first make a request to load an image, rather than detecting it statically)

If we can't find a solution (or until we do?) perhaps we could handle an empty hostname as not enabling "Anonymous" cross-origin support, returning back to better behavior for all-local projects, which is most common

Leander1P:
> Would we get a different hostname value if we check it later (such as when we first make a request to load an image, rather than detecting it statically)

My understanding is that `crossorigin` needs to be set before `src` to take effect.  It seems like this would reduce checking later to effectively the same as `onerror` retry.  But I may be reading things wrong.

> If we can't find a solution (or until we do?) perhaps we could handle an empty hostname as not enabling "Anonymous" cross-origin support,

I'll make this change and add a new commit here in the interim.

Leander1P:
> Would we get a different hostname value if we check it later (such as when we first make a request to load an image, rather than detecting it statically)

Oh -- to clarify -- what I was seeing was `a.hostname` (and `a.port` and `a.protocol`) that ends up empty, not `private static var __origin_hostname`, in this case...

jgranick:
Oh

Perhaps we can fallback to a string comparison in that case (or always use string comparison, or at least cache the anchor element we generate?)

We should be able to compare after "://" and see if it starts with the host name, and if "://" is not present, assume its local, right?

Leander1P:
> Perhaps we can fallback to a string comparison in that case (or always use string comparison,

I wanted to avoid re-implementing URL/URI parsing.  Not all protocols match the `://` pattern (see `about:blank` for a contrived example, and the various horrible `file` schemes on Windows for more concrete examples), and then I'm (hopefully justifiably) terrified of parsing stuff correctly that contains unicode chars.  https://url.spec.whatwg.org/#url-syntax has some of the complex parser flow.

If we could be sure that this would always "fail-safe" (and maybe failing in either direction in this case doesn't matter: "no crossorigin" or `crossorigin="Anonymous"` are both relatively harmless?) then maybe a naive implementation isn't an issue?

String parsing **would** fix the IE issue as well as alleviate the need for a temporary anchor element, strange URLs excepted.

>  or at least cache the anchor element we generate?)

Yeah; seems like a good idea to do this if anything like the current solution is kept (vs e.g. string parsing).  May add a commit to do that (and the empty-hostname check) before diving into string parsing.

Leander1P:
The above looks really ugly now.  I have a feeling I munged the rebase somehow.  Will investigate.  =(

Leander1P:
Okay, that should be a better version.  Still uses the DOM element, but caches it rather than creating anew, and deals with the Internet Explorer "missing `.hostname`" etc bug, as well as a few other "default port = empty string" things.

Doesn't deal with `__fromBase64`, but I'm not sure how we would, and that's a much rarer case.

Let me know if I need to squash the commits or anything like that.  Sorry it took so long to follow up!

jgranick:
Thank you so much for all your work on this, and apologies for the delay. I have just added this functionality internally, please let me know if you see any issues, or help confirm if this helps makes things work better in this use case :smile:

Leander1P:
We finally pulled a few projects up to recent OpenFL+Lime and can confirm this has worked for our usecase.  =)

For what it's worth:

We've needed similar for NetStream's __video in at least one case.  The `__isSameOrigin` method you introduced here would probably be a better fit than the workaround we did.  =)  Holler if you'd like a separate issue for this.

PR #698, opened 2016-03-02
Add window flag colors32b (and fixes)

pperidont:
- Add a new window flag that define each colors channels size to 8 bits
- Fix lime.graphics.format.PNG.fromBytes() and lime.graphics.format.JPEG.fromBytes() 
- Fix android lib path for Android armv7


kiroukou:
Is there any questions about that PR that could help merging ?

_(Not sure who is the one to ping ?)_
@larsiusprime  or @jgranick

larsiusprime:
I'm not the best person to ask about this :) But in any case, it might help to describe what problem this solves for you?

kiroukou:
Oh sorry for the ping then...:)   ( Is there a link with people who administrate a repo ?)

The window flag allow to set Opengl with full color resolution (8bits per canal) instead of the 565 which is used by default.  For some of our games, having full color resolution makes gradients really look different ;)

About the PNG/JPG, it allows a MUCH faster texture loading, since there's no need to make bytes conversion (copy).

That is for the most important adjustments.

lime/graphics/format/JPEG.hx
```diff
@@ -28,8 +29,9 @@ class JPEG {
 		var bufferData:Dynamic = lime_jpeg_decode_bytes (bytes, decodeData);
 		
 		if (bufferData != null) {
-			
-			var buffer = new ImageBuffer (bufferData.data, bufferData.width, bufferData.height, bufferData.bpp, bufferData.format);
+
+			var u8a = new UInt8Array (@:privateAccess new Bytes (bufferData.data.length, bufferData.data.b));
+			var buffer = new ImageBuffer (u8a, bufferData.width, bufferData.height, bufferData.bpp, bufferData.format);
 			buffer.transparent = bufferData.transparent;
```
crazyjul: This is required if the texture is to be used by any processing function ( copyPixels, copyChannels).
But it also breaks if decode data is false.
A fix would be 
```
var u8a:UInt8Array = null;
if(decodeData){
    u8a = new ....
}
```

This is valid for the 4 changes. Apart from that, those changes need to be merge :-)

jgranick:
Thank you for this! I apologize for the delay in merging this in. It should be available now use `<window color-depth="32" />`

Do you think we should do this by default? I wonder how many applications need the extra colors, and don't mind the extra memory.

PR #710, opened 2016-04-08
Support for 64 bits (Windows)

aW4KeNiNG:
With these changes, you can build  64 bits apps with Windows.

You need to specify "-64" in the command line. 

For example:
`lime build windows -64`

Note: `SetClassLong` is changed to `SetClassLongPtr` for 32/64 bits compatibility (more info [here](https://msdn.microsoft.com/es-es/library/windows/desktop/ms633589%28v=vs.85%29.aspx))


aW4KeNiNG:
Sync with version 4.

intoxopox:
Is Windows64 still an active target? I can't seem to build for it.

PR #718, opened 2016-04-23
[DONT MERGE] Fix openURL on linux

ibilon:
This isn't needed, the command returns immediately,
and keeping it does:
```
xdg-open: unexpected argument '&'
Try 'xdg-open --help' for more information.
```


ibilon:
So the "&" is needed actually, just have to figure out a way to send it to the shell and not as an argument.

PR #719, opened 2016-04-23
Load assets from next of the executable

ibilon:
Fix https://github.com/openfl/openfl/issues/723

Allows the usage of `setCwd` without breaking the asset loading.

### DIFF
--- a/templates/haxe/DefaultAssetLibrary.hx
+++ b/templates/haxe/DefaultAssetLibrary.hx
@@ -18,6 +18,7 @@ import lime.utils.UInt8Array;
 import lime.Assets;
 
 #if sys
+import haxe.io.Path;
 import sys.FileSystem;
 #end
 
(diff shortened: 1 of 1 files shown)


vroad:
This change broke Android build. (Android loads assets usually from apk.)

ibilon:
Hmm android, they use the `fromFile` functions too?
Maybe replace `#if sys` with `#if desktop`? (if we have that)

vroad:
On Android, DefaultAssetLibrary tries to load from path like this:
"/storage/emulated/0/Android/data/org.openfl.lime.samples.herokushaders/files/.__override__/assets/6147.1.frag"

And if you pass null to GL.shaderSource, that will make app crash.
Of course asset is not there. I've tried replacing with that and worked.

ibilon:
668f59ae1cf2228e3989fc85b727ca3a0f4a1484 should do it.

PR #729, opened 2016-05-13
Added controllable caching

ShaheedLegion:
Added a mechanism whereby users can override the default caching behavior for their resources.

This fixes the issue mentioned here: http://community.openfl.org/t/how-to-disable-cache-breaking-string/7432/4


ibilon:
Haven't looked at the changes themselves, just superficially,
* the changes on the submodules shouldn't be there, could you rebase on latest develop/submodules?
* the indentations on the templates doesn't match the rest of the file
* the trace in the preloader, on purpose or left over from debugging?

Thanks for contributing.

PR #630, opened 2015-11-18
Add <prebuild /> and <postbuild /> tags

player-03:
I finally got around to implementing the functionality I described in issue #612. For convenience, here's the relevant part of that post:

### `<postbuild />`
The `<postbuild />` tag defines a command to be run after each successful build:

    <postbuild command="echo Android build finished!" if="android" />

If you prefer, you can execute a Haxe file. This requires a file with a static main method.

    <postbuild haxe="PostBuild.hx" />
    
    <postbuild command="haxe -main PostBuild.hx --interp" />

You can use the `open` attribute to open a file, folder, or url. This uses a system-specific command (`start` for Windows, `open` for Mac, or `xdg-open` for Linux) to find the right program.

    <postbuild open="Export/mac" if="mac" />
    <postbuild open="http://www.example.com/" />

To include spaces within arguments, use `<arg />` tags:

    <postbuild command="cp">
        <arg>sample/path/with spaces/file.txt</arg>
        <arg>another path/with spaces/file.txt</arg>
    </postbuild>

If you define multiple `<postbuild />` tags, they'll be executed in order:

    <postbuild command="mkdir folder" />
    <postbuild command="touch folder/file.txt" />
    <postbuild open="folder/file.txt" />

### `<prebuild />`
The `<prebuild />` tag defines a command to be run before building. The syntax is exactly the same as for `<postbuild />`.

PR #731, opened 2016-05-16
Add <prebuild /> and <postbuild /> tags

player-03:
(Submitting #630 to the correct branch.)

This allows users to run arbitrary commands before or after the build process. Since these are xml tags, users can choose to run different commands for different targets, or different build types.

Sample use cases:
- A haxelib that automatically runs some code during a mobile build. (I'd definitely use this in [ExtensionBoilerplate](https://github.com/player-03/ExtensionBoilerplate) if I could.)
- Delete a specific file before a clean build.
- Save a copy of an important file after a release build.
- Make sure the output file meets certain criteria.
- Record when the build started and finished in an ongoing log file.

Documentation:
### `<postbuild />`
The `<postbuild />` tag defines a command to be run after each successful build:

    <postbuild command="echo Android build finished!" if="android" />

If you prefer, you can execute a Haxe file. This requires a file with a static main method.

    <postbuild haxe="PostBuild.hx" />
    
    <postbuild command="haxe -main PostBuild.hx --interp" />

You can use the `open` attribute to open a file, folder, or url. This uses a system-specific command (`start` for Windows, `open` for Mac, or `xdg-open` for Linux) to find the right program.

    <postbuild open="Export/mac" if="mac" />
    <postbuild open="http://www.example.com/" />

To include spaces within arguments, use `<arg />` tags:

    <postbuild command="cp">
        <arg>sample/path/with spaces/file.txt</arg>
        <arg>another path/with spaces/file.txt</arg>
    </postbuild>

If you define multiple `<postbuild />` tags, they'll be executed in order:

    <postbuild command="mkdir folder" />
    <postbuild command="touch folder/file.txt" />
    <postbuild open="folder/file.txt" />

### `<prebuild />`
The `<prebuild />` tag defines a command to be run before building. The syntax is exactly the same as for `<postbuild />`.


james4k:
Just to get a discussion going here, I think the counterargument to this is that you might instead invoke `lime build` as part of a greater build process, even if that is just batch/shell scripts.

Can you describe how you would use this in ExtensionBoilerplate?

player-03:
Good question. I did some digging, and it turns out my plan wouldn't work because Lime doesn't pass information about what project imported what other project. If it did pass that information, here's what I'd add to ExtensionBoilerplate's import.xml:

```xml
<prebuild command="haxelib run ExtensionBoilerplate ${importedFrom}" if="mobile" />
```

`importedFrom` is a hypothetical variable containing the directory of the source project, or the project name if it's registered as a haxelib. After a few more changes to ExtensionBoilerplate, using it would be as easy as adding this to your extension's import.xml:

```xml
<haxelib name="ExtensionBoilerplate" />
```

That way, ExtensionBoilerplate would automatically run before any mobile build. It would recompile your C++ code if necessary, so any changes you make will automatically make it into the app. Sure, you could write your own build script for this, but it would be so much easier to type "`<haxelib name="ExtensionBoilerplate" />`" and be done with it.

lime/tools/helpers/CommandHelper.hx
```diff
@@ -0,0 +1,56 @@
...
+		
+		for (c in commands) {
+			
+			Sys.command(c.command, c.args);
+			
+		}
+		
+	}
+	
+	public static function openFile (file:String):CLICommand {
```
ibilon: We already have `ProcessHelper.openFile` https://github.com/openfl/lime/blob/75bb6905af1b64b01af68892134afc00550846e0/lime/tools/helpers/ProcessHelper.hx#L21

ibilon:
Code looks good, I'll give it a try today and if working merge it.

ibilon:
Actually had time right away,
no issue found.

ibilon:
Thanks!

ibilon:
@player-03 For the missing information, like the hypothetical variable `importedFrom`, could you open an issue with what you need?

player-03:
Done! See #758.

PR #736, opened 2016-05-18
Fix to audio looping/Neko OpenAL support.

mauvecow:
Two changes in this PR:
- Looping buffers in an AudioSource is now handled by queueing them, instead of a timer. This creates sample-accurate looping. Limited numbers of buffers are used to allow for absurd loop values with no cost.
- Changes val_array_int and val_array_float references in OpenALBindings.cpp to temporary buffers, to fix compatibility with neko.
  - Possible improvement, if more annoying to maintain: Keep the original val_array_* code around and only use these as a fallback for when they fail.

Some other things noted:
- This change breaks the starting offset on loop, and will always reset to the start of the buffer. OpenAL does not support looping offsets, to my knowledge.
  - Suggestion: When buffer streaming gets added, allow streaming to be done on a 'view' of an existing AudioBuffer.
- AudioSource's looping timer does not work correctly if the current playback time is set to a non-zero value.
  - Suggestion: This could be fixed by forcing it to restart the timer on a loop. Did not perform this change, but I can on request.

### DIFF
--- a/lime/audio/AudioSource.hx
+++ b/lime/audio/AudioSource.hx
@@ -138,7 +138,7 @@ class AudioSource {
 				}
 				
 				id = al.genSource ();
-				al.sourcei (id, al.BUFFER, buffer.id);
+				al.sourcei (id, al.BUFFER, 0);
 			
 			default:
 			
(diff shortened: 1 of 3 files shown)


mauvecow:
Closing this due to requiring more extensive work on audio streaming before a merge. Moved neko changes to #736.

PR #744, opened 2016-05-22
Set cpp targetFlag correctly from CommandLineTools

vroad:


jgranick:
Some old NME projects expected `nme test cpp` (as opposed to `nme test neko`), so this was just a shortcut. I'm not sure we can assume that dynamically added targets are going to be C++ based, so I don't think we can modify the "default" in this case

PR #750, opened 2016-05-31
Allow to include TBD dependencies in XML

fbricker:
This allows lime extensions to add things like:
<dependency name="libz.tbd" />

This is the only missing extension, since dylib packages are deprecated and replaced by TBD on xcode 7+


jgranick:
Thanks :smile:

PR #756, opened 2016-06-06
Do not exit with Sys.exit()

vroad:


ibilon:
The return value of main isn't used as system exit code.

PR #758, opened 2016-06-10
Access more information from project.xml.

player-03:
Templates can access `<app />`, `<meta />`, and `<window />` values, but for some reason, project.xml can't. In project.xml, all you can access are `<set />` or `<setenv />` values.

I fixed this, but alone it isn't enough. What if a library project wants to get the window dimensions? Libraries don't set the dimensions, the main project does. And since the main project doesn't pass this information down, the library project can't access them.

To fix this, I added an option to access a parent project - just add `parent:` in front of your variable. For instance, `${parent:width}` and `${parent:height}` return the parent project's window dimensions. This is recursive, so `${parent:parent:width}` and `${parent:parent:height}` return the grandparent project's dimensions.

As a side-effect, you can test whether a parent project exists using `if="${parent:}`.

Finally, you can now access a project's directory using `${directory}`. This is fully compatible with the "`parent:`" syntax.


ibilon:
Not sure what you mean by "parent project".

player-03:
When one project imports another, I call the former the "parent" and the latter the "child."

If you have a different term, I'd be happy to use that instead.

jgranick:
After you put forward your pull request, I added support for `HXProject` object properties, such as `${app.path}`, `${meta.title}` and `${window.width}`. 

I'm not sure if I want to create a parent/child relationship between `HXProject` instances, or require that a `HXProject` has an associated project directory. For now, I have added `${projectDirectory}` to XML, but I think some things may be better features for HXP.

I know HXP has run into some issues in the past, so I hope we can build more ways to improve testing and stability in the HXP project format :smile:

Thank you so much for all of your help!

player-03:
Object access looks good!

As for the project directories, I guess I can do without a parent/child relationship. People can still pass that information to a library using `<set name="parentDirectory" value="${projectDirectory}" />`.

I realize HXProjects could already call `Sys.getCwd()`, but that only helped for the root project. Included projects, like haxelibs, are loaded by [`HXProject.fromPath()`](https://github.com/openfl/lime/blob/develop/lime/project/HXProject.hx#L570), which only checks for XML files.

Issue #759, opened 2016-06-14
ImageDataUtil.fillRect alpha-premultiplies repeatedly


MrCoft:
In OpenFL, the BitmapData constructor produces a "gradient" instead of filling it with color on cpp and neko targets, it works properly in flash and html5.
```haxe
stage.color = 0x0000ff;
stage.addChild(new Bitmap(new BitmapData(20, 20, true, 0xfeff0000)));
```
![image](https://cloud.githubusercontent.com/assets/12239922/16030969/8009da7c-31f6-11e6-81f9-e31055e9f81a.png)
When the alpha is 0x80 instead of 0xfe, it is only about 3 pixels long - it seems the color is alpha premultiplied repeatedly.

azrafe7:
@MrCoft, can you please test if the proposed change fixes the issue for you?

MrCoft:
I wasn't able to run it with the normal nor nightly openfl build and I don't know how to test this in lime alone. But just changing the two lines in ImageDataUtil.hx did not fix the issue.

azrafe7:
You can try this:

 - apply the changes to both `ImageDataUtil.hx` and `ImageDataUtil.cpp`
 - run `lime rebuild windows'

MrCoft:
The haxelib lime's `project` folder is in `/legacy` and looks differenty - it does not have an `ImageDataUtil.cpp`, and I could not find the CFFI function. The loop in the haxe file produces correct color if I force the branch.
I guess it works, just lime got restructured so I can't confirm it.

ibilon:
You need the github version to modify cpp files and rebuild the ndll.

MrCoft:
After building it properly using github versions fillRect works.

pozirk:
There is still a problem on Windows target, but with slightly different example:

```
var bmpd:BitmapData = new BitmapData(100, 100, false, 0xffffff);
var bmp:Bitmap = new Bitmap(bmpd);
addChild(bmp);
```

I have white square as expected on Flash and Android, but nothing on Windows.
Any quick fix for that?

pozirk:
There is still a problem on Windows target, but with slightly different example:

```
var bmpd:BitmapData = new BitmapData(100, 100, false, 0xffffff);
var bmp:Bitmap = new Bitmap(bmpd);
addChild(bmp);
```

I have white square as expected on Flash and Android, but nothing on Windows.
Any quick fix for that?

PR #761, opened 2016-06-16
Fix fillRect behaviour when using alpha on neko/cpp (see #759 and #631)

azrafe7:
Previously fillColor would get multiplied repeatedly in a loop, this PR aims to fix it by computing the premultiplied alpha value once, and using it in calls to fillColor.writeUInt8 with the premultiplied arg set to false.

See #759 and #631

### DIFF
--- a/lime/graphics/utils/ImageDataUtil.hx
+++ b/lime/graphics/utils/ImageDataUtil.hx
@@ -333,6 +333,7 @@ class ImageDataUtil {
 			
 			var format = image.buffer.format;
 			var premultiplied = image.buffer.premultiplied;
+			if (premultiplied) fillColor.multiplyAlpha();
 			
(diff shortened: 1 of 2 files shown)


azrafe7:
Hey @jgranick, was scanning ImageDataUtil for occurrences of the same issue and it seems that floodFill() suffers from the same problem.

jgranick:
Maybe most tests were done without alpha

azrafe7:
No worries, let's fixed that too...

Is it ok if I send another (almost equal) PR for that, or you want to address it?

jgranick:
Go for it, the UsingBitmapData sample has a floodFill test, if you need something to use to confirm it, we could use a semi-transparent fill value. The help is very appreciated

PR #769, opened 2016-06-29
Added ProgressEvent feedback for HTML5 target

TomByrne:
This is required by an OpenFL change I will push soon.
I have kept the original method of image loading in html for requests made without a ProgressEvent listener. This is for two reasons:
- The old method will probably perform a little better in some browsers, which will decode the image in a separate thread.
- The new method should be tested in more browsers before it is considered the standard (adding a ProgressEvent handler is effectively an opt-in).

Older browsers might require the XMLHttpRequest response to be manually from a string to an ArrayBuffer. I haven't done these tests and I'm not sure what the backwards compatibility policy is.

### DIFF
--- a/lime/graphics/Image.hx
+++ b/lime/graphics/Image.hx
@@ -26,7 +26,10 @@ import lime.utils.UInt8Array;
 import js.html.CanvasElement;
 import js.html.ImageElement;
 import js.html.Image in JSImage;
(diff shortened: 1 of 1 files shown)


jgranick:
I hope to have this done soon, I'll use your code as a reference and see where we are now that we've completed the AssetLibrary refactor for preload progress :smile:

https://github.com/openfl/openfl/issues/1435

PR #777, opened 2016-07-10
Add Gradle support

player-03:
Projects tested:

- DisplayingABitmap - success
- AddingAnimation - success
- AddingText - success
- PlayingSound - success
- SimpleOpenGLView - blank screen, but that happens with Ant too
- BunnyMark - success
- PiratePig - success
- HerokuShaders - success


jgranick:
Thank you so much for all of this hard work!

 * Are there known differences or possible regressions when using this code? Would ordinary Lime and OpenFL projects be expected to build as-is? What about existing native extensions?
 * Does using this code affect what versions of Android you are able to target?
 * Should we handle GRADLE_HOME, as we handle ANT_HOME?
 * Are there any other steps required to set up a working build environment?

Thank you :smile:

jgranick:
Will users need to change their configuration to point to the exact version of Gradle they have installed? Can you use a newer Gradle install, even if a project requests an older version?

player-03:
> Are there known differences or possible regressions when using this code? Would ordinary Lime and OpenFL projects be expected to build as-is? What about existing native extensions?

I'm not aware of any regressions. Existing projects should be able to build as-is, unless they rely on a custom Ant `build.xml` file. In that case, they'll have to make a simple `build.gradle` file with the line "`ant.importBuild 'build.xml'`."

On the other hand, extensions that rely on Ant should be fine. This is because extensions are treated as "subprojects," and I've added [custom compilation rules](https://github.com/player-03/lime/blob/gradle/templates/android/template/build.gradle#L29-L59) for compatibility with the old extension structure. (Though if they changed the structure and the extension's `build.xml` file, all bets are off.)

> Does using this code affect what versions of Android you are able to target?

No, and in fact it doesn't change how you specify a target (or minimum) sdk version. As seen [here](https://github.com/player-03/lime/blob/gradle/templates/android/template/gradle.properties#L19) and [here](https://github.com/player-03/lime/blob/gradle/templates/android/template/app/build.gradle#L10), the `ANDROID_TARGET_SDK_VERSION` variable is passed all the way through to Gradle.

I forget why I [hard-coded `ANDROID_BUILD_TOOLS_VERSION=23.0.3`](https://github.com/player-03/lime/blob/gradle/templates/android/template/gradle.properties#L22). I think there might have been a bug in version 24, or maybe 23.0.0. So the user will need that particular version installed.

> - Should we handle GRADLE_HOME, as we handle ANT_HOME?
> - Are there any other steps required to set up a working build environment?
> 
> Will users need to change their configuration to point to the exact version of Gradle they have installed?

No to all of the above.

In fact, the user won't need to install Gradle manually. As @restorer suggested, I included a [Gradle wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html) in the Android templates. This wrapper consists of the two script files in the [root folder](https://github.com/player-03/lime/tree/gradle/templates/android/template), plus a jar and some settings under [gradle/wrapper/](https://github.com/player-03/lime/tree/gradle/templates/android/template/gradle/wrapper). The script files (one for Windows, one for Mac and Linux) simply execute the jar, and the jar downloads and installs Gradle, choosing the version set in `gradle-wrapper.properties`.

To test this, I deleted all copies of Gradle from my machine, and then ran a build. The wrapper downloaded Gradle without any trouble, and the build finished without any further input from me.

@restorer [claimed](http://community.openfl.org/t/switching-to-gradle/7703/14) that any 2.X version is fine, so I chose Gradle 2.10 (defined [here](https://github.com/player-03/lime/blob/gradle/lime/tools/platforms/AndroidPlatform.hx#L314)). He also said there's usually no need to keep Gradle up-to-date, as long as you're using the right major version.

> Can you use a newer Gradle install, even if a project requests an older version?

[The value in `gradle-wrapper.properties`](https://github.com/player-03/lime/blob/gradle/templates/android/template/gradle/wrapper/gradle-wrapper.properties#L5) will override everything else, no matter which version(s) of Gradle are currently installed. Fortunately, you can set that value [from project.xml](https://github.com/player-03/lime/blob/gradle/lime/project/ProjectXMLParser.hx#L1567-L1569).

jgranick:
Merged!

A very big thank you :smile:

player-03:
Yay! And you're welcome, of course. It's nice to have this ready for public use.

Let me know if any issues come up.

jobs-git:
Hi, can we still use ant as before? I am afraid it would alianate existing ant users.

jobs-git:
I think there should be an option to use ant or gradle. 

Like

openfl -gradle

so that transition will be smooth?

jgranick:
@jobs-git 

Google has deprecated Ant support, it is only a matter of time before it may break

The implementation here uses a Gradle wrapper which can automatically downloads any missing packages, this in effect removes one dependency (Ant) but does not add a new one in its place, Java and the Android NDK+SDK should be required to build.

player-03:
@jobs-git Are you having trouble with the transition? It should just work for most projects.

mastef:
As said before @player-03 kudos again on a well done job! 👍 

We just were debugging on slack as somebody ran into an issue with a fresh dev build : 

a) Instead of calling `gradlew`, runCommand needs to be called in the AndroidHelper.hx with parameter `./gradlew` if on Linux, and `gradlew.bat` if on Windows ( [same case as already handled for adbName](https://github.com/openfl/lime/blob/0aac32f55120792c824d920bcedc24f4f759e1a3/lime/tools/helpers/AndroidHelper.hx#L120-L140) ), otherwise the user will get a `sh: gradlew: command not found`. This error _might not_ happen to users who have gradle installed on the system globally

b) When gradlew is copied over to the project directory it loses its' permissions, so `chmod +x` needs to be set before executing, otherwise it'll throw a permission denied error

@jgranick is working already on [a commit covering both](https://github.com/openfl/lime/commit/978fac419fe77ef14c3cb0f09191e67da0f2f8a2), just mentioning here the issue that caused it :)

player-03:
I guess I didn't test it on enough different machines. Oops!

player-03:
> I forget why I hard-coded `ANDROID_BUILD_TOOLS_VERSION=23.0.3`.

Figured out why; it's because of errors like this:

```text
Exception in thread "main" java.lang.UnsupportedClassVersionError: com/android/dx/command/Main : Unsupported major.minor version 52.0
```

And according to [this](http://stackoverflow.com/a/36027375/804200), the fix is to go back a version. However, as described [here](http://stackoverflow.com/a/35866015/804200), an equally-valid solution is to update to Java SE 8.

I updated Java after hard-coding version 23.0.3, and I guess that's why I couldn't reproduce the error.

jgranick:
Does a newer build tools work, or is Java 8 now a requirement?

player-03:
I haven't tried Java 9 yet, sorry.

Either way, maybe it's time to add a config value for this. Then if someone wants to use a different Java version, all they have to do is update project.xml. A more advanced option is to call `java -version` and set the build tools version based on the result.

jgranick:
@player-03 

I'm inclined to believe that if the tools will require Java 8, and (as we've seen) targeting current Android versions requires a newer NDK, then better to force the upgrade all at once, then to half-upgrade and run into trouble when we need to push up the minimum versions for some fix or new OS.

We could call `java -version` (perhaps) at least to print an error that people can understand if we go all Java 8

player-03:
Sounds good.

If it helps, it's possible to use Gradle to examine the reason for failure, and then print an error message based on that. [Here's one example](https://github.com/player-03/google-play-services/blob/master/google-play-services/build.gradle#L62-L87), even though it might be a little more complex than necessary. (Instead of calling `getCause()` a bunch of times, I could have just called `getMessage()` and checked whether it contained the substring I was looking for.)

restorer:
What about something like that (in top-level `build.gradle`)?

```java
if (VersionNumber.parse(project.ANDROID_BUILD_TOOLS_VERSION.replaceAll('^([0-9.]+).*$', '$1')) > VersionNumber.parse('23.0.3')
    && VersionNumber.parse(System.getProperty('java.specification.version')) < VersionNumber.parse('1.8')
) {
    throw new RuntimeException("Please use the Android SDK Manager to update the build tools.\nAlternatively, set a different version in project.xml.");
}
```

jobs-git:
It maybe better to still keep the Ant support as some might encounter issues that was previously working on Lime.

ibilon:
@jobs-git having both is probably hard, and their project will stop working too if they upgrade the android sdk (ndk?)

jobs-git:
Around **~63%** of Android phones were still on versions under 21,  I think good deal of android apps on this version were build with Ant. On the other hand there are **~34%** users of Android Lollipop that is still on API lvl 21-22. so in many of these projects, that is about ~97% Android apps, works nice with Ant, to date only **~2.4%** are using the latest android where Gradle works great.

**I think, it would be in the best Interest of this Code to support ~97% of Android phones to maximize the extent at which Lime/OpenFl is used, of course, I maybe wrong, and Lime developers always got the last statement on this.**

It is always easy to argue that by the time the user base of Android at which Gradle works nice became noticable Lime/OpenFl is ready for it and we can all forget Ant, its been 2015 since Marshmallow was released and it garnered ~2% user base.  Here is some stats last March 8, 2016
http://www.androidcentral.com/tag/versions

restorer:
@jobs-git build tools version != target sdk version != min sdk version

You can use latest build tools to build app even for Android 1.5

jobs-git:
I love updates, I just gave a friendly suggestion that Ant maybe a good fallback.

player-03:
> [...] the latest android where Gradle works great.
> [...] it would be in the best Interest of this Code to support ~97% (or maybe less, but large enough number) of Android phones [...]

Android devices don't run Gradle, nor do they need to. Ant and Gradle are build tools that run on your computer, and once the APK file is built, their job is done.

If you're worried about Gradle building an incompatible APK, don't be. Building with Gradle will have no effect on compatibility. If your app required Android 6.0, it will still require 6.0. If your app was compatible with everything since Android 2.4, it will still be.

To control which devices your app runs on, set `<android min-sdk-version="`[`[API level]`](https://developer.android.com/guide/topics/manifest/uses-sdk-element.html#ApiLevels)`" />` in project.xml. This setting works just as well with Gradle as it did with Ant.

jobs-git:
I may have translated it wrongly, but yhea, some still use ANT to build apps that maybe targeting 97% android phones.

player-03:
You can also use Gradle to build apps that target 97% of Android phones.

PR #799, opened 2016-08-24
Initial C# implementation

vroad:
Hello World example from lime-samples should work with this, at least on Windows. This one still needs non-standard hxcs, because official one does not support setting CPU architecture and Android projects. Also, this may not work with Haxe v3.2.x. I would recommend using the latest version of Haxe found in haxe git builds.

https://github.com/vroad/hxcs
http://hxbuilds.s3-website-us-east-1.amazonaws.com/builds/haxe/index.html

`lime test cs`
`lime test android-cs`


lime/tools/helpers/CSHelper.hx
```diff
@@ -0,0 +1,301 @@
...
+	
+	inline public static function buildSln (path:String, slnPath:String, task:String = null) {
+		
+		buildCSProj (path, slnPath, task);
+		
+	}
+	
+	public static function buildCSProj (path:String, csprojPath:String, task:String = null) {
+		
+		var msBuildPath = "C:/Program Files (x86)/MSBuild/14.0/Bin/MSBuild.exe";
```
ibilon: That looks way too hardcoded, why not just `MSBuild.exe`, is it not in path?
Also linux mono.

vroad: Probably I should use `ProcessHelper.runCommand ("", "msbuild", params)` instead?

vroad: `msbuild` is not in path on windows unless you set variables with `vcvars32.bat`. And currently `msbuild` is not used for building windows executable. It's currently used only for building Android apk.
 
I could write some code that finds `msbuild ` from environmental variables or settings.

Linux equivalent would be `xbuild`, but you could also use `msbuild` on linux as well as it's opensource now.

https://github.com/Microsoft/msbuild

ibilon:
The desktop part looks good, will have to test it's interesting :)

For android I have more reserve, looks like gradle is included, likely you started working on that before we had support for it in lime.
Also a bit strange to have a special template folder for it.

vroad:
Yes, it was created before gradle is included. Maybe we could reuse code/template from that, but can I enable/disable target specific code (hxcpp/cs) easily using flags?

There are many differences that making code sharing a bit difficult. 

- `SDLActivity` should start running C# code instead of hxcpp code.
- You need to create a C# wrapper to use Java classes from C# (such as `GameActivity`).
- On Xamarin.Android, `MainActivity` should be written in C#.
- `AndroidManifest.xml` in android-cs directory is not directly used. Xamarin's build tool automatically inserts activity xml element, based on attributes found in `MainActivity.cs`. https://github.com/vroad/lime/blob/2d3aa8729427c82c161bbdebc2d3ac4ebe79a765/templates/android-cs/template/Sources/MainActivity.cs#L9-L11

PR #802, opened 2016-08-31
Build directory

player-03:
Windows has a hard limit (240 characters) on the length of a file path. Gradle uses deeply nested files, and some libraries (looking at you, Google Play Game Services!) include extremely long file names.

Gradle's paths are still under the limit, but Lime's directory structure `Export/android/release/bin` is enough to push it over in some cases.

Fortunately, Gradle lets you put its build folders somewhere else; ideally somewhere with a very short filepath. For instance, by setting `<android gradle-build-directory="C:/Build" />`, you'll dramatically reduce your file path length without having to move your project. (If you don't set `gradle-build-directory`, everything works exactly as before.)

### DIFF
--- a/lime/project/ProjectXMLParser.hx
+++ b/lime/project/ProjectXMLParser.hx
@@ -1573,6 +1573,10 @@ class ProjectXMLParser extends HXProject {
 									
 									config.set ("android.gradle-version", value);
 								
+								case "gradle-build-directory":
+									
+									config.set ("android.gradle-build-directory", value);
+								
 								default:
 									
 									name = formatAttributeName (attribute);
--- a/lime/tools/platforms/AndroidPlatform.hx
+++ b/lime/tools/platforms/AndroidPlatform.hx
@@ -187,7 +187,8 @@ class AndroidPlatform extends PlatformTarget {
 			
 		}
 		
-		var apkPath = FileSystem.fullPath (targetDirectory) + "/bin/app/build/outputs/apk/";
+		var apkPath = FileSystem.fullPath (targetDirectory) + "/bin/app/build";
+		apkPath = project.config.getString ("android.gradle-build-directory", apkPath) + "/outputs/apk/";
 		var apkSuffix = "-" + build + ".apk";
 		
 		File.copy (apkPath + "app" + apkSuffix, apkPath + project.app.file + apkSuffix);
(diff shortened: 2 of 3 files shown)


player-03:
The actual path is `buildDir` plus the project title. So if you add "`<android gradle-build-directory="C:/Build" />`" to HandlingMouseEvents, the build files will go in `C:/Build/HandlingMouseEvents`.

I could set it up to go in `C:/Build/HandlingMouseEvents/debug` or `/release`. Is that worth adding a few characters for?

jgranick:
Reconsidering file structure:

 * What do you think of "android-release" (or just going back to plain "android") instead of "android/release"? I know this is only one folder level, but might help
 * It's minor, but what about "libs" instead of "deps" for Android library dependency projects?
 * How important is the Gradle build directory? What if I used a temporary system folder instead of leaving it in the project directory?
 * How short of a path do we need? With all the Google Play Services stuff, how long of a path does that end up being?

player-03:
> What do you think of "android-release" (or just going back to plain "android") instead of "android/release"?

Windows counts characters, not folder levels, so "android-release" would be exactly the same. Going back to "android" would help, but if someone wants to store a project in "`C:/Users/UserName/CompanyName/Category/ProjectName`," they're still going to hit the limit no matter how many characters we remove here.

But why would someone who's that picky about folders want a build directory like this? The reason is that everything in the build directory is entirely disposable, and will most likely never need to be checked into version control. (Like OpenFL`s `Export` directory, but more so.)

> It's minor, but what about "libs" instead of "deps" for Android library dependency projects?

Gradle refers to them as both "dependencies" (less specific) and "subprojects" (more specific). I'd say "deps" is fine.

> How important is the Gradle build directory?

It helps fix a compiler error on Windows and does almost nothing on Mac and Linux. There are other ways to fix the error (generally involving moving your project folder itself), but this was the simplest option I found.

> What if I used a temporary system folder instead of leaving it in the project directory?

As in, setting `buildDir = "::TEMP_DIRECTORY::/::APP_FILE::/${project.name}"`? That would work too, unless the user has a ridiculously long name or moved their temp folder somewhere weird.

> How short of a path do we need? With all the Google Play Services stuff, how long of a path does that end up being?

It has to be under 240 characters. The exact path length varies based on lots of factors, but here's the path that caused the error in my tests:

```text
C:\Users\Joseph\Project\DisplayingABitmap\Export\android\release\bin\deps\google-play-services\build\intermediates\exploded-aar\com.google.android.gms\play-services-base\9.4.0\res\drawable-mdpi-v4\common_google_signin_btn_text_light_disabled.9.png
```

Here's the same path with `buildDir` set to a couple different places:

```text
C:\Build\DisplayingABitmap\google-play-services\intermediates\exploded-aar\com.google.android.gms\play-services-base\9.4.0\res\drawable-mdpi-v4\common_google_signin_btn_text_light_disabled.9.png

C:\Users\Joseph\AppData\Local\Temp\DisplayingABitmap\google-play-services\intermediates\exploded-aar\com.google.android.gms\play-services-base\9.4.0\res\drawable-mdpi-v4\common_google_signin_btn_text_light_pressed.9.png
```

The first is 247 characters, the second is 194 characters, and the third is 219 characters. However, you should add at least three to each of those numbers, because these were all in the `drawable-mdpi-v4` folder, and there's an `drawable-xxhdpi-v4` folder with all the same files.

jgranick:
Also, have you seen if newer Gradle versions work? I wonder if they have done anything to rein in the long paths since 2.1.0?

player-03:
I don't think the Gradle devs consider this their problem, since it's caused by a Windows limitation, and the files causing the error belong to Google Play Services. I searched around some more, but no one's suggested updating Gradle as a solution to this.

The Google Play Services devs don't seem to care either; these files have been there ever since version 3.1.36 (the current version is 9.6.1). The images changed each time Google released a new logo, but the paths stayed the same.

jgranick:
I believe I have this merged, but let me know if you find any issues. Thank you!

PR #805, opened 2016-09-04
Correct window.title setter

tienery:
Simple typo that needed correcting, so that people can have the ability to change the title of the window if need be.


ibilon:
Thanks.

PR #812, opened 2016-09-12
FLASH_PLAYER_EXE typo fix (option 1)

player-03:
This code checks for the existence of `FLASH_PLAYER_EXE` but copies `SWF_PLAYER` instead. Also, fixing this typo isn't enough, since [`FlashHelper` looks in the system environment variables](https://github.com/openfl/lime/blob/develop/lime/tools/helpers/FlashHelper.hx#L959), not `project.environment`, for a value named `FLASH_PLAYER_EXE`.

I've created two pull requests for you to choose from:

1. Deletes the code. #812
2. Fixes the typo and calls `Sys.putEnv()`. #813

Personally, I suggest option 1. No one will ever need to set `FLASH_PLAYER_EXE` from project.xml, since they can get exactly the same result by setting `SWF_PLAYER`. This code is just taking up space.

### DIFF
--- a/lime/project/ProjectXMLParser.hx
+++ b/lime/project/ProjectXMLParser.hx
@@ -149,10 +149,6 @@ class ProjectXMLParser extends HXProject {
 			
 			environment.set ("SWF_PLAYER", defines.get ("SWF_PLAYER"));
 			
-		} else if (defines.exists ("FLASH_PLAYER_EXE")) {
-			
-			environment.set ("FLASH_PLAYER_EXE", defines.get ("SWF_PLAYER"));
-			
 		}
 		
 		defines.set (Std.string (target).toLowerCase (), "1");


ibilon:
IIRC doing something like `SWF_PLAYER=myflash.exe lime test flash` works right?
If so yeah option one looks better.

player-03:
That's Bash syntax, and you'd have to use a special shell to make it work on Windows, but the point stands.

Since `FLASH_PLAYER_EXE` is the environment variable and `SWF_PLAYER` is the value in `project.defines`, you'd run `FLASH_PLAYER_EXE=myflash.exe lime test flash`.

Note that `FLASH_PLAYER_EXE=myflash.exe lime test flash` will still work even after you accept option 1. After all, option 1 doesn't modify `FlashHelper.hx` at all, and `FlashHelper.hx` is what calls `Sys.getEnv("FLASH_PLAYER_EXE")`.

ibilon:
> Since FLASH_PLAYER_EXE is the environment variable and SWF_PLAYER is the value in project.defines

That's good, no need to define the env var from inside the project's define when you have `SWF_PLAYER` indeed.

PR #830, opened 2016-10-20
Never allocate CFFI utility classes on heap to avoid GC issues

vroad:
Currently Bytes is using AutoGCRoot to avoid values from being collected, but it's not necessary if it's on stack. GC values needs to be on stack to be visible from the collector, or you should use AutoGCRoot to protect them. Older version of lime had issues because some of CFFI utility classes allocated GC values on heap, and values were not protected with AutoGCRoot.

I changed the code to allocate instance of those classes on stack, rather than protecting GC values with them. This would probably solve the problem with garbage collector, and allows us to write code in much cleaner way (one that does not need many `#if !cs`). And this would make CFFI slightly faster, because we no longer need to allocate many small class instances on heap.


vroad:
Why are Bytes and ArrayBufferView allocated on heap in current implementation?? Is there any reason to do this?

jgranick:
I have had problems with HXCPP running garbage collection while we are still inside of CFFI, discarding objects that were allocated only on the stack, while they are still in use. It only occurs with custom objects (like Bytes), so using the heap is necessary to ensure stability on HXCPP. It's unfortunate this makes the code messier for C#

PR #832, opened 2016-10-22
Implement lime.ui.Haptic.vibrate(....) feature

fbricker:
Compiling and tested on Android.
Compiling but not tested on iOS (since I don't have a real device here)


jgranick:
thanks again :smile:

PR #843, opened 2016-11-17
Needed to set MACOSX_DEPLOYMENT_TARGET to 10.7 in order to fix issue 839

matthewswallace:
Not sure if there is a better way to solve this ?

### DIFF
--- a/include.xml
+++ b/include.xml
@@ -1,35 +1,35 @@
 <?xml version="1.0" encoding="utf-8"?>
 <extension>
-	
+
 	<log error="Lime ${lime} is not compatible with Haxe ${haxe} (version 3.2.0 or higher is required)" if="${haxe < 3.2.0}" />
-	
+
 	<set name="lime" />
-	
+
 	<haxedef name="native" if="cpp || neko || nodejs || cs" />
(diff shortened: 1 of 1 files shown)


jgranick:
I'm doing a bit more research, SDL 2.0.5 requires macOS 10.6 or greater (when previously HXCPP targeted 10.5 as the minimum target, or deployment target)

I believe this macro is based on the maximum supported version, that would probably be based upon the SDK used when building.

HXCPP uses this logic to discover the latest SDK available:

https://github.com/HaxeFoundation/hxcpp/blob/5d3f85072516e7a86968ae1c3b133981cf5b3811/tools/hxcpp/BuildTool.hx#L1751-L1773

I think MAC_OS_X_VERSION_MAX_ALLOWED is based on the SDK used?

jgranick:
I recently did tests with older SDKs, using a 10.6 SDK does not work, but the 10.7 SDK with deployment for 10.6 works. I cannot test 10.6 right now, though, so we can update to 10.7 if we don't have 10.6 support, otherwise I think we can leave as-is.

Thank you for your help :smile:

PR #850, opened 2016-11-24
Refactor Assets.hx - Support loading non-preloaded `library.dat`

vizanto:


jgranick:
This should be possible now :smile:

PR #851, opened 2016-12-01
Fix the IOS Accelerometer which wasn't fully linked to lime

kiroukou:
The SDL implementation of IOS Accelerometer was correct.  Wrapped around the Joystick API for commodity I guess, but working.
The problem was in fact that LIME didn't linked the joystick itself...
After few long hours, the obvious was found, the define was wrong.   IPHONE instead of IOS did fix the problem.


kiroukou:
This has been tested BTW.

I only see a difference between Android and iOs on the acceleration values.  There might be something to dig into here (but later for me).

PR #856, opened 2016-12-12
Fix Assets.loadLibrary for HTML5 target

vizanto:


jgranick:
Thank you! I got `Bytes.loadFromFile`, but `loadText` instead of `getText` for `loadLibrary` is a great idea -- just committed :smile:

PR #858, opened 2016-12-12
Fix for HTML5 Image.fromFile not applying alpha channel

vizanto:


jgranick:
I don't want to force all images loaded to be of type `CANVAS`, do you have a test case I can use that shows the problem you're having?

PR #859, opened 2016-12-12
Support Image (pre)loading progress on HTML5

vizanto:
Also passthrough loading errors to Future.
Is there a reason errors are discarded?


vizanto:
I see now I forgot to update the new `Image.loadFromFile` method to include progress updates.

jgranick:
I'm going to take a look, I want to do this a different way, but it's an important feature. Tracking here:

https://github.com/openfl/lime/issues/896

Thanks a lot 😀

PR #864, opened 2016-12-18
Add new constants to GL.hx

starburst997:
While trying to compile [Heaps](https://github.com/ncannasse/heaps), I ran into an issue where some GL constants were missing. Not sure if this is something interesting to add to Lime or if this should be added to Heaps itself.


jgranick:
Lime 4 should support these using the `gl.getExtension` API :smile:

PR #923, opened 2017-02-13
Allow setting a different Gradle PLUGIN version.

player-03:
The [Android plugin for Gradle](https://developer.android.com/studio/releases/gradle-plugin.html) uses numbers that look suspiciously similar to Gradle's numbers, but they aren't the same.

I failed to notice the extra period, so my previous pull request (#922) overwrote "`2.1.0`" with "`2.10`," thereby breaking Android builds. This request fixes it and adds an option to change the two values separately.

### DIFF
--- a/dependencies/extension-api/build.gradle
+++ b/dependencies/extension-api/build.gradle
@@ -6,7 +6,7 @@ buildscript {
 	}
 	
 	dependencies {
-		classpath 'com.android.tools.build:gradle:2.1.0'
+		classpath 'com.android.tools.build:gradle:::ANDROID_GRADLE_PLUGIN::'
 	}
 }
 
--- a/dependencies/extension-api/build.gradle
+++ b/dependencies/extension-api/build.gradle
@@ -6,7 +6,7 @@ buildscript {
 	}
 	
 	dependencies {
-		classpath 'com.android.tools.build:gradle:::ANDROID_GRADLE_VERSION::'
+		classpath 'com.android.tools.build:gradle:2.1.0'
 	}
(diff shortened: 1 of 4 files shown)


player-03:
Sorry about that!

jgranick:
No problem 😄

player-03:
Ok, so this leads to another issue. Specifically, [this file](https://github.com/openfl/lime/blob/develop/templates/extension/dependencies/android/build.gradle#L9) is run through the template processor twice.

The first time is when the user runs `lime create extension`, and the second is when they actually use the extension in a project. The first time this is run, `ANDROID_GRADLE_PLUGIN` isn't defined, so it sets a value of `null` (which is, of course, invalid).

However, it isn't safe to revert it to `2.1.0`, since the user might not even have that version installed. What we really need is for "`::ANDROID_GRADLE_PLUGIN::`" to survive the first template pass. That way, it can be replaced with the correct value during compilation.

The tricky part is that, as far as I know, Haxe's template system provides no way to escape colons. Instead, you have to define your own variable, like this:

```haxe
context.ANDROID_GRADLE_PLUGIN = "::ANDROID_GRADLE_PLUGIN::";
```

Here's a more general alternative:

```haxe
context.DOUBLE_COLON = "::";
```

On the plus side, it makes it clearer that you're escaping the colons. On the downside, the template becomes more verbose:

```text
classpath 'com.android.tools.build:gradle:::DOUBLE_COLON::ANDROID_GRADLE_PLUGIN::DOUBLE_COLON::'
```

PR #924, opened 2017-02-14
allow to pass "assetsPrefix" in flash target

kasparsj:
allow to pass "assetsPrefix" in flash target via parameters


jgranick:
https://github.com/openfl/openfl/pull/1496 :grinning:

PR #935, opened 2017-03-07
first time you build lime, binary is not built or in global path

sbimikesmullin:
actually, i'm not sure what to do to get lime in the global path on a fresh windows install.
i just always use `haxelib run lime ...`

but this was necessary to build on windows for me. hoping to save others time since its more likely to work than without.

```
C:\Users\Mike>lime
'lime' is not recognized as an internal or external command,
operable program or batch file.
```


jgranick:
I think that usually we recommend that users install OpenFL and Lime as a release version before doing a development version, because it helps install the dependencies and the `lime` and `openfl` alias shortcuts. However, I do think the use-case of cloning Lime and OpenFL from GIT on a clean install deserves a bit more polish, I'll have to look into it

PR #937, opened 2017-03-09
remove deprecated meta tag from html5 template

sbimikesmullin:
see also: https://github.com/openfl/openfl/pull/1520


jgranick:
Whoops, just did this myself manually. Thanks :grinning:

PR #944, opened 2017-03-21
Fix android apk installation for platform tools >= 23

justin-espedal:
`-d` flag doesn't exist in android sdk platform tools before version 23. Also, versions 23 and higher accept combined flags (`-rd`) instead of individually listed flags (`-r -d`).

### DIFF
--- a/lime/tools/helpers/AndroidHelper.hx
+++ b/lime/tools/helpers/AndroidHelper.hx
@@ -113,6 +113,25 @@ class AndroidHelper {
 		
 		return 0;
 	}
+
+	public static function getPlatformToolsVersion ():String {
+
+		var propertiesPath = adbPath + "source.properties";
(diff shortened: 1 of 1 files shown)


jgranick:
I really appreciate your help on this, but unfortunately, the issue is a bit more complicated.

First, the nature of auto-generated build version numbers means (especially) that allowing a downgrade is important to help us avoid confusing `lime test` scenarios where your application is built, but an old version is installed.

The platform-tools version should work if the user has a fairly recent version installed (last three years I think), but the `-d` flag requires a newer device as well. This is the trickier one to get right. I disabled the device version check because our code was bad, I think we need to write to a local file on the device, then `adb pull` to get it, rather than piping in the code we had before.

I'm closing this for now (since we recommend a new platform-tools version) but I still think we should try and sort out the install for old devices scenario :confused:

PR #949, opened 2017-04-03
Improve DCE behavior.

player-03:
As discussed [here](https://haxe.org/blog/codingtips-new/), not only is this slightly faster, it lets DCE know it's supposed to keep these constructor functions.


jgranick:
Do you now if the fields of these classes will still be removed by DCE?

Perhaps we'll need to `@:keep` each of the classes as well to prevent that?

player-03:
Based on the article, it should work without any `@:keep` annotations. The idea is that DCE is able to tell that this is a reference to the constructor, whereas previously it only saw a reference to the class. I assume what happened is it kept the class itself but removed the constructor, and that's why `Type.createInstance()` failed.

PR #956, opened 2017-04-14
HTML5GLRenderContext compressedTexImage fixes

hklindworth:
To avoid a runtime error, I had to create an UInt8Array after the call to __prepareData

This is the same approach that is in use for the regular texImage functions (without compression)


jgranick:
Would you mind trying [this branch](https://github.com/openfl/lime/tree/feature/fast-webgl-bindings) to see if it works for you?

hklindworth:
Thanks, the branch fast-webl-bindings works fine!

PR #974, opened 2017-05-07
Fix DCE breaking GL extensions

MrCoft:
With dce=full, GL returns extension objects with null attributes. The new openfl Context3D sets some constants from them and so even basic GPU functionality breaks with DCE. I added "@:keep" before every class with 1+ fields in lime.graphics.opengl.ext and it fixed my problems.
```
var bgraExtension = lime.graphics.opengl.GL.getExtension("EXT_bgra");  
trace(bgraExtension.BGRA_EXT);
```
This would trace null now.


jgranick:
Thank you :smile:

PR #980, opened 2017-05-14
UTF Strings support

RblSb:
Dependency Files


lime/text/UTFString.hx
```diff
@@ -0,0 +1,354 @@
...
+		Utf8ExtInternal.fillUpperToLowerMap(lcaseMap);
+		Utf8ExtInternal.fillLowerToUpperMap(ucaseMap);
+		inited = true;
+		#end
+	}
+	
+	/**
+		Returns a String where all characters of `this` String are upper case.
+		
+		Affects the characters `a-z`. Other characters remain unchanged.
```
larsiusprime: Technically, this function is unicode aware and would affect more than just a-z ;)

lime/text/UTFString.hx
```diff
@@ -0,0 +1,354 @@
...
+		return r.toString();
+		#else
+		return this.toUpperCase();
+		#end
+	}
+
+	/**
+		Returns a String where all characters of `this` String are lower case.
+		
+		Affects the characters `A-Z`. Other characters remain unchanged.
```
larsiusprime: Same here. This second line could be removed in both functions to make the comment more accurate.

Issue #903, opened 2017-01-24
HXP support broken (command line too long)


Gama11:
I can't get projects with HXP project files to work at all anymore (using Lime 3.6.0, Windows 10). This means that even with a completely empty `Project.hxp` file:

```haxe
class Project extends lime.project.HXProject {}
```

...running `lime build <target>` ends with "Command line is too long".

Running a `-verbose` build reveals that the command line seems to contain my entire `PATH`, which is indeed quite long on this system.

Yanrishatum:
I've managed to make a temporary fix, but it does not compatible with current helper of Lime. Thing is, Lime passes all environmental variables via command line. I've changed it from passing them via command line to passing them via `stdin`.
Pros:
* Do work.
* Do not bloat command line

Cons:
* Output from HXP process do not printed to console (unless you read the `stdout`/`srderr` of process manually)

1. Instead of data, I pass the String length of that data via arguments.
2. In `main()` after argument count check: 
```haxe
    for (i in 0...args.length)
    {
      args[i] = Sys.stdin().readString(Std.parseInt(args[i]));
    }
```
2. Instead of calling `ProcessHelper.runCommand("", "neko", ...)` I used custom code in `fromFile` function:
```haxe

    var argList:Array<String> = [HXProject._command, name, 
      Std.string (HXProject._target), Std.string (HXProject._debug), Serializer.run (HXProject._targetFlags), 
      Serializer.run (HXProject._templatePaths), Serializer.run (HXProject._userDefines), Serializer.run (HXProject._environment), temporaryFile];
    var nekoArgList:Array<String> = [for (a in argList) Std.string(a.length)];
    nekoArgList.unshift(FileSystem.fullPath(nekoOutput));
    var p:Process = new Process("neko", nekoArgList);
    for (arg in argList)
    {
      p.stdin.writeString(arg);
    }
    var code:Int = p.exitCode();
    p.close();
```
Quite simple, but there is room for improvement.

PR #982, opened 2017-05-17
Don't pass env vars to neko via command line when processing HXP project files

atom-b:
neko will use the current process environment's variables when passed "null" as the environment variable parameter.

Sets haxe_ver, haxe, and haxe[major version] via Sys.putEnv so they're available to HXP project files.

resolves openfl/lime#903


jgranick:
Thanks! I think I have this working now with a temporary file, but extra help testing it would be great :smile:

PR #985, opened 2017-05-20
HTML5 setFullscreen/Icon/Title

RblSb:
Works, but I can not understand how to make onFullscreen.dispatch() for openfl event.
Waiting for review.
Test:
https://gist.github.com/RblSb/dfc314f9f107dd02bb3df2b840b0e4eb

### DIFF
--- a/lime/_backend/html5/HTML5Window.hx
+++ b/lime/_backend/html5/HTML5Window.hx
@@ -774,14 +774,112 @@ class HTML5Window {
 	
 	public function setFullscreen (value:Bool):Bool {
 		
+		trace ("setFullscreen", value);
+		if (value) {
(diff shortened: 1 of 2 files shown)


jgranick:
Thank you!

I have added an initial implementation of fullscreen, setIcon and updating title, based on your code. Title is the same, and setIcon should be simplified a bit (an icon of a greater size than 16x16 seems to work alright for Firefox and Opera, but Edge doesn't appear to be working right. Was this necessary on another browser?)

I believe I have `onFullscreen` dispatching as expected as well :smile:

Let me know if we need anything more to get it working as well your original code did in your tests :smile:

RblSb:
I am glad that I helped with these sketches of the code =)

PR #1000, opened 2017-06-21
sounds are loaded by howler.js automatically

kasparsj:
no need to call `load` since sounds are loaded by howler.js automatically


jgranick:
Please try out this change, I think we want `preload: false` in order to avoid dispatching events before we've added the listeners :smile:

kasparsj:
tried - works, thanks!

PR #1069, opened 2017-09-07
Add possibility to build source map file for minified JS file

brainexe:
As lime supports the google closure compiler to minify assets it would be great to build source map file for this minified JS version as well to have improved debugging experiences in the production version as well. (only if -source-map flag was passed).

### DIFF
--- a/lime/tools/helpers/HTML5Helper.hx
+++ b/lime/tools/helpers/HTML5Helper.hx
@@ -191,6 +191,20 @@ class HTML5Helper {
 					
 				}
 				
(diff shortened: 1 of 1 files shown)


jgranick:
Nice! Does this update the Google Closure compiler, if so, what version?

brainexe:
Yep, I chose the most recent one: v20170806 which includes a lot of fixed/improvements since 2015

Gama11:
Is there a reason to hide this behind an additional flag (rather than doing what Haxe does, which is to generate a source map when `-debug` is enabled)?

PR #1074, opened 2017-09-15
UWP html5

jarbot:
Here are the templates necessary for a Universal Windows Platform js project.

What works:
- running ```openfl build windows -uwp``` creates a Visual Studio UWP js project and builds it with msbuild

Requirements
- VS 2017
- build must be ran from the VS Developer command prompt

TODO:
- the 'test' target does not work correctly but I've done some investigation, notes are in WindowsPlatform.hx in the run function.
- I'm assuming this should work without the VS Developer command prompt but I have no idea how to set this up
- Build and interface for native WinJS commands for IAP etc


jgranick:
Manually merged :shipit:

PR #1077, opened 2017-09-27
Start HTTPRequest timeout when doWork is called

starburst997:
Otherwise we get some false-positive timeout since the HTTPRequest was never actually "called" but was waiting for other request to finish in the ThreadPool.


jgranick:
Thanks! This inspired the following fix

https://github.com/openfl/lime/commit/dc0e59f13ef92f0170f4d8e6f2f68873f1365882

starburst997:
Yay!

PR #1092, opened 2017-10-18
 Add http headers in native + fixes

aW4KeNiNG:


jgranick:
Thanks :smile:

I've made minor changes to this, if you don't mind trying the latest development version. I made `enableResponseHeaders` false again by default (it became try in the "Content-Length" optimization), using your logic to fill the array if they are enabled, otherwise I think it should work the same :smile:

aW4KeNiNG:
Perfect 👍

PR #1094, opened 2017-10-20
Added better progress event logic:

TomByrne:
- Works for 'data:' URLs
- Works in IE11

### DIFF
--- a/lime/_backend/html5/HTML5HTTPRequest.hx
+++ b/lime/_backend/html5/HTML5HTTPRequest.hx
@@ -2,12 +2,14 @@ package lime._backend.html5;
 
 
 import js.html.AnchorElement;
+import js.html.ErrorEvent;
 import js.html.Event;
 import js.html.Image in JSImage;
 import js.html.ProgressEvent;
 import js.html.XMLHttpRequest;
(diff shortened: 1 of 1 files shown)


jgranick:
Our current behavior is that we load images as an HTML5 `ImageElement`, first, then convert to a `CanvasElement` and/or `ImageData` later, depending on how you access the pixels. I believe there are performance benefits to having an `ImageElement` over a `CanvasElement` (when possible), as well as better integration with the DOM target. Does an `ImageElement` have problems with its progress events? Thanks :smile:

jgranick:
Perhaps we can use the XHR behavior if `onprogress` does not exist on the `Image` element. I think that might support newer browsers that do have image progress events, while also providing progress to those that don't. I'll try and take a look at this soon

jgranick:
Thanks again for your help!

Here's how I changed the behavior:

https://github.com/openfl/lime/commit/70e1bed0ae0b7adbeff24d1ea07c0188012ac57e

Would you mind testing it?

This should (theoretically) use `HTMLImageElement` for browsers that have `onprogress` support, but fallback to XHR on browsers that do not

PR #1111, opened 2017-11-15
Added <window httpReqTimeout= > attribute to project.xml

Greg209:
Allows the 'httpReqTimeout=' to be added as an attribute to the <window> element overriding the default 30000 ms for the HttpRequest timeout. 

e.g. <window httpReqTimeout="180000" if="html5"/>


jgranick:
Thank you!

I've added `openfl.net.URLRequestDefaults` to help implement this on the OpenFL side, let me know if you think we need any more :smile:

https://github.com/openfl/openfl/commit/27c227da2177976178ed81b42e210b8b9a134065

PR #1112, opened 2017-11-15
Lib folder seperation for targets hosted in the same location

Greg209:
When SWF assets are to be shared across targets (e.g. HTML & Flash) due to differences in the library.json file separate, target specific versions are needed. This PR along with an equivalent PR for OpenFL (https://github.com/openfl/openfl/pull/1757) changes the generated 'lib' folder to be target specific e.g. 'lib-html5' or 'lib-flash'. 

This allows the two targets to be copied to the same location without conflict and avoiding duplication of assets.

NOTE: This is to be merged along with the equivalent OpenFL PR (https://github.com/openfl/openfl/pull/1757)


jgranick:
Resolved another way, but we can come back and revisit this if we think there's a real need :smile:

PR #1131, opened 2017-12-17
Use canvas for copyPixel operations

hklindworth:
I did some tests regarding copyPixels and except for very small images (around 10x10 pixels) it is significantly faster to use the canvas implementation instead of manipulating individual pixels via javascript.

E.g. in this benchmark BitmapDataScene it his around 20 times faster to use canvas than data:
https://github.com/innogames/openfl-mark

### DIFF
--- a/lime/graphics/Image.hx
+++ b/lime/graphics/Image.hx
@@ -365,9 +365,7 @@ class Image {
 			
 			case CANVAS:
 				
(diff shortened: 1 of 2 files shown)


jgranick:
Thanks! I presumed that canvas was faster, but didn't expect the cut-off to be about 10x10 pixels. Thanks for testing this :smile:

PR #1136, opened 2018-01-03
OpenAL Effects Extension 

CawawaC:
It was already in there, I just exposed the API. It works, very nicely I might add :)

Also, this needs a documentation and some functions for ease of use. I mean not everyone knows what's an auxiliary send and how to use it :/ and OpenAL EFX loves those. Any suggestion on how to go about this?


jgranick:
This is awesome, thank you!

PR #1140, opened 2018-01-19
Sound improvements

chatziko:
2 sound-related improvements:

1. mp3 detection failed for some of my mp3s. I used the test from [is-mp3](https://github.com/hemanth/is-mp3/blob/master/index.js) which allows one more byte combination than `AudioBuffer._getCodec` did.

2. Playback of dynamically created Sounds (via `loadCompressedDataFromByteArray`) was not working for me in html5. After some debugging I found that Howler did not like calling `volume (value, id)` with `id == -1`. This seems useless anyway, so I disabled it and now playback works fine.


jgranick:
Awesome, thank you :smile:

PR #1158, opened 2018-02-22
Minification fixes/improvements

chatziko:
Some closure-related fixes/improvements:

1. pass `--strict_mode_input false`, closes openfl/openfl#1827

2. pass `--source_map_location_mapping "/|file:///"`, which causes paths to be prefixed by `file://` (similarly to the haxe-generated .js.map) so that the debugger can find them.

3. Even when closure creates a .js.map, it does _not_ add a `//# sourceMappingURL` to the .js file! We do it ourselves so that the debugger can find it.

4. Haxe can generate source maps for `-final` builds using `-Djs-souce.map`, but minification discards them cause it is triggered by `debug || source-map`. I changed the condition to `<souce.js.map exists> || source-map` which is simpler.

With the above changes debugging a minified app in chrome dev tools works pretty well.


jgranick:
Perfect 👍

PR #1168, opened 2018-03-20
UTF8String: fix substr() calls with len == null, closes #1167

### LINKED ISSUE #1167: UTF8String.substr() call without len doesn't work properly
```haxe
var s:String = "foo";
trace(s.substr(2)); // o

var s:lime.text.UTF8String = "foo";
trace(s.substr(2));
```

The second example crashes on Neko, while on CPP the result is incorrect (the result is an empty string).

```
Called from lime.text._UTF8String.UTF8String_Impl_::substr line 177
Called from haxe.Utf8::sub line 92
```

This was causing some issues in Flixel's debugger overlay, which calls `substr()` on the `text` of an OpenFL `TextField`, which is a `UTF8String`.

Gama11:


jgranick:
Awesome, thanks

PR #1171, opened 2018-03-26
Export multiple IPA types with one deploy command

atom-b:
Avoids re-generating an xcarchive for each IPA type you wish to create so all IPAs have the same UUID and can be symbolicated with the same symbols.

Usage: `openfl deploy ios [-adhoc] [-appstore] [-enterprise] [-development]`


jgranick:
:shipit:

PR #1187, opened 2018-05-27
WIP: Add language support for air, ios, tvios, mac, android

Hersir88:
Added supported language tag for air, mac, tvos, ios, android. This prop lets to control what languages will be bundled in apps. 
Did testing on iOS, air, had issues to set up android, will try to test later.

Tried to align as much as possible with code style, but as it's my first PR here, please comment if something is wrong.

To use this feature language tag is used inside the `project.xml` for example like this `<language supported="en de" />`

### DIFF
--- a/lime/project/HXProject.hx
+++ b/lime/project/HXProject.hx
@@ -64,6 +64,7 @@ class HXProject {
 	public var targetHandlers:Map<String, String>;
 	public var templateContext (get_templateContext, null):Dynamic;
 	public var templatePaths:Array<String>;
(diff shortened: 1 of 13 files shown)


jgranick:
Thank you!

I'm sorry that things are a little bit turned around, since there's some semi-major refactoring going on.

I've merged `languages` support in the tools, using `<language name="en" />`. I also updated the templates in order to use this value, when present:

https://github.com/openfl/lime/commit/673d1cf0145139f02429c59208d53312207952c9

PR #1194, opened 2018-06-01
add RASPBERRYPI conditional in openglbindings

gepatto:
RaspberryPi is Linux but gl_debug  is not avaible. 
Adding a conditional (3x) for RASPBERRYPI in openglbindings.cpp


Gama11:
Oh, sorry about that. I guess inverting the conditionals didn't make it _that_ much simpler after all.

PR #1198, opened 2018-06-05
fix electron hxml template files

peteshand:


Gama11:
Why `dce no`?

peteshand:
yes, no dce for Electron.hx

peteshand:
can someone merge this

jgranick:
@peteshand Should we do a `@:keep` on the Electron extern file? Is that the issue? It would be nice to know what underlying issue this fixes. Thanks :smile:

peteshand:
Hey, sorry about the delayed reply. yeah maybe `-dce no` is a bit overkill... the more important change is removing `::HAXE_FLAGS::` from the ElectronSetup compile as keeping this can cause runtime error, I'm mainly seem this when I've including swf assets

PR #1199, opened 2018-06-05
Don't do any magical implicit to-canvas conversion in the Image.src getter.

nadako:
We want to access normal source value which can be HTML Image element, also suitable for texture uploading.

If anyone needs a canvas, they can convert explicitly (as it's actually done already).

### DIFF
--- a/lime/graphics/Image.hx
+++ b/lime/graphics/Image.hx
@@ -1828,14 +1828,6 @@ class Image {
 	
 	private function get_src ():Dynamic {
 		
-		#if (js && html5)
-		if (buffer.__srcCanvas == null) {
-			
-			ImageCanvasUtil.convertToCanvas (this);
-			
-		}
-		#end
-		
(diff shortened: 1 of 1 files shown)


nadako:
This is one part we had to change to avoid to-canvas conversion of all the `BitmapData`s loaded over the network. The texture can be uploaded directly from the HTML Image element, so conversion is redundant for that purpose and is very slow on MS Edge.

nadako:
The second part is here: https://github.com/openfl/openfl/pull/1913

jgranick:
I'm trying this:

https://github.com/openfl/lime/commit/76854e801a88574a9b2b77babfb88332dc79f085

I guess the issue is we have no `IMAGE` type for Lime images, or similar. Perhaps we don't need one, but the intention here was supposed to be that `image.src` would return an HTML image if no canvas is necessary. Thank you for your help!

PR #1200, opened 2018-06-05
Update ImageDataUtil's resizeBuffer method so it clears source data

Beeblerox:
This modification helped me to reduce memory consumption on html5 target (especially with activated `openfl_power_of_two` haxedef).
I'm not sure if it's right to do so, and ask you to consider this change


jgranick:
Yeah, I think this makes sense, since the source content no longer matches the size. It would be possible to reuse the canvas element later (if you were performing `copyPixels` and other operations that relies on having a canvas element) but that would be at the risk of extra memory. This seems safe :smile:

PR #1203, opened 2018-07-01
Add adobe air custom simulator support

Hersir88:
Improved Adobe air target:
- auto launch adobe air app on iOS simulator
- add universal device support for adobe air
- add adobe air custom simulator support

Same logic as for native iOS builds is reused (ios.device for device support and -simulator-name as argument).
Tested on adobe air target, will check if native works as before as soon as will fix [errors](http://community.openfl.org/t/mobile-build-fails-with-dev-version-of-lime/10687) on the native build.

Right now moved sim check code as-is to separate file so it can be reused, but in future maybe could be refactored a bit as well and covered with unit tests.

More tests and feedbacks are welcome :).

### DIFF
--- a/lime/tools/helpers/IOSHelper.hx
+++ b/lime/tools/helpers/IOSHelper.hx
@@ -5,6 +5,7 @@ import haxe.io.Path;
 import lime.project.Platform;
 import lime.tools.helpers.PathHelper;
 import lime.tools.helpers.ProcessHelper;
+import lime.tools.helpers.XCodeHelper;
 import lime.project.Haxelib;
 import lime.project.HXProject;
 import sys.io.Process;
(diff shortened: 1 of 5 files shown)


lime/tools/helpers/IOSHelper.hx
```diff
@@ -319,144 +320,52 @@ class IOSHelper {
...
-					
-				} else {
-					
-					currentDeviceID = devices.get ("iphone-6");
-					
-				}
-				
 			}
+			// TODO: check if needed as not used
+			var templatePaths = [PathHelper.combine(PathHelper.getHaxelib(new Haxelib("lime")), "templates")].concat(project.templatePaths);
```
Hersir88: this prop is not used in the code in this branch safe to remove or it should be used?

lime/tools/helpers/IOSHelper.hx
```diff
@@ -319,144 +320,52 @@ class IOSHelper {
...
-			
-			var xcodeVersion = getXcodeVersion ();
-			
-			ProcessHelper.runCommand ("", launcher, [ "install", "--noninteractive", "--debug", "--bundle", FileSystem.fullPath (applicationPath) ]);
-			
+			var templatePaths = [PathHelper.combine(PathHelper.getHaxelib(new Haxelib("lime")), "templates")].concat(project.templatePaths);
+			var launcher = PathHelper.findTemplate(templatePaths, "bin/ios-deploy");
+			Sys.command("chmod", ["+x", launcher]);
+			// TODO: check if needed as not used
+			var xcodeVersion = getXcodeVersion();
```
Hersir88: this prop is not used in the code in this branch safe to remove or it should be used?

jgranick:
Thanks for this! I'll try and look soon

jgranick:
Okay, should have this merged now. Thank you 😀

PR #1222, opened 2018-08-12
Html5 mp3 detection update

andrew-git:
Looks like MP3 could have FF FX bits (https://www.garykessler.net/library/file_sigs.html),
but in my sample for mp3 22k second bit was 243 and after change normally plays in Firefox and Chrome.
Need test and maybe update algo to take into account all second bits started with F


jgranick:
Thanks!

PR #1231, opened 2018-09-05
HTML5 externs fix: void functions may return non-void

haxiomic:
A number of functions are annotated as `: Void` but return the result of call. In recent versions of the HTMLMediaElement specification `play()` [returns a promise](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play#Return_value). To avoid errors when the externs are updated I've removed the `return`.

(See https://github.com/HaxeFoundation/haxe/pull/7354)


jgranick:
Thanks!

PR #1239, opened 2018-09-16
calculate AssetsMacro.cacheVersion at compile time

shohei909:
AssetsMacro.cacheVersion was changed in this commit.
https://github.com/openfl/lime/commit/aa4c4b6a05a31d0b70c409682e889b59c35cadc3

And now, it is calculated at run time.

However, I think compile time calculation is better than run time one, because of browser cache in html5.


jgranick:
This was unintentional. Thank you so much for your help catching and fixing this :smile:

(What's the use of a cache break string if it's always breaking? :frowning: )

PR #1243, opened 2018-09-24
Updated Howler to enable sound position

miltoncandelero:
- Updated Howler to version 2.0.15
- Implemented the 3D Positional sound*

---

\* My knowledge of sound isn't enough to understand where the "w" variable of the position goes (as "pos()" uses only x, y and z). My guess is that "pannerAttr()" might have the access to what we are looking for, but again I don't fully understand sound.


jgranick:
This is a great start! We could add externs for the rest of the positional sound API, but for now, this seems like a good step forward

PR #1246, opened 2018-10-02
add language-specific rules (for e.g. Turkish) to Utf8String casing

larsiusprime:
Turns out that having a map of upper to lower case unicode glyphs, even taking into account asymmetrical operations, isn't enough. A good example of this is Turkish.

Turkish uses what looks pretty much just like some letters from the Latin alphabet with a few variants with diacritics. 

https://unicode-table.com/en/alphabets/turkish/

And indeed, these are all just using the same glyphs as the various european alphabets do. Same exact unicode points. BUT, the rules for Turkish are subtly different.

Notably, Turkish has dotted and undotted i's, in both lower and upper case.

Capital I: https://unicode-table.com/en/0049/
Small dotless I: https://unicode-table.com/en/0131/

Capital dotted I: https://unicode-table.com/en/0130/
Small dotless I: https://unicode-table.com/en/0069/

The tricky bit is, that the "i" and "I" characters are shared with standard latin alphabets and it's impossible to tell from text alone whether the user wants the Turkish capitalization rules or not.

The only solution for this, is for the user to be able to tell the system which language rules they intend to use.

So this proposed change adds an optional parameter to the Utf8String toUpperCase() and toLowerCase() functions, which allows the user to specify the language. For now I'm just adding "TURKISH" and "STANDARD" (everything else), and Turkish technically only has the one case to take care of properly.

Technically you only need it going from lower case to upper case to make sure that a dotted lowercase i becomes a dotted capital I, because a dotless lowercase I is already mapped to a dotless capital I, and a dotted capital I is already mapped to a dotted lowercase I, but I put both cases in explicitly for Turkish since I figure there will be other language-specific gotchas in the future, and it helps to have the symmetrical structure already in place.

### DIFF
--- a/src/lime/text/UTF8String.hx
+++ b/src/lime/text/UTF8String.hx
@@ -209,22 +209,35 @@ abstract UTF8String(String) from String to String {
 		Returns a String where all characters of `this` String are lower case.
 
 		Affects the characters `A-Z`. Other characters remain unchanged.
+		
+		If `language` is specified, language-specific casing rules will be followed.
 	**/
-	public function toLowerCase ():String {
+	public function toLowerCase (language:Language=null):String {
 
+		if(language == null) language = STANDARD;
+		
 		#if sys
 
 		if (lowercaseMap == null) {
 
 			lowercaseMap = new Map<Int, Int> ();
-			Utf8Ext.fillUpperToLowerMap (uppercaseMap);
-
+			Utf8Ext.fillUpperToLowerMap (lowercaseMap);
+			
 		}
 
 		var r = new Utf8 ();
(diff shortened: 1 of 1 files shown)


jgranick:
Do you use `UTF8String` directly, or would this need to be implemented behind the OpenFL `TextField` somehow?

larsiusprime:
I use UTF8String directly. It's not just about displaying the text correctly on screen, it's about handling the data value itself (such as e.g. writing to a file, passing on to other functions/libraries/etc). I just need a proper locale-aware API that can properly do the transformation at the data level.

PR #1254, opened 2018-10-29
Add forceLocal define to allow loading locally

loudoweb:
Compiling with -DforceLocal allow files to be served locally when browser supports this.
Use case : Cordova use a webview and many files can't be loaded without allowing http status 0 as successful. So this PR allows apps made with openfl to run on Cordova.


loudoweb:
I don't know if there is a way to know at runtime if the app runs on Cordova. So this is why I propose to introduce this tag.

jgranick:
I renamed the flag to `-Dallow-status-0`, let me know if you think another name would be even better, but I agree, it isn't always possible to know at runtime whether the page will be accessed from a local environment or not.

PR #1268, opened 2018-11-24
Allow use of launch storyboard instead of launch images on iOS

justin-espedal:
Apple seems to be pushing more and more to stop the use of static launch images. As far as I can tell, it's not even possible to target the 11-inch iPad with a static launch image, like it has been for previous iDevices.

If a launch storyboard is used, static launch images will be ignored. The storyboard has to be set up in xcode, but perhaps some simple templates can be included in lime.

### DIFF
--- a/src/lime/tools/HXProject.hx
+++ b/src/lime/tools/HXProject.hx
@@ -40,6 +40,7 @@ class HXProject extends Script
 	public var javaPaths:Array<String>;
 	public var keystore:Keystore;
 	public var languages:Array<String>;
+	public var launchStoryboard:LaunchStoryboard;
 	public var libraries:Array<Library>;
 	public var libraryHandlers:Map<String, String>;
 	public var meta:MetaData;
@@ -412,6 +413,11 @@ class HXProject extends Script
 		}
 
 		project.languages = languages.copy();
+		
+		if (launchStoryboard != null)
+		{
+			project.launchStoryboard = launchStoryboard.clone();
+		}
 
 		for (library in libraries)
 		{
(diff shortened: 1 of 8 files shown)


justin-espedal:
A simple template is included now. Here's an example of its usage.

```
<launchStoryboard template="splash.storyboard" assetsPath="Images/splash"
    bg="0xFFFFFF"
    contentMode="center"
    portrait="iphone-splash.png"
    ipad="ipad-splash.png">
            
    <imageset name="ipad-splash.png" />
    <imageset name="iphone-splash.png" />
        
</launchStoryboard>
```

For each `imageset`, lime looks for `{name}.png`, `{name}@2x.png`, and `{name}@3x.png` in the specified `assetsPath`, copying any of those that exist into an Xcode imageset. Most iPhones these days require the @2x or @3x image, while iPads require @2x.

The included template, `splash.storyboard` can be customized in four ways:

`bg` is the background color, in case you're not using a full-screen content mode.

`contentMode` is how the images are scaled to fit the screen.

- `center` is unscaled
- `scaleAspectFill` scales to fill the screen, with original aspect ratio
- `scaleAspectFit`: scales to fit inside the screen (letterbox), with original aspect ratio
- `scaleToFill`: stretches the image to fill the screen

`portrait`, `landscape`, and `ipad` are used to specify up to three different imagesets to use for different devices/orientations.

jgranick:
I'm sorry this has sat this long

I am not able to test this very well right now. If this has been working stable for you guys then I will merge it as-is

Does the older launch image as a flat image still work with this code?

Thank you very much

justin-espedal:
Yeah, if a launch storyboard is not specified, it'll still try to use the flat launch images.

I've updated the PR to be a single commit that can be merged as-is. The formatting has mostly been updated to match lime's current standards as well, but I did that manually, so it could be off in places.

PR #1282, opened 2019-01-25
Add Method to override Electron Main Process entry class

peteshand:
When targeting electron the Main Process entry class is automatically generated based off
~/lime/templates/electron/haxe/ElectronSetup.hx

In some cases it may be desirable to override this class with something custom.

This pull request addresses this by allowing the user to define an override class in the lime project.xml file as follows:

```
// add a new template dir reference
<templatePath name="./template" />

// add a reference to the Main Process entry haxe class
<config:electron electron-main="ElectronMainOverride" />
```


peteshand:
Does anyone have any issues to this PR?
If not can you please merge.

jgranick:
Can the user override with the template system already?

1. Create "templates/electron/haxe/ElectronSetup.hx"
2. Add `<templates path="templates" />` to project

This is how you override `ApplicationMain` or other files, currently

peteshand:
Hmmm nope, doesn't appear to work.

jgranick:
I see the following code:

```haxe
			ProjectHelper.recursiveSmartCopyTemplate (project, "electron/template", destination, context);

			if (project.app.main != null) {

				ProjectHelper.recursiveSmartCopyTemplate (project, "electron/haxe", targetDirectory + "/haxe", context, true, false);
				ProjectHelper.recursiveSmartCopyTemplate (project, "electron/hxml", targetDirectory + "/haxe", context);

			}
```

So this should resolve `<template />` paths in order, like class paths. That means you will need to add your `<template />` tag after including OpenFL/Lime, not before, in order to resolve after.

The file will also have to be a duplicate of the Lime version, plus tweaks, in order for it to work properly

peteshand:
Thanks for the pointers, The issue turned out to be due to the fact that I copy and pasted from above
```
Add <templates path="templates" /> to project
```

should be
```
Add <template path="templates" /> to project
```

PR #1286, opened 2019-02-04
Hide __ASSET__ classes with @:noCompletion

Gama11:
The `#if display private #end` trick doesn't work with Haxe 4 because during "indexing" / creation of the parser cache, `display` is not defined (because the results have to be usable for compilation too). So currently the `__ASSET__` classes show up in completion:

`@:noCompletion` fixes that.

I also found some types in `HTTPRequest` prefixed with underscores that don't look like they should show up?


jgranick:
Thanks!

PR #1287, opened 2019-02-07
fix support of android.gradle-plugin config value

Mataemon:
The ANDROID_GRADLE_PLUGIN config value is used in AndroidPlatform.hx, but was not read in ProjectXMLParser


jgranick:
The `<android />` tag is deprecated.

Can you set this using `<config:android gradle-plugin="" />` instead?

Mataemon:
It's working.
My bad, I close the request.
Thank you!

PR #1292, opened 2019-03-03
Lime changes to ios keyboard and android status bar for non fullscreen.

PXshadow:


player-03:
Do you remember what this was trying to accomplish? The title doesn't provide much to go on.

PXshadow:
> 

I think so. It was something to do with the status bar at the top changing by the keyboard being pulled up.

player-03:
A lot of these changes look project-specific, better suited to an [extension](https://player03.com/openfl/openfl-extensions/). For instance, if you want a transparent status bar, you can set that yourself.

As for the Gradle and SDL updates, that's way too much stuff for a single pull request. (Oh, and 8.2.0-Dev already has an updated version of SDL.) It's much easier to verify a small pull request.

Hope you don't mind if I close this. If you figure out what you were going for, feel free to make a new PR with only those changes.

PR #1297, opened 2019-03-15
Gamepad mappings for the Android device is empty

0velychk0:
Gamepad mappings for the Android device is empty, So what if we use the same as for Linux?

### DIFF
--- a/src/lime/ui/Window.hx
+++ b/src/lime/ui/Window.hx
@@ -243,7 +243,7 @@ class Window
 		];
 
 		Gamepad.addMappings(mappings);
-		#elseif linux
+		#elseif (linux || android)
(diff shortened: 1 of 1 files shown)


jgranick:
I think the mappings may need to be a little different (different IDs?) for different platforms. Would you mind adding any missing items from this list:

https://github.com/gabomdq/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt

It has a number of gamepads for Android, and ones we may be missing for iOS, Windows, macOS or Linux

0velychk0:
Thanks for the Link. Looks like it is what I need.

jgranick:
Would love a pull request 😉

PR #1319, opened 2019-05-15
Add batch event processing in lime/sdl

aprothman:
Provide the BatchUpdate(n) method in the SDLApplication class
that supports pulling multiple events at a time from the sdl
event queue. BatchUpdate does similar work to Update, but it
will never wait indefinitely for an event to become available,
and it accomplishes its work more efficiently. Events are
fetched to a buffer at the class level, and this buffer needs
to be freed in ~SDLApplication().

Because BatchUpdate will never wait indefinitely, it can be
called in Init to process all pending events at startup (they
can accumulate) and to initialize the framerate timer for targets
that make use of it. This allows us to clean up Update by
removing the firstTime boolean in its event loop.

Going forward, BatchUpdate is a tool that can be used to
refactor the event loop logic, improving it for all targets. It
can also be used to implement the event loop inversion of control
demonstrated at Haxe Summit 2019 for embedding OpenFL apps into
the windows of host applications.


project/src/backend/sdl/SDLApplication.h
```diff
@@ -51,13 +52,18 @@ namespace lime {
 			void ProcessTouchEvent (SDL_Event* event);
 			void ProcessWindowEvent (SDL_Event* event);
 			int WaitEvent (SDL_Event* event);
+			int GetPendingEvents (SDL_Event* event, int NumEvents);
```
player-03, line 55: Shouldn't the argument be `int maxEvents` to match the implementation?

project/src/backend/sdl/SDLApplication.cpp
```diff
@@ -923,6 +932,83 @@ namespace lime {
...
+			if (NULL != eventQueue) {
+				SDL_free (eventQueue);
+			}
+
+			eventQueue = reinterpret_cast<SDL_Event*> (SDL_malloc (queueLength * sizeof (SDL_Event)));
+			if (NULL == eventQueue) {
+				queueLength = 0;
+				queueMaxLength = 0;
+			} else {
+				queueMaxLength = queueLength;
```
player-03, line 945: I feel like `queueMaxLength` is the wrong name if you can just increase it. From the name, I'd expect excess events to be discarded or something.

Actually, what is the `queueLength` variable even for? It stores the last value of `numEvents`, but it's never needed outside this function. We could have `queueLength` do what `queueMaxLength` is currently doing, and use `numEvents` as the number of events. (That is, find and replace "queueLength" → "numEvents", then replace "queueMaxLength" → "queueLength".)

project/src/backend/sdl/SDLApplication.cpp
```diff
@@ -923,6 +932,83 @@ namespace lime {
...
+				eventQueue[nextEvent].type = -1;
+				if (!active) {
+					return -1;
+				}
+			}
+		}
+
+		// handle the mouse motion event if there is one this batch
+		if (NULL != mouseMoved) {
+			HandleEvent (mouseMoved);
```
player-03, line 972: As I said elsewhere, I'm not a fan of doing this out of order. It could get skipped if the above loop returns. Another approach would be to add a second loop above the first:

```cpp
for (nextEvent = 0; nextEvent < numPending; ++nextEvent) {
    if (eventQueue[nextEvent].type == SDL_MOUSEMOTION) {
        if (mouseMoved != NULL) mouseMoved->type = -1;
        mouseMoved = &eventQueue[nextEvent];
    }
}
```

This will set all but the final mouse moved event's type to -1. Then the main loop would skip over those events, and execute all >= 0 events in the right order.

project/src/backend/sdl/SDLApplication.cpp
```diff
@@ -923,6 +932,83 @@ namespace lime {
 	}
 
 
+	int SDLApplication::BatchUpdate (int numEvents) {
+
+		SDL_Event* mouseMoved = NULL;
```
player-03, line 926: Might as well move this down to where it's used.

project/src/backend/sdl/SDLApplication.cpp
```diff
@@ -923,6 +932,83 @@ namespace lime {
...
+				queueMaxLength = 0;
+			} else {
+				queueMaxLength = queueLength;
+			}
+		}
+		
+		int numPending = GetPendingEvents (eventQueue, queueLength);
+
+		// only allow GC free objects while we're handling events
+		if (isGCBlocking) System::GCExitBlocking ();
```
player-03, line 952: As I said elsewhere, this seems disruptive. Maybe it's appropriate, but the comment doesn't do a good job of explaining why.

PR #1325, opened 2019-05-24
Updated the documentation for createWindow

aprothman:
In the block documentation for Application.createWindow, changed
the legacy window parameter to the current attributes parameter


jgranick:
Thank you

PR #1338, opened 2019-07-06
Fix HTML5Window antialiasing check

haxecoder:
ApplicationMain template always has antialiasing parameter, so it is needed to be ckecked for value


jgranick:
Awesome thank you

Issue #1371, opened 2020-01-20
Needed update for gradle files


PlutoniumBytes:
Hi, 

All urls like "http://jcenter.bintray.com/" needs to be https, http is not working anymore

jgranick:
Resolved by #1372

PR #1372, opened 2020-01-21
Android gradle-related files update (http->https)

mc0239:
- Use [gradle's built-in repository notations](https://docs.gradle.org/current/userguide/declaring_repositories.html) for jcenter & google.

- Use https for `distributonUrl` in gradle-wrapper.properties as `services.gradle.org/distributions/` is only accessible over https.


mc0239:
Fixes issue #1371

PR #1379, opened 2020-02-01
Air deployment

Beckwahn:
Replaced file compression by usage of a target flag for deploying with captive or shared runtime option.
- - - 
openfl deploy air
openfl deploy air -captive
- - - 
Also changed default AIR SDK version to 32.0


player-03:
We might want to default to the bundle for backwards compatibility. Also maybe this should be something you set via `<config:air />`. So you'd set either `<config:air runtime="captive" />` or `<config:air runtime="shared" />`.

joshtynjala:
> We might want to default to the bundle for backwards compatibility

Yes, it should default to bundle. In that case, if there's still a command line flag, it should be `-shared` instead of `-captive`.

> Also maybe this should be something you set via <config:air />. So you'd set either <config:air runtime="captive" /> or <config:air runtime="shared" />.

It's worth mentioning that the choice between shared and captive runtime affects desktop only, at least in modern AIR SDKs. However, in the past, you could choose between shared and captive runtime on Android too. In modern AIR SDKs, captive runtime is always forced on Android, regardless of choice between the "apk" and "apk-captive-runtime" targets because Adobe/HARMAN stopped distributing the AIR shared runtime for Android. On iOS, there has never been a shared runtime, and it's effectively always been virtually similar to captive (technically AOT compiled as native code instead of the bytecode being run with JIT).

Calling it desktop-runtime may be better, just to make it clear which platforms are affected.

It's also worth mentioning that we already have command line flags that affect `lime build air -ios`. We use `-appstore` and `-adhoc` to select the `ipa-app-store` and `ipa-ad-hoc` targets. With that in mind, it would make sense to have a similar option to affect the desktop target.

PR #1390, opened 2020-04-04
Update Google Closure to v20200315

Gama11:
We have a Flixel demo (https://haxeflixel.com/demos/FlxNapeTilemap/) that breaks when compiled with `-final`, but works fine with `-dce full`, meaning that minification is the culprit. According to the commit history, the closure binary bundled with Lime was last updated in 2017, and updating it indeed solves the problem.


Gama11:
@jgranick ping.

PR #1391, opened 2020-04-06
WIP: Trying to fix support for unicode text for native targets

restorer:
## WORK IN PROGRESS, DO NOT MERGE

Currently, it seems that support for non-ASCII text is broken when using the latest versions of Haxe 4, Lime, and OpenFl.

- Haxe - 4.0.5
- Lime - latest from `develop` branch
- OpenFl - 8.9.6
- hxcpp - 4.0.64
- Target - mac or android, probably windows, linux and ios are affected too

### Before fix

**[just lime] TextRendering sample from lime-samples (mac):**
Screen Shot 2020-04-06 at 10 04 02

**[openfl + lime] Some random game (android):**
photo_2020-04-06_13-18-53

I tried to figure this out and found out that `HxString` may contain data with different encodings. I tried to change `c_str()` to `hxs_utf8()` in the `HarfbuzzBindings.cpp`, and I thought it might help (at least if I print them using `printf()`, the console will show the correct lines). However, there's something else I haven't found yet.

### After fix

**[just lime] TextRendering sample from lime-samples (mac):**
Screen Shot 2020-04-06 at 13 10 50

I don't know Arabic or Chinese, but the Russian text I added is cut in the middle. The correct text should be: `Съешь же ещё этих мягких французских булок да выпей чаю.`.

**[openfl + lime] Some random game (android):**
photo_2020-04-06_13-24-58

OpenFl gives a more weird result. It removes all spaces and cuts off individual words. For example instead of `Узнай больше` it renders `Узнбол` (`Узн` + `бол`)

### DIFF
--- a/project/src/text/harfbuzz/HarfbuzzBindings.cpp
+++ b/project/src/text/harfbuzz/HarfbuzzBindings.cpp
@@ -258,7 +258,7 @@ namespace lime {
 
 	void lime_hb_buffer_add_utf8 (value buffer, HxString text, int itemOffset, int itemLength) {
 
-		hb_buffer_add_utf8 ((hb_buffer_t*)val_data (buffer), text.c_str (), text.length, itemOffset, itemLength);
+		hb_buffer_add_utf8 ((hb_buffer_t*)val_data (buffer), hxs_utf8 (text, 0), text.length, itemOffset, itemLength);
 
 	}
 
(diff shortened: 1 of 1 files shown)


MSGhero:
Be careful, because at least half of these issues are openfl-related. Even the Mac rendering may be openfl (look at PR 2300).

player-03:
I can confirm that this appears to fix the issue on Linux. And it's straightforward enough, we could consider merging it.

PR #1397, opened 2020-05-12
change __js__ for js.Syntax.code

nanjizal:


nanjizal:
update for haxe 4.1.0, to avoid compile time js warnings.

nanjizal:
@jgranick 
This allows this type of code

```haxe
lime._internal.unifill.Syntax.code('untyped js string');
```
instead of
```haxe
untyped __js__('untyped js string');
```

```haxe
package lime._internal.unifill;
import haxe.extern.Rest;
@:access( js.Syntax.code )
extern class Syntax extends js.Syntax {
    #if( haxe >= "4.0.0" ) 
    public static function code(code:String, args:Rest<Dynamic>):Dynamic {
       return untyped js.Syntax.code( code, args );
    }
    #else
    public inline static function code(code:String):Dynamic {
        return untyped  __js__( code );
    }
    #end
}
```
Would this be a nice backwards compatible approach?

nanjizal:
My main concern about using # code within the main code is that it gets very messy with large chunks of text.  We could just use this for these situations.  The solution was derived by looking inside standard library.

nanjizal:
This seems to work and is more backwards, what version of haxe minimum does OpenFL/Lime support currently?
```haxe
package lime._internal.unifill;
import haxe.extern.Rest;
@:access( js.Syntax.code )
extern class Syntax /*extends js.Syntax*/ {
    #if( haxe >= "4.0.0" ) 
    public static function code(code:String, args:Rest<Dynamic>):Dynamic {
       return untyped js.Syntax.code( code, args );
    }
    #else
    public inline static function code(code:String):Dynamic {
        return untyped  __js__( code );
    }
    #end
}
```

PR #1401, opened 2020-05-25
Visibility-change API implementation added

gyfak:
Show and blur events not worked on mobile browsers with visibility-change api.

https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

Fix for https://community.openfl.org/t/html-how-can-i-query-document-visibilitystate-and-or-listen-to-the-visibilitychange-event/6621

### DIFF
--- a/src/lime/_internal/backend/html5/HTML5Application.hx
+++ b/src/lime/_internal/backend/html5/HTML5Application.hx
@@ -23,6 +23,8 @@ import lime.ui.Window;
 @:access(lime.ui.Window)
 class HTML5Application
 {
+	private static var VISIBILITY_EVENT_NAME:String = "visibilitychange";
+	private static var HIDDEN_PROPERTY_NAME:String = "hidden";
 	private var gameDeviceCache = new Map<Int, GameDeviceData>();
 	private var accelerometer:Sensor;
 	private var currentUpdate:Float;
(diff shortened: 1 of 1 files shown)


src/lime/_internal/backend/html5/HTML5Application.hx
```diff
@@ -270,10 +272,26 @@ class HTML5Application
 
 	public function exec():Int
 	{
+		untyped #if haxe4 js.Syntax.code #else __js__ #end("
+			// Set the name of the hidden property and the change event for visibility
+			var hidden; 
+			if (typeof document.hidden !== 'undefined') { // Opera 12.10 and Firefox 18 and later support
```
player-03, line 278: What is this variable used for?

src/lime/_internal/backend/html5/HTML5Application.hx
```diff
@@ -343,6 +361,17 @@ class HTML5Application
 		return 0;
 	}
 
+	public function checkFlagHidden():Bool
```
player-03, line 364: If this function is going to be public, we need to think carefully about the name because it's going to become a (semi)permanent part of the API.

I don't think "checkFlagHidden" does a good job of describing the function. Someone only looking at the name wouldn't know which flag you were talking about. "checkDocumentHidden" lets you know what is (or isn't) hidden: the HTML document.

I'm also not a fan of "check", mainly because Lime doesn't normally use the word in that way. Based on how Lime uses words, the most appropriate wording would be "getDocumentHidden". Either that or a property with a getter: "`var documentHidden(get, never)`".

src/lime/_internal/backend/html5/HTML5Application.hx
```diff
@@ -448,6 +477,15 @@ class HTML5Application
 					parent.window.onFocusOut.dispatch();
 					parent.window.onDeactivate.dispatch();
 
+				case VISIBILITY_EVENT_NAME:
+					if (checkFlagHidden()==true) {
```
player-03, line 481: `checkFlagHidden()==true` can be simplified to `checkFlagHidden()`.

PR #1407, opened 2020-06-12
Github Actions: also compiles some samples for a bunch of platforms

joshtynjala:
Currently, does not include iOS, but I plan to add that later.

I also consolidated everything into a single job with the operating systems added to the matrix instead of separate jobs for each operating system, which helps avoid too much duplication.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,15 +4,14 @@ on: [push]
 
 jobs:
 
-  build-linux:
-    runs-on: ubuntu-16.04
-
-    env:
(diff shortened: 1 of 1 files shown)


joshtynjala:
Thank you!

Looks like there was some unrelated network trouble during the Android build. 

> Downloading https://services.gradle.org/distributions/gradle-5.6.3-bin.zip
> ##[error]Exception in thread "main" javax.net.ssl.SSLHandshakeException: Remote host closed connection during handshake

Can you re-run the job? Thanks!

Issue #1409, opened 2020-06-19
lime build won't read .build if git is present


giuppe:
Hi,
it seems that it is impossible to use a _.build_ file to keep track of android/ios build number. I can override the build number with `<meta build-number="x" />`, but the old mechanic of a file auto-incremented by `lime build` does not seem to work anymore.

However, it works if I remove the _.git_ directory.

I searched in the code and found this in [CommandLineTools.hx:1250](https://github.com/haxelime/lime/blob/0355103ccf2b9c5b5bf7134cd2cc10f9c4027a4f/tools/CommandLineTools.hx#L1250):

1. it starts with the build number from meta xml, if exists;
1. if empty OR if build number starts with "git", then use number of git commits;
1. if still empty OR if build number starts with "svn". does the same over SVN;
1. if still empty *then* checks the _.build_ file.

This seems wrong, as, if the project is under git or svn, it will never check for the _.build_ file. If the project is under git or svn the build number won't be empty in the last check.

My understanding here is that `lime build` should use git or svn *if and only if* the _project.xml_ has a  `<meta build-number="git" />`. If so, then the OR should be AND:

1. starts with the build number from meta xml, if exists;
1. if empty AND if build number starts with "git", then use number of git commits;
1. if still empty AND if build number starts with "svn". does the same over SVN;
1. if still empty it finally checks the _.build_ file.

I will make a PR if needed.

player-03:
So as discussed in the pull request, this alone doesn't work. The problem is here:

> if empty AND if build number starts with "git", then use number of git commits;

An empty string can't start with "git", so that would never happen. If the user typed `<meta build-number="git" />`, that would mean the string is _not_ empty and also starts with "git".

And that might work, but it would break existing builds that rely on Lime auto-detecting git.

Instead, perhaps we could add a flag to say "ignore the version control; just use the default method." If you set `build-number` to a non-null non-numeric string besides "git" or "svn", that will bypass both git and svn, and then in step 4 we could add a case to catch it. A string like that would previously have been a compile error, so it won't break anyone's code.

PR #1410, opened 2020-06-20
Fix for lime build won't read .build if git is present

giuppe:
This fixes issue #1409.

### DIFF
--- a/tools/CommandLineTools.hx
+++ b/tools/CommandLineTools.hx
@@ -1247,12 +1247,12 @@ class CommandLineTools
 	{
 		var buildNumber = project.meta.buildNumber;
 
-		if (buildNumber == null || StringTools.startsWith(buildNumber, "git"))
+		if (buildNumber == null && StringTools.startsWith(buildNumber, "git"))
 		{
 			buildNumber = getBuildNumber_GIT(project, increment);
 		}
 
-		if (buildNumber == null || StringTools.startsWith(buildNumber, "svn"))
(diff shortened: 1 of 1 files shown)


joshtynjala:
This change causes the following exception when I run `lime rebuild tools`:

```
Called from ? line 1
Called from CommandLineTools.hx line 1895
Called from CommandLineTools.hx line 22
Called from a C function
Called from CommandLineTools.hx line 390
Called from CommandLineTools.hx line 1888
Called from CommandLineTools.hx line 1250
Called from hxp/StringTools.hx line 457
Called from /usr/local/lib/haxe/std/StringTools.hx line 241
Uncaught exception - Invalid field access : length
```

The call to `StringTools.startsWith()` now fails when `buildNumber` is null.

joshtynjala:
I reverted this commit for now so that it doesn't prevent anyone from working on Lime.

giuppe:
I understand why this happens: if `buildNumber` is null, it can't be a string.

So the working code should first check if `buildNumber` is null (and nothing else). 
Then, if it's null, should check for the `.build` file. 
Then, if EITHER it's still null OR it's a string that starts with "git", should calculate the git build number.

The fix should not be about changing OR into AND (that was wrong) but about changing the order of the `if`s
I'm preparing a PR with the fix.

joshtynjala:
Yeah, that sounds like the right way to do it to me.

player-03:
Sounds good.

It's worth noting that git/svn users may have old .build files lying around, so we should mention this in the release notes. "If your build number changes unexpectedly, try removing the .build file from your output folder."

Issue #1414, opened 2020-07-02
lime/tools/CommandLineTools.hx change around line 1250 causing error


intoxopox:
```
Called from ? line 1
Called from CommandLineTools.hx line 1895
Called from CommandLineTools.hx line 22
Called from a C function
Called from CommandLineTools.hx line 125
Called from CommandLineTools.hx line 1888
Called from CommandLineTools.hx line 1250
Called from hxp/StringTools.hx line 457
Called from /usr/local/lib/haxe/std/StringTools.hx line 241
Uncaught exception - Invalid field access : length
```

The offending lines:
```if (buildNumber == null && StringTools.startsWith(buildNumber, "git"))
{
    buildNumber = getBuildNumber_GIT(project, increment);
}

if (buildNumber == null && StringTools.startsWith(buildNumber, "svn"))
{
    buildNumber = getBuildNumber_SVN(project, increment);
}...```
Where, I suspect, meant to read as != null... so:
```if (buildNumber != null && StringTools.startsWith(buildNumber, "git"))
{
    buildNumber = getBuildNumber_GIT(project, increment);
}

if (buildNumber != null && StringTools.startsWith(buildNumber, "svn"))
{
    buildNumber = getBuildNumber_SVN(project, increment);
}...```

player-03:
Closing this because #1409 has been reverted.

PR #1417, opened 2020-07-07
Add flag to read buildNumber from .build.

### LINKED ISSUE #1409: lime build won't read .build if git is present

player-03:
By setting `<meta build-number=".build" />`, you can force Lime not to calculate a build number from git or svn, but instead use the default option of getting it from the .build file.

Since setting this value would previously have produced useless build numbers, it's unlikely that this affects any existing builds.

This should fix #1409 without causing #1414 to happen again.

### DIFF
--- a/tools/CommandLineTools.hx
+++ b/tools/CommandLineTools.hx
@@ -1257,7 +1257,7 @@ class CommandLineTools
 			buildNumber = getBuildNumber_SVN(project, increment);
 		}
 
-		if (buildNumber == null)
+		if (buildNumber == null || buildNumber == ".build")
 		{
 			var versionFile = Path.combine(project.app.path, ".build");
 			var version = 1;


giuppe:
It is a way to resolve the issue; however the historical default, when `meta build-number` is not present, was to read and, if missing, generate the `.build` file; the new default instead becomes "git", as the code will search for git even when `meta build-number` is not present. I don't know if this is ok.

player-03:
I left [a long comment](https://github.com/haxelime/lime/pull/1419#issuecomment-659590608) in #1419 explaining the differences between the two pull requests, and why I believe this one is better.

To make a long story short: using version control is more reliable in general, and if users should have to opt out if they want to use an unversioned file like `.build`. Unless someone comes up with something I missed, or points out a way in which this pull request disrupts existing workflows, I'm going to go ahead and merge it in a week.

PR #1418, opened 2020-07-07
[feature] add removeAll feature for lime event

loudoweb:
quick method to remove all listeners from a lime event

### DIFF
--- a/src/lime/app/Event.hx
+++ b/src/lime/app/Event.hx
@@ -155,4 +155,19 @@ class Event<T>
 		}
 		#end
 	}
+	
+	/**
+		Removes all event listeners
+	**/
+	public function removeAll():Void
+	{
+		#if !macro
+		var len = __listeners.length;
+
+		__listeners.splice(0, len);
+		__priorities.splice(0, len);
+		__repeat.splice(0, len);
+		
+		#end
+	}
 }


loudoweb:
It's easy to test the feature. Here is a sample:

```
package;


import lime.app.Event;
import openfl.display.Sprite;

class Main extends Sprite {
	
	
	var onTest:Event<Bool->Void>;
		
		public function new() {
			super();
			onTest = new Event<Bool->Void>();
			
			onTest.add(test1);
			onTest.add(test2);
			onTest.add(test3);
			
			trace("init test : result should be 3 count");
			onTest.dispatch(true);
			trace("remove test : result should be 2 count");
			onTest.remove(test1);
			onTest.dispatch(true);
			trace("remove all test : result should be 0 count");
			onTest.add(test1);
			onTest.removeAll();
			onTest.dispatch(true);
		}
		static function test1(b:Bool):Void
			{
				trace(b);
			}
		static function test2(b:Bool):Void
			{
				trace(b);
			}
				
				
		static function test3(b:Bool):Void
			{
				trace(b);
			}
}
```
And here is the log you should get:

```
Source/Main.hx:20: init test : result should be 3 count
Source/Main.hx:32: true
Source/Main.hx:36: true
Source/Main.hx:42: true
Source/Main.hx:22: remove test : result should be 2 count
Source/Main.hx:36: true
Source/Main.hx:42: true
Source/Main.hx:25: remove all test : result should be 0 count
```

PR #1415, opened 2020-07-06
getBuildNumber "git" and "svn" check: != instead of ==. Fixes #1414.

intoxopox:

PR #1419, opened 2020-07-07
Fixes #1409 and prevents #1414

### LINKED ISSUE #1409: lime build won't read .build if git is present

giuppe:
I gave a look at the git history of the piece of code that computes the build number; this PR is based on the following interpretation of what the default behaviour of `lime build` should be:
- if `meta build-number` is not set, then it should read the `.build` file and, if missing, it should create it;
- if `meta build-number` is set to "git*", then it should calculate the build number from `git rev-list`;
- if `meta build-number` is set to "svn*", then it should calculate the build number from svn revision number;
- if `meta build-number` is set to something else, then it should use this as the build number.

The code, currently, does this:
- if `meta build-number` is not set, the build number will be calculated from git;
- if `meta build-number` is not set and git is not found, the build number will be calculated  from svn;
- if `meta build-number` is not set and neither git nor svn are found, the build number will be read from the `.build` file;
- if `meta build-number` is set to "git*", the build number will be calculated from git;
- if `meta build-number` is set to "svn*", the build number will be calculated from svn;
- if `meta build-number` is set to "whatever", the build number will be "whatever".

The code is changed in this way:
- first, checks if the buildNumber is null (i.e. the meta build-number is not set in the project config);
- if `buildNumber` is null it reads the `.build` file and creates it if it's missing;
- else if `buildNumber` is a string starting with "git", it calls the git version of getBuildNumber;
- else if `buildNumber` is a string starting with "svn", it calls the svn version of getBuildNumber

What should happen:
- `meta build-number` is not set -> `.build` file;
- `meta build-number` set to "git*" -> git rev;
- `meta build-number` set to "svn*" -> svn rev;
- `meta build-number` set to "whatever" -> "whatever".

Fixes #1409, prevents #1414; supersedes #1410 and #1417.

### DIFF
--- a/tools/CommandLineTools.hx
+++ b/tools/CommandLineTools.hx
@@ -1247,16 +1247,6 @@ class CommandLineTools
 	{
 		var buildNumber = project.meta.buildNumber;
 
-		if (buildNumber == null || StringTools.startsWith(buildNumber, "git"))
-		{
-			buildNumber = getBuildNumber_GIT(project, increment);
-		}
-
-		if (buildNumber == null || StringTools.startsWith(buildNumber, "svn"))
-		{
-			buildNumber = getBuildNumber_SVN(project, increment);
-		}
-
 		if (buildNumber == null)
 		{
 			var versionFile = Path.combine(project.app.path, ".build");
@@ -1295,6 +1285,14 @@ class CommandLineTools
 				}
 				catch (e:Dynamic) {}
 			}
+		}
+		else if (StringTools.startsWith(buildNumber, "git"))
+		{
+			buildNumber = getBuildNumber_GIT(project, increment);
+		}
+		else if (StringTools.startsWith(buildNumber, "svn"))
+		{
+			buildNumber = getBuildNumber_SVN(project, increment);
 		}
 	}
 


player-03:
Hmm. This seems to be exactly equivalent to #1415, which was rejected. @joshtynjala, care to take another look?

player-03:
That said, good writeup.

I'd also like to quote your comment from #1417 so we can discuss it here.

> It is a way to resolve the issue; however the historical default, when `meta build-number` is not present, was to read and, if missing, generate the `.build` file; the new default instead becomes "git", as the code will search for git even when `meta build-number` is not present. I don't know if this is ok.

The logic makes sense, but it depends on when things happened. If the "new default" is only a month or two old, we could apply this commit without messing up too many workflows. However, this behavior [has been around for three and a half years](https://github.com/haxelime/lime/commit/f0c6b4dcf3d3e497e9e174e2ff28be00b869163d), which means modern projects are more likely to rely on it than to use the .build file.

player-03:
To help make a decision, let's lay this out in detail.

The two problems
---
In past releases of Lime, builds were counted using a `.build` file stored in the export directory. This file was incremented after every build. This strategy ensured that different builds would have different build numbers, and that the build number would never decrease. The main drawback is that it wasn't easy to use with multiple collaborators working on a single project: if you didn't go to great lengths to synchronize your `.build` files, you risked having redundant (or even decreasing!) build numbers.

Around three and a half years ago, two new build-counting strategies were added. These strategies checked whether the project was a Git or Subversion repository, and if so, counted the number of commits. While these strategies don't guarantee two separate builds have different build numbers, they perform far better in a collaborative environment.

Even so, some developers prefer using a `.build` file with Git or SVN. They might want the guarantee of a different build number for every build, or perhaps the number in their `.build` file is far higher than their number of commits, meaning that if they switch systems their build number will suddenly drop. (Which could lead to the Google Play Store rejecting the app.) Thus, we have two problems:

- **Problem 1:** In the current release version of Lime, there's no way to use the `.build` file in a Git or SVN repository.
- **Problem 2:** Changing build-counting strategies will usually cause the version number to change in unexpected and unwanted ways. (This has already happened 3.5 years ago, but it would be nice to keep it from happening again. It also happens whenever a project is created without Git and then Git is added.)

The three solutions
---
I can think of three reasonable ways to handle this, though I'm open to other ideas.

- **Solution 1:** Use Git or SVN by default, but allow users to manually request using the `.build` file instead. This preserves the behavior that's been in place for the past three years. Submitted as pull request #1417.  
Pros: fixes problem 1, avoids problem 2 for modern projects (i.e. projects from the past three years that use version control).  
Cons: still causes problem 2 in specific cases.
- **Solution 2:** Use the `.build` file by default, but allow users to manually request Git or SVN instead. This restores the historical default behavior. Submitted as pull request #1419 (this one).  
Pros: fixes problem 1, avoids problem 2 in specific cases.  
Cons: likely to cause problem 2 in modern projects.
- **Solution 3:** Use the `.build` file if it already exists, then use Git or SVN if found, and create a `.build` file if all else fails. This is kind of a middle ground between the first two solutions, trying to intelligently guess which strategy to use. No pull request has been submitted for this yet.  
Pros: fixes problem 1, almost completely avoids problem 2.  
Cons: if a project is shared across machines, only one of which has a `.build` file, version numbers will vary wildly and unexpectedly.

My thoughts
---
It should come as no surprise that I support the pull request that I submitted. It could just be bias, but hear me out.

I think that three years is long enough for most developers to have gotten used to version-control-based counting, and that changing the default at this point will cause more harm than good.

And though solution 3 sounds good at first, I think it ends up showing the dangers of relying on an unversioned file. Imagine you're the collaborator on a project, and your machine doesn't have a `.build` file. You compile a build, but it gets rejected from Google Play because the build number is so much lower than what's already on Google Play. But the lead developer isn't even aware of the `.build` file; it was created by default before they set up version control, and it's been used by default ever since. How do you even start to debug that?

Solution 2 helps by keeping things consistent across computers. Either everyone uses `.build` or no one does. But even then, by default you'll use independent copies of `.build`. As I mentioned, it's unversioned unless you go to extra effort to include it in version control. But if no one on the team notices that the file exists, they won't know to commit it, and sooner or later someone will run into build number problems.

Solution 1 helps immensely by making it opt-in. If you're using the `.build` file, you're doing so intentionally and you know what it does. Even if the person who made that decision leaves the team, there's still an enormous hint in the form of a `<meta build-number=".build" />` tag in project.xml.

Conclusion: none of the three solutions I listed is perfect, but solution 1 seems quite a bit better than the other two.

giuppe:
So either option isn't perfect and would break workflows. Also in the cons of git I think that `git ref-list` could *decrease* in case of rebases and branch deletions. 

I now feel that neither option is the "saner default", moving forward. Either could have its tag ("build", "git", etc) but maybe we should find something that could be the "path of least surprise" for the newbie of the future. Like Apache Cordova, where, if you don't override the build number, it is computed from the version number (with a formula like 10000*major + 100*minor + patch + other).

player-03:
It's a good thought, but I'm not sure there _is_ a path of least surprise here. (At least, not one we'd want to take.) Yes the build number could decrease using Git, but it could just as easily decrease using `.build`. Cordova's method would keep it from decreasing, but like solutions 2 and 3, we'd be changing behavior that's been in place for 3 years, which is _very_ surprising. Putting the `.build` file outside the build folder would also help, since then it would be tracked by default, but in addition to being a sudden change, I can only assume they considered and rejected this possibility back when they first implemented `.build`.

It may be worth making a big change, but I'd argue that should only happen during a major release, when breaking changes are to be expected.

player-03:
Ultimately, I decided to go with the smaller change. Not to say that the two are mutually exclusive, but I don't have time to resolve merge conflicts, debug, or provide tech support to anyone whose build was affected. I can verify at a glance that my one-line pull request won't break anything (that wasn't already broken), whereas it turns out that this one had a bug for five months that neither of us noticed.

Both `getBuildNumber_SVN()` and `getBuildNumber_GIT()` are capable of returning `null`, and if they do, that value will be used as-is. It was only luck that I happened to notice this bug, and though I think it's the only one, I'm not sure.

PR #1436, opened 2020-10-14
Add failIfMajorPerformanceCaveat setting

Beeblerox:
Recent update of Google Chrome has changed behavior of `failIfMajorPerformanceCaveat` option for creation of WebGL context - it fails more frequently for old videocards.
This change add setting for `failIfMajorPerformanceCaveat`, which is `false` by default now, but you could change it with "project.xml" or "project.hx" files by adding attribute `failIfMajorPerformanceCaveat` to the window

### DIFF
--- a/src/lime/_internal/backend/html5/HTML5Window.hx
+++ b/src/lime/_internal/backend/html5/HTML5Window.hx
@@ -287,7 +287,7 @@ class HTML5Window
 						premultipliedAlpha: true,
 						stencil: Reflect.hasField(contextAttributes, "stencil") ? contextAttributes.stencil : false,
 						preserveDrawingBuffer: false,
-						failIfMajorPerformanceCaveat: true
+						failIfMajorPerformanceCaveat: Reflect.hasField(contextAttributes, "failIfMajorPerformanceCaveat") ? contextAttributes.failIfMajorPerformanceCaveat : false,
 					};
 
 				var glContextType = ["webgl", "experimental-webgl"];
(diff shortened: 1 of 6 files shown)


Aboideau:
This setting now affects Firefox 83 (I currently only tested this on Linux Mint) as I can't run Away3d samples without getting the `Failed to create WebGL context: failIfMajorPerformanceCaveat: Compositor is not hardware-accelerated.` error.

When reading the bug report [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1679671), I saw these interesting snippets:

> Honestly it sounds like failIfMajorPerformanceCaveat: true used to not work, and now it's working.
> The minimal testcase explicitly asks for failIfMajorPerformanceCaveat: true, which if you have non-accelerated compositing, will cause context creation to fail. (by design and by spec)

and

> This is the expected outcome on Linux. Without hardware compositing, contexts created with failIfMajorPerformanceCaveat will fail. If want it back you can either set webgl.disable-fail-if-major-performance-caveat=true to disable the check or gfx.webrender.all=true to turn on accelerated compositing.

True enough, if I change the `failIfMajorPerformanceCaveat` to `false` then everything works, but as of the current `haxelib lime` version, you have to change the JS file manually.

Furthermore, 
>The answer for frustrated website authors is straightforward: Don't ask for failIfMajorPerformanceCaveat: true unless you do want context creation to fail for many many users. failIfMajorPerformanceCaveat has no other use besides this "bug", so don't ask for it if you don't want this behavior.

seems like an issue begging a new version to be pushed to `haxelib`.

player-03:
I'm seeing the same issues in Firefox 84 on Ubuntu. Definitely worth fixing.

I'd love to help get this out the door, but I've never been a part of that process. @jgranick, is it time for a release?

PR #1447, opened 2021-01-27
rootPath-related improvements

chatziko:
This PR includes the following `rootPath`-related improvements:

- For pak libraries, `/` is always appended to `rootPath`, even it already ends with `/` (which is the case of `rootPath`'s default value `./`). This leads to the pak URL containing double `//`, which web servers might or might not serve correctly. The PR fixes this via `Path.join`.

- In `ManifestResources` there are various instances of `rootPath + <some path>` which fail if `rootPath` does not end with `/`. The PR appends a `/` in such case (it might be preferable to use `Path.join` everywhere, but I wanted to keep the changes minimal).


player-03:
I considered suggesting `Path.addTrailingSlash()`, but I don't know if it's worth importing `Path` just for that. Honestly it looks good to me the way it is.

jgranick:
I added a commit to help resolve the double-slash issue:

https://github.com/haxelime/lime/commit/105cc69cb7d9f495f5a353a3359f4436f3519e8c

Always adding a slash to root path will cause trouble, particularly on Android which does not resolve paths as well as a standard file system. "myfile.png" and "/myfile.png" are not the same. It may be helpful to discuss in what cases this is failing

player-03:
Huh, the web UI didn't allow me to customize the commit message. I would've added that I cleaned up a bit in the process of merging.

PR #1448, opened 2021-02-13
Feature/haxe 4 2 support

m0rkeulv:
Here's an attempt to make Lime build with Haxe 4.2
It looks like the behavior for Abstracts has been changed a bit. 
- `@:transitive` is necessary for "chained" implicit casts  
- it looks like  `@:generic`  is not allowed on abstracts constructors.
- Not sure if it affects all kinds of abstracts or just some cases, but the compiler won’t  accept  uninitialized  variables of some abstract types.

i also upgraded the hxcpp version in the workflow.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install Haxe dependencies
         run: |
-          haxelib install hxcpp 4.0.64 --quiet
+          haxelib install hxcpp 4.2.1 --quiet
           haxelib install format  --quiet
           haxelib install hxp  --quiet
           haxelib git lime-samples https://github.com/openfl/lime-samples --quiet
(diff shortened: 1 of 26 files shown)


CanIGetaPR:
Hey you committed some code changes to HBBuffer in https://github.com/haxelime/lime/pull/1448/commits/59ddb9636bca6618a6d9dba08478fb5bdb5696bc was that intentional?

m0rkeulv:
> 
> 
> Hey you committed some code changes to HBBuffer in [59ddb96](https://github.com/haxelime/lime/commit/59ddb9636bca6618a6d9dba08478fb5bdb5696bc) was that intentional?

Good catch, that was not intended to be a part of the commit ( i have some tweaks on my own branch to speed up text rendering and i apparently did not shelf  that part before committing :) 

i have removed them now

player-03:
Looks pretty straightforward. Unless anyone else sees anything, I think we should be good to merge.

maitag:
looks good and can be merged

PR #1449, opened 2021-02-22
Fix the "Can't cast bytes to String" error in FileDialog.hx

KGriss:


player-03:
It looks like the `String.fromUTF8()` only exists while targeting HashLink. And even if it existed on other platforms, it isn't a public function which means it's subject to change. A better solution might be using [`Bytes.toString()`](https://api.haxe.org/haxe/io/Bytes.html#toString) - as a public method with API documentation, it isn't going anywhere any time soon.

My other concern is that since `result` isn't explicitly typed, it might be a different type on other platforms. Perhaps on most platforms it's already a string, and that's why this bug was never found. Fortunately, if you call `toString()` on a string, it'll just return itself and no harm will be done.

maitag:
i did a new lime-sample for better target-testing here: <https://github.com/maitag/lime-samples/tree/master/features/file-io/filedialog>

maitag:
If i test `var path:String = result.toString();` (haxe 4.2, hashlink 1.11, win7) it gives me error: " Can't cast bytes to virtual<toString:(String ())>".
The `String.fromUTF8()` works with a hl-compilercondition at that point but have some issue if the filename contains unicode-chars (like german "äöüß"). Is there a way to convert it into an UnicodeString ? (maybeif via: <https://api.haxe.org/hl/Bytes.html#utf16ToUtf8>?... have no idea how to set the parameters for that function to test out -_-)

Apprentice-Alchemist:
> It looks like the `String.fromUTF8()` only exists while targeting HashLink. And even if it existed on other platforms, it isn't a public function which means it's subject to change. A better solution might be using [`Bytes.toString()`](https://api.haxe.org/haxe/io/Bytes.html#toString) - as a public method with API documentation, it isn't going anywhere any time soon.
> 
> My other concern is that since `result` isn't explicitly typed, it might be a different type on other platforms. Perhaps on most platforms it's already a string, and that's why this bug was never found. Fortunately, if you call `toString()` on a string, it'll just return itself and no harm will be done.
1. `hl.Bytes` does not have a toString method.
2. `String.fromUTF8` is used extensively in 'official' libs such as hlsdl, so I think it's safe to use. 
> The `String.fromUTF8()` works with a hl-compilercondition at that point but have some issue if the filename contains unicode-chars (like german "äöüß"). Is there a way to convert it into an UnicodeString ? (maybeif via: https://api.haxe.org/hl/Bytes.html#utf16ToUtf8?... have no idea how to set the parameters for that function to test out -_-)

What exactly is the issue?

player-03:
You're right about `hl.Bytes`. Kind of annoying that they have two incompatible classes with the same name, but it's hardly the first time.

I dug deeper into `Std.string()`, and under the hood it calls the Ocaml function `utf8_to_utf16()`, which is the same thing `String.fromUTF8()` ends up calling. That's promising for forward compatibility, but it doesn't solve the Unicode character problem. (I didn't suggest `Std.string()` in my last post because the documentation simply says it calls `toString()`, so I figured they were equivalent. Not on this platform!)

Since all HL strings are stored in UTF-16, you may be able to solve the problem by converting to UTF-8 only for `Std.string()` to convert back. Kind of inelegant, though. `String._alloc_()` would be a simpler solution: that's the function for making a string without modifying the underlying bytes in any way, and it's a much closer analogue to `var path:String = cast result`. (You do have to specify a length, so try `ucs2Length(0)`. The implementation loops through the string until it hits a "`\0`" value, which seems right.)

maitag:
> What exactly is the issue?

The issue is how result of `NativeCFFI.lime_file_dialog_open_file()` (<https://github.com/haxelime/lime/blob/develop/src/lime/ui/FileDialog.hx#L50>) is to encode for some unicode characters e.g.: `schön` -> `sch` or `äöüß.txt` -> `巼`
So maybe because windows is using utf16 for filesnames or something behind the cffi-bindings? (have no idea about cffi)
On cpp and neko targets it works well!

maitag:
a dirty fix for me at now is into using the unifill haxelibrary:
```
#if hl
@:privateAccess String.fromUTF8( result );
var path:String = unifill.Utf16.fromArray([for (i in 0...result.ucs2Length(0)*2) result.getUI8(i) ] ).toString();
#else
var path:String = cast result;
#end
```

jgranick:
I think that if the returned value is not a String, the issue is in ExternalInterface.cpp (hl_lime_file_dialog_open_directory, etc) and not in the Haxe code here

Apprentice-Alchemist:
`hl_lime_file_dialog_open_directory` and others return `vbyte*`, ie `hl.Bytes`.
As far as I can see those functions return utf8 C strings.
So `var path:String = #if hl @:privateAccess String.fromUTF8(result) #else cast result #end;` should be the correct solution.

jgranick:
Yeah, it's been a while... you're right

jgranick:
Merged, thank you!

PR #1452, opened 2021-03-14
fix: Shader compiler is giving warnings only when there are errors present

### LINKED ISSUE #1451: Shader compiler is only giving warnings when there are errors present
The [shader compiler in its function `fromSource()`](https://github.com/haxelime/lime/blob/5345430d0f9984bc7a141a51567f146e5a2a7c78/src/lime/graphics/opengl/GLShader.hx#L24)` is using two GL functions to report on errors and warnings during compilation:
- `getShaderParameter(shader, COMPILE_STATUS)`, that returns 0 (GL_FALSE) on errors and 1 (GL_TRUE) on success; if this returns zero, then
- `getShaderInfoLog(shader)` is executed, that returns a string with the message from the compiler.

The problem is: when there are only warnings and no errors, `getShaderParameter(shader, COMPILE_STATUS)` will return GL_TRUE and so the reporting code is not executed.
I recently had a case where an implicit cast, in a shader, would be treated as a warning on my machine (and Lime wasn't reporting anything), while the program instantly crashed on another machine that treated it as an error. 

The OpenGL Reference is noting that [we should get the info log even when there aren't errors, if we're debugging a shader](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGetShaderInfoLog.xhtml):

> The information log for a shader object is the OpenGL implementer's primary mechanism for conveying information about the compilation process. Therefore, the information log can be helpful to application developers during the development process, even when compilation is successful. 

So, if we want warnings while in debug mode, we should instead check if the `getShaderInfoLog(shader)` is not null, and only then call `getShaderParameter(shader, COMPILE_STATUS)` to separate errors from warnings.

I could be very difficult to debug shaders without this information, as there is too much variance in how opengl compilers are implemented. 

I can make a PR, if needed.

giuppe:
Fixes #1451.
The new code uses both `getShaderInfoLog()` and `getShaderParameter()` to recognize the following states:

- no errors, no warnings: (`shaderInfoLog = null`, `compileStatus = 1`) --> show nothing;
- no errors,  warnings present: (`shaderInfoLog != null`, `compileStatus = 1`) --> show "Info compiling shader..." & infoLog on Log.debug();
- errors present: (`shaderInfoLog != null`, `compileStatus = 0`) --> show "Error compliling shader..." & infoLog on Log.error().

### DIFF
--- a/src/lime/graphics/opengl/GLShader.hx
+++ b/src/lime/graphics/opengl/GLShader.hx
@@ -26,18 +26,27 @@ public static function fromSource(gl:WebGLRenderContext, source:String, type:Int
 	var shader = gl.createShader(type);
 	gl.shaderSource(shader, source);
 	gl.compileShader(shader);
+	var shaderInfoLog = gl.getShaderInfoLog(shader);
+	var compileStatus = gl.getShaderParameter(shader, gl.COMPILE_STATUS);
 
-	if (gl.getShaderParameter(shader, gl.COMPILE_STATUS) == 0)
+	if (shaderInfoLog != null || compileStatus == 0)
 	{
 		var message;
 
-		if (type == gl.VERTEX_SHADER) message = "Error compiling vertex shader";
-		else if (type == gl.FRAGMENT_SHADER) message = "Error compiling fragment shader";
-		else
-			message = "Error compiling unknown shader type";
+		if (compileStatus == 0)	message = "Error ";
+		else message = "Info ";
 
-		message += "\n" + gl.getShaderInfoLog(shader);
-		Log.error(message);
+		if (type == gl.VERTEX_SHADER) message = "compiling vertex shader";
+		else if (type == gl.FRAGMENT_SHADER) message = "compiling fragment shader";
+		else
+			message = "compiling unknown shader type";
+
+		message += "\n" + shaderInfoLog;
+
+		if(compileStatus == 0)
+			Log.error(message);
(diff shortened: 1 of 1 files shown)


player-03:
Haven't tested this myself, but it all looks correct.

loudoweb:
@giuppe do you have a simple shader sample that I can test (a shader with warnings and no errors)?

giuppe:
This example on my machine (Win10, nvidia driver 461.72, gpu Geforce GT1030) gives me no error and runs ok:

    void main(){
         float temp = 1;
         gl_FragColor = vec4(temp);
    }

but it contains an implicit cast from int to float, that fails on other machines. If I add an error, like

    void main(){
         float temp = 1;
         gl_FragColor = vec3(temp);
    }

I get both the error and the warning:

    warning C7011: implicit cast from "int" to "float"
    error C1035: assignment of incompatible types

loudoweb:
I've tested your [PR in openfl ](https://github.com/openfl/openfl/pull/2473) and confirmed that it's ok.

Regarding this Lime PR, it should be also ok since it's the same code ([without the source log](https://github.com/openfl/openfl/pull/2473/files#diff-7659020478f1c35d995e23f7e3792c413053eb648f368d4dc363137dd10fda44R333)) but I didn't test it (I guess we need to use Lime only without openfl to test this).

PR #1453, opened 2021-03-17
Fix "adb devices" stdout redirection to file

414n:
The current implementation does not correctly work on non-Windows platforms because the `>` redirection operator must be interpreted by a shell, while the `System.runCommand()` function runs the command directly on the system with no shell involved.
This enables the stdout redirection used in `AndroidHelper.listDevices()` to work on non-Windows platforms too.

Tested on a Linux system running Haxe-4.2.1. If anyone could test it on a MacOS system too that would be useful.

PS: can anyone explain me why there's the need to redirect the command output to file in the first place, instead of reading it back as a String like the Windows codepath does?

### DIFF
--- a/src/lime/tools/AndroidHelper.hx
+++ b/src/lime/tools/AndroidHelper.hx
@@ -305,7 +305,7 @@ class AndroidHelper
 		{
 			var tempFile = System.getTemporaryFile();
 
-			System.runCommand(adbPath, adbName, ["devices", ">", tempFile], true, true);
+			System.runCommand('/bin', 'sh',  ["-c", '${adbPath}${adbName} devices > ${tempFile}'], true, true);
 
 			if (FileSystem.exists(tempFile))
 			{
--- a/src/lime/tools/AndroidHelper.hx
+++ b/src/lime/tools/AndroidHelper.hx
@@ -303,20 +303,14 @@ class AndroidHelper
 
 		if (System.hostPlatform != WINDOWS)
 		{
-			var tempFile = System.getTemporaryFile();
-
-			System.runCommand('/bin', 'sh',  ["-c", '${adbPath}${adbName} devices > ${tempFile}'], true, true);
-
-			if (FileSystem.exists(tempFile))
-			{
-				output = File.getContent(tempFile);
-				FileSystem.deleteFile(tempFile);
-			}
+			/*
+			 * using System.runProcess on *NIX platforms for `adb devices` usually
+			 * hangs when adb also starts the server.
+			 * To avoid this, we start the server beforehand
+			 */
+			System.runCommand(adbPath, adbName, ["start-server"], true);
 		}
-		else
-		{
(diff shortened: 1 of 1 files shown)


player-03:
I wasn't there, but in my testing I've found two possible reasons. First, if ADB isn't in your path, you'll get an error saying `adb: command not found`, and it'll crash.

Second, if ADB isn't currently running, `runProcess()` ends up looping indefinitely, waiting for ADB to stop. I don't know why it does this, but since `runCommand()` works differently, I was able to find a workaround.

```haxe
if (System.hostPlatform != WINDOWS)
{
	System.runCommand(adbPath, "./" + adbName, ["start-server"], true);
	output = System.runProcess(adbPath, "./" + adbName, ["devices"], true, true);
}
else
{
	output = System.runProcess(adbPath, adbName, ["devices"], true, true);
}
```

It isn't as clean as I'd like, but it's cleaner than the existing code.

player-03:
The problem with this implementation is it's still missing `"./"` before `adbName`, so it won't work if ADB isn't in the path.

414n:
> The problem with this implementation is it's still missing `"./"` before `adbName`, so it won't work if ADB isn't in the path.

Sorry, but I fail to see the problem with adb being in `PATH` or not and needing a `./` prefix.

**Please do correct me if my thinking is wrong somewhere, I'm still a haxe and lime newbie**

The `adbPath` field is only set inside the `AndroidHelper.initialize()` method and won't ever be set to a sensible value if the `ANDROID_HOME` environment variable is not set to a valid Android SDK path, regardless of any value in `PATH`.
In fact, if no `ANDROID_HOME` env variable is present:

 1. the `AndroidHelper.initialize()` method will set the `adbPath` field to `null/platform-tools`, which is -- usually -- a non-existent path;
 2. calling `System.runCommand(adbPath, adbName, ["start-server"], true);` won't even try to launch the executable whichever it may be, because the `safeExecute` flag set will enforce a valid path check beforehand that will result in an error being logged and bailing out from the `runCommand` method altogether.

The consequence from 1. is that `adbPath` can't be an empty string or null with the current code, implying that no `PATH` lookup for the value of `abdName` will ever take place.
Besides, the `AndroidHelper.initialize()` method contains this snippet:
```haxe
if (System.hostPlatform != WINDOWS)
{
	adbName = "./" + adbName;
	androidName = "./" + androidName;
	emulatorName = "./" + emulatorName;
}
```
where a `./` is prepended to `adbName`, probably because it is (correctly) assumed that the following `runCommand/runProcess` functions will first change directory to `adbPath` and then launch the executable.

player-03:
Ah, in that case I withdraw my objection. I only tested the code snippet in isolation, and didn't review `initialize()`.

414n:
No problem, I liked your suggestion and implemented it right away.
I did test this on another branch I have where the `AndroidHelper.listDevices()` function is used to check if the `--device` argument is a valid adb USB device, and it seems to work fine.

PR #1456, opened 2021-03-28
Making NativeAudio buffers configurable &  playback  behaving like other sources.

m0rkeulv:
In order to make  music streaming usable in OpenFL on native targets some tweaks to Lime is necessary.

1. when using stop, expected behavior in OpenFL is to return to the beginning of a track ( the current behavior is more like a pause )  so i have changed this  (see line 312)

2. when the backend runs out of buffer it stops playing (and seems to refuse to continue), this becomes an issue when the main thread is occupied for too long (loading assets, resizing window etc.) so i have added check to see if it was playing before it ran out of buffer and if so, resume playback. (line 296 - 302)

3. Since the buffers are quite small and only 3 by default  any small holdup in the main thread might cause stuttering, i have therefor added an option to set a different amount of buffers when compiling  by setting `native_audio_buffers`to  the desired amount.

### DIFF
--- a/src/lime/_internal/backend/native/NativeAudioSource.hx
+++ b/src/lime/_internal/backend/native/NativeAudioSource.hx
@@ -19,7 +19,11 @@ import lime.utils.UInt8Array;
 class NativeAudioSource
 {
 	private static var STREAM_BUFFER_SIZE = 48000;
+	#if !(!native_audio_buffers || macro)
 	private static var STREAM_NUM_BUFFERS = 3;
+	#else
+	private static var STREAM_NUM_BUFFERS =  Std.parseInt(haxe.macro.Compiler.getDefine("native_audio_buffers"));
+	#end
 	private static var STREAM_TIMER_FREQUENCY = 100;
 
 	private var buffers:Array<ALBuffer>;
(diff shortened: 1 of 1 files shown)


player-03:
I found what I could, but I'm not too familiar with OpenAL so I may have missed something.

src/lime/_internal/backend/native/NativeAudioSource.hx
```diff
@@ -294,9 +294,10 @@ class NativeAudioSource
 
 			AL.sourceQueueBuffers(handle, numBuffers, buffers);
 
-			// If openAL runs out of buffer it will stop playback.
-			// This check is here to recover from this and resume playback
-			// this situation typically happens when resizing window or other operations that freezes the main thread
+			// OpenAL can unexpectedly stop playback if the buffers fill up,
```
m0rkeulv, line 297: Not sure this is the right wording.
There's a loop that make sure there is always data to be read  for the audio engine, when the main thread "freezes" this loop stops feeding  the buffers so  it runs out of data. it stops playing because there's no more data to be read, not because the buffers are filled with data.

justin-espedal:
> when using stop, expected behavior in OpenFL is to return to the beginning of a track ( the current behavior is more like a pause ) so i have changed this (see line 312)

I'm not sure this is the right implementation (see my comments here https://github.com/haxelime/lime/commit/feae6af85a437bd00efbd277cf5bc5dec4bd9daf#r51910077)

What types of sounds was this change tested with? Did you never experience any sound-repeating behavior?

src/lime/_internal/backend/native/NativeAudioSource.hx
player-03: Oh. That's probably true now that I take another look at it. I was trying to keep the original meaning, but I interpreted "runs out of buffer" to mean "runs out of space in the buffer" (full), but it could also mean "runs out of data in the buffer" (empty).

PR #1465, opened 2021-06-09
Fix playback behavior in `NativeAudioSource`

player-03:
The sequel to pull #1456, this request fixes two issues brought up in that thread. I would love to have amended that request, but I didn't feel like spending the time to figure out how.

Anyway, the big issue is that sound could potentially play despite `playing` being false. Besides that, one of the comments was incorrect.


justin-espedal:
I heard from the developer who complained about sound playback, said it’s working great for him again. Thanks!

PR #1470, opened 2021-06-25
Don't let packed asset files begin with a compression header

justin-espedal:
Packed asset libraries are sequences of individual files, sometimes compressed, and concatenated. If each file is compressed, this causes the asset pack as a whole to begin with compressed file headers, which may cause some software to mistake the file as being one large compressed file, which it isn't.

For example, some html5 web portals (in particular, the very popular itch.io) may try to automatically determine if a file is gzip-compressed, and if they are, serve the file with the "content-encoding: gzip" header. That will cause a web browser to decompress the file while downloading it, leading to two issues:

1. The compressed file header is only for the first file in the pack, so we'll only receive the data for the first file.
2. The data will be in an already-decompressed state, which we don't expect.

To illustrate, the file looks like this:
```
[                               default.pak                                           ]
[packed-asset-1.gzip][packed-asset-2.gzip][packed-asset-3.gzip]...[packed-asset-N.gzip]
```

What we receive from the download is this:
```
[asset-1-already-unzipped] ... nothing else
```

This will manifest as an error like "Uncaught incorrect header check" when trying to load and decompress what we expect to be compressed data.

This change prepends the string "asset-pack" to the start of the file. It's applied to all pack types, not just compressed ones. For uncompressed packs, I presume that there's the possibility that whatever file header happens to be present on the first file in the pack has other inadvertent effects.

### DIFF
--- a/src/lime/tools/AssetHelper.hx
+++ b/src/lime/tools/AssetHelper.hx
@@ -295,6 +295,13 @@ class AssetHelper
 		if (asset.type == TEMPLATE) return null;
 		if (asset.library == library.name || (asset.library == null && library.name == DEFAULT_LIBRARY_NAME))
 		{
+			if (output.tell() == 0)
+			{
+				//write some dummy text at the start of the packed asset file just to prevent
+				//the file from beginning with a packed file header.
+				output.writeString("asset-pack");
+			}
+
 			var assetData:Dynamic =
 				{
 					id: asset.id,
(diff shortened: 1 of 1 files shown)


player-03:
This all makes sense to me.

I'm tempted to say the string should be "lime-asset-pack" or "lime-assets" or something, for better disambiguation. It's unlikely but not inconceivable that someone out there already used the string "asset-pack."

justin-espedal:
Yeah, sounds reasonable enough. I've made that change.

PR #1475, opened 2021-08-30
Added "-listen" flag to AIR build params

Borman2000:
Build AIR app with _-listen_ flag instead of _-connect_ to debug AIR app on device **over the usb** (default port 7936)


player-03:
The first and third commits look fine at a glance, but 9f4f124 isn't a great way to do this.

- All projects have to keep track of `assetsFolders`, even though only AIR projects need it.
- Since you're passing a whole directory, AIR will ignore any asset filters (include/exclude) from project.xml.
- AIR doesn't know how to process Haxe templates, so any of those are present it may break the build.

I do see the reasoning here. Windows has a hard limit on command line length, and their [suggestions to work around it](https://docs.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/command-line-string-limitation) amount to "pass less data" and "store your data in a file for the program to read." Passing less data isn't a great option (as described above), so it would be nice to use the second option. But of course [AIR doesn't support it](https://help.adobe.com/en_US/air/build/WS901d38e593cd1bac1e63e3d128fc240122-7ff2.html).

How about this for a solution?

1. Collect the assets individually (instead of using `assetsFolders`).
2. Create a temporary folder. Call it `targetDirectory + "/assets"` or something.
3. Copy all non-embedded assets to this folder, one by one. So instead of `files.push(asset.targetPath)`, you run `AssetHelper.copyAsset(asset, targetDirectory + "/assets")`.
4. Now that the temporary folder has the correct assets, pass it to AIR: `files.push(targetDirectory + "/assets")`.

externs/air/sys/FileSystem.hx
```diff
@@ -37,5 +37,5 @@ class FileSystem
 
 	public static function deleteDirectory(path:String):Void {}
 
-	public static function readDirectory(path:String):Array<String> {}
+	public static function readDirectory(path:String):Array<String> {return [];}
```
player-03, line 40: Not necessary. If this is causing errors, something is set up incorrectly on your end.

src/lime/tools/AIRHelper.hx
```diff
@@ -146,14 +149,22 @@ class AIRHelper
 
 			if (project.debug)
 			{
-				args.push("-connect");
+				if(project.targetFlags.exists("listen"))
+					args.push("-listen");
+				else
+					args.push("-connect");
```
player-03, line 155: Should follow Lime's formatting conventions.

src/lime/tools/AIRHelper.hx
```diff
@@ -199,14 +210,36 @@ class AIRHelper
...
+                        var x64Path = StringTools.replace(targetPath, project.app.file, project.app.file + "_x64");
+                        FileSystem.rename(workingDirectory + "/" + targetPath, workingDirectory + "/" + x64Path);
+                    }
+
+                    args.remove("-arch");
+                    args.remove("armv8");
+
+                    System.runCommand(workingDirectory, project.defines.get("AIR_SDK") + "/bin/adt", args);
+                }
+            }
```
player-03, line 234: This whole section should use tabs instead of spaces.

src/lime/tools/AIRHelper.hx
```diff
@@ -199,14 +210,36 @@ class AIRHelper
...
+            if(project.config.getString("air.android_package", "apk") == "aab"){		// build aab
+                args.push("-platformsdk");
+                args.push(project.defines.get("ANDROID_SDK"));
+
+                System.runCommand(workingDirectory, project.defines.get("AIR_SDK") + "/bin/adt", args);
+            } else {	// build apk x32 and/or x64
+                if(hasARM64)
+                    System.runCommand(workingDirectory, project.defines.get("AIR_SDK") + "/bin/adt", args);
+
+                if(hasARMV7 || project.architectures.length == 0){
```
player-03, line 223: This is redundant. `hasARMV7` was already defined as `ArrayTools.containsValue(project.architectures, Architecture.ARMV7) || project.architectures.length == 0`.

src/lime/tools/AIRHelper.hx
```diff
@@ -199,14 +210,36 @@ class AIRHelper
...
+            }
 		}
 
 		if (targetPlatform == IOS)
 		{
 			Sys.putEnv("AIR_IOS_SIMULATOR_DEVICE", XCodeHelper.getSimulatorName(project));
-		}
 
-		System.runCommand(workingDirectory, project.defines.get("AIR_SDK") + "/bin/adt", args);
+            System.runCommand(workingDirectory, project.defines.get("AIR_SDK") + "/bin/adt", args);
```
player-03, line 241: This line should use tabs.

src/lime/tools/AIRHelper.hx
```diff
@@ -258,8 +291,8 @@ class AIRHelper
 				System.runCommand("", "killall", ["iPhone Simulator"], true, true);
 			}
 
-			System.runCommand(workingDirectory, project.defines.get("AIR_SDK") + "/bin/adt",
-				["-uninstallApp"].concat(args).concat(["-appid", project.meta.packageName]), true, true);
+//			System.runCommand(workingDirectory, project.defines.get("AIR_SDK") + "/bin/adt",
+//				["-uninstallApp"].concat(args).concat(["-appid", project.meta.packageName]), true, true);
```
player-03, line 295: This will need to be cleaned up. Either delete or create a compiler flag to toggle it.

src/lime/tools/AndroidHelper.hx
```diff
@@ -395,6 +395,7 @@ class AndroidHelper
 
 	public static function uninstall(packageName:String, deviceID:String = null):Void
 	{
+		return;
```
player-03, line 398: You can't just disable an AndroidHelper function because you want Air apps to function differently. Find the call to `uninstall()` and disable that.

src/lime/tools/HXProject.hx
```diff
@@ -26,6 +26,7 @@ class HXProject extends Script
 	public var app:ApplicationData;
 	public var architectures:Array<Architecture>;
 	public var assets:Array<Asset>;
+	public var assetsFolders:Array<String>;
```
player-03, line 29: As I said in 2021, modifying `HXProject` isn't the way to do this. We'll need to revert all `HXProject` changes and try a different approach. [See my original comment for a detailed suggestion](https://github.com/openfl/lime/pull/1475#issuecomment-972369334).

tools/platforms/AIRPlatform.hx
```diff
@@ -143,8 +143,15 @@ class AIRPlatform extends FlashPlatform
 			{
 				if (asset.embed == false && asset.type != TEMPLATE)
 				{
-					files.push(asset.targetPath);
+//					files.push(asset.targetPath);
 				}
+				if (asset.type == MANIFEST)
+					files.push(asset.targetPath);
```
player-03, line 149: Why create a new `if` block instead of changing the original?

player-03 (changes requested):

joshtynjala:
Closing, since I added support for -listen in commit f8ec10b969d82946664a8fe6b719a8b776b574a7

PR #1480, opened 2021-10-06
Added comment/docs for some classes.

pfoof:
Added comments for `lime.ui.FileDialog` that can clear things up on how to use this class.

### DIFF
--- a/src/lime/net/URIParser.hx
+++ b/src/lime/net/URIParser.hx
@@ -29,7 +29,7 @@ class URIParser
 
 	/**
 		Hostname and port along with the credentials found in this URI.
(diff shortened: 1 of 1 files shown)


player-03:
Ok, I didn't realize GitHub would strip consecutive tabs and ruin my indentation, but turns out [it isn't the first time](https://github.com/haxelime/lime/blob/develop/src/lime/app/Application.hx#L221-L226). Perhaps someday I'll go through the whole repo and switch to spaces, but today is not that day.

I also considered improving the behavior of `save()`'s `type` parameter, but this is a documentation PR, so I settled for documenting it.

Issue #1486, opened 2021-10-19
How to retrieve target architecture using defines?


EliteMasterEric:
The following XML tags function as expected:

```
	<echo value="Platform architecture: 32-bit" if="HXCPP_M32" />
	<echo value="Platform architecture: 64-bit" if="HXCPP_M64" />
```

However, there is an issue with this; if the user has omitted these compile flags and simply uses the default `lime build linux` command (which builds for their current architecture), the result is that **neither of these flags will be true**.

Is there a way to print when the user is, say, building for x64, then the user has not explicitly set the `HXCPP_M64` flag?

player-03:
Ok, give my pull request a try, and give [include.hxp](https://lime.software/docs/project-files/hxp-format/) a shot. You're interested in the [`architectures` variable](https://github.com/haxelime/lime/blob/develop/src/lime/tools/HXProject.hx#L27), which will contain all the default architectures for this platform.

You can also check the [`defines` variable](https://github.com/haxelime/hxp/blob/master/src/hxp/Script.hx#L15) for `HXCPP_M32` or `HXCPP_M64` in case the user set those by hand.

MasterEric:
@player-03 I didn't test your PR since it's only for haxelibs, but when I built a `project.hxp` and ran it I got this output:

```
E:\Program Files\HaxeToolkits\424\haxe\lib\hxp/1,2,1/src/hxp/ObjectTools.hx:53: characters 12-29 : Warning : Std.is is deprecated. Use Std.isOfType instead.
E:\Program Files\HaxeToolkits\424\haxe\lib\hxp/1,2,1/src/hxp/ObjectTools.hx:58: characters 12-28 : Warning : Std.is is deprecated. Use Std.isOfType instead.
E:\Program Files\HaxeToolkits\424\haxe\lib\hxp/1,2,1/src/hxp/ObjectTools.hx:78: characters 12-32 : Warning : Std.is is deprecated. Use Std.isOfType instead.
E:\Program Files\HaxeToolkits\424\haxe\lib\hxp/1,2,1/src/hxp/ObjectTools.hx:94: characters 12-29 : Warning : Std.is is deprecated. Use Std.isOfType instead.
E:\Program Files\HaxeToolkits\424\haxe\lib\hxp/1,2,1/src/hxp/ObjectTools.hx:110: characters 12-27 : Warning : Std.is is deprecated. Use Std.isOfType instead.
Called from lime/_internal/backend/native/NativeCFFI.hx line 609
[...]

PR #1494, opened 2021-11-06
Add support for include.hxp.

player-03:
I think I've mentioned before how included projects can't always get the data they need, and it came up again in #1486. It's incredibly easy to get the architectures in an HXP project, but there's just no way in XML, nor is there an elegant way to add support.

There's one issue to be aware of here: all paths in include.hxp will be relative to the main project, which is practically guaranteed to break local paths. So, you have to type out `Sys.getCwd() + "local/path"` each time. Not a deal-breaker, just something people will have to be aware of.

Oh yeah, I refactored a few things while I was at it. IMO, the whole function should have been named `fromDirectory()` in the first place, because otherwise it sounds the same as `fromFile()`. It's a little late to change that now, so I'll settle for making the arguments more explicit.

### DIFF
--- a/src/lime/tools/HXProject.hx
+++ b/src/lime/tools/HXProject.hx
@@ -540,37 +540,48 @@ class HXProject extends Script
 		return HXProject.fromPath(path, userDefines);
 	}
 
-	public static function fromPath(path:String, userDefines:Map<String, Dynamic> = null):HXProject
+	public static function fromPath(directory:String, userDefines:Map<String, Dynamic> = null):HXProject
 	{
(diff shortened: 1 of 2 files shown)


player-03:
I've done only basic testing here, but since this is based on the well-tested `HXProject.fromFile()` function, I don't anticipate major issues. Testing results:

1. include.xml supersedes include.hxp when both exist.
2. After removing include.xml, include.hxp can trace messages.
3. Specifying a relative source path fails; files aren't found.
4. Specifying an absolute source path succeeds; files are found.

Issue #1476, opened 2021-09-21
Deprecated Gradle features were used in this build, making it incompatible with Gradle 6.0.


pozirk:
Hello!

I'm trying to update [extension-admob](https://github.com/HaxeExtension/extension-admob) to work with the latest version of admob sdk 20+ and get the following error:

> Deprecated Gradle features were used in this build, making it incompatible with Gradle 6.0.
> play-services-ads-lite-20.3.0\AndroidManifest.xml:27:5-43:15: AAPT: error: unexpected element &lt;queries&gt; found in &lt;manifest&gt;.

According to this article, the problem is with older version of Gradle plugin:
https://android-developers.googleblog.com/2020/07/preparing-your-build-for-package-visibility-in-android-11.html

As I can see, since Lime 7.7.0, we are using Gradle 5.6.3 and Android Gradle Plugin 3.5.1, but I need a newer version.
Anyway to fix that?
Thank you!

flashultra:
You can set in ```project.xml``` file the following params ( or any other version , check here -> https://developer.android.com/studio/releases/gradle-plugin) :
```xml
	<config:android gradle-version="6.7.1" />
	<config:android gradle-plugin="4.2.0" />
```

pozirk:
Thanks a lot @flashultra!
It seems to work fine now, but I had to fix some AndroidX error by adding the following to **gradle.properties**:

> android.useAndroidX=true
> android.enableJetifier=true

And there are those warnings now:

> WARNING:: NDK was located by using ndk.dir property. This method is deprecated and will be removed in a future release. Please use android.ndkVersion or android.ndkPath in build.gradle to specify the NDK to use. https://developer.android.com/r/studio-ui/ndk-dir
> NDK was located by using ndk.dir property. This method is deprecated and will be removed in a future release. Please use android.ndkVersion or android.ndkPath in build.gradle to specify the NDK to use. https://developer.android.com/r/studio-ui/ndk-dir

So, hope there will be official release with updated Gradle after all.  :smile:

player-03:
> ```groovy
> android.useAndroidX=true
> android.enableJetifier=true
> ```

Yep, these are a good idea nowadays. We'll just want to check `::if (ANDROID_TARGET_SDK_VERSION>=28)::`.

> NDK was located by using ndk.dir property. This method is deprecated and will be removed in a future release. Please use android.ndkVersion or android.ndkPath in build.gradle to specify the NDK to use.

[...]

PR #1500, opened 2021-11-10
Fix warnings and errors in the Android build process

player-03:
This addresses the (reproducible) issues brought up in #1476, plus some that came up when testing the changes.


player-03:
Apparently this breaks PiratePig. Something to do with uploading textures, which is odd because none of these changes should have affected much of anything.

I was hoping to get this merged in time for the next release, but now that's off the table. This is apparently not the self-contained change I thought it was, and will require extra testing.

player-03:
Ok, so it turns out I messed up the [tagged pointer](https://source.android.com/docs/security/test/tagged-pointers) workaround. PiratePig works fine when I include it. That said, I still don't plan to merge just yet, in case someone else finds something.

MAJigsaw77:
Why this doesn't get pushed?

Issue #1462, opened 2021-05-06
SDL out of date


rvh1986:
Getting the following error trying to compile to Android:

`> Task :app:compileDebugJavaWithJavac
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: D:\Drive\DEV\IDEA\SquareSpinner\Export\android\bin\app\src\main\java\org\libsdl\app\SDL.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.`

It looks like this is solved in the latest version of SDL: https://github.com/libsdl-org/SDL/blob/main/android-project/app/src/main/java/org/libsdl/app/SDL.java

I can't find a way to update the SDL.java file in my own project as it gets rebuild every time I compile.

Haxe 4.2.1
Lime 7.9.0
OpenFL 9.1.0

player-03:
Late response, but Lime contains [a template folder for SDL](https://github.com/haxelime/lime/tree/develop/templates/android/template/app/src/main/java/org/libsdl/app). If you overwrite Lime's copy, it'll keep the changes.

PR #1501, opened 2021-11-13
Update to SDL 2.0.16

player-03:
Between #1462 and a couple errors I found on my own, I decided it's time for an update. Days of debugging later, it all seems to work again.

See also native-toolkit/sdl#16.

Notes:

- In addition to the CI tests, I've tested both DisplayingABitmap and PiratePig on both Linux and Android.
- We're now targeting iOS 13 by default, because SDL uses the Core Haptics framework.
- Since writing the commit, I've tested the SDL_RWops thing on Linux. (Turns out only `PNG.decodeFile()` and `JPEG.decodeFile()` use it.)
- I considered trying to [build libhidapi.so](https://github.com/native-toolkit/sdl/blob/master/src/hidapi/README.txt) for Android, but decided to leave it for later. It's probably not _that_ hard, just have to figure out how best to point the NDK at [Android.mk](https://github.com/native-toolkit/sdl/blob/master/src/hidapi/android/jni/Android.mk). I noticed [this folder](https://github.com/native-toolkit/sdl/tree/master/src/hidapi/android) has a project.properties file - maybe it can be included as a dependency project?

### DIFF
--- a/project/src/backend/sdl/SDLSystem.cpp
+++ b/project/src/backend/sdl/SDLSystem.cpp
@@ -31,6 +31,10 @@
 #endif
 #endif
 
+#ifdef ANDROID
+#include <android/asset_manager_jni.h>
+#endif
+
 #include <SDL.h>
 #include <string>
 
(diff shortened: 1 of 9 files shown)


inovento:
Hello, upon building for android we're getting error in `src/backend/sdl/SDLSystem.cpp`
`- src/backend/sdl/SDLSystem.cpp
Error: ./src/backend/sdl/SDLSystem.cpp:588:86: error: no member named 'asset' in 'SDL_RWops::(anonymous struct at lib/sdl/include/SDL_rwops.h:97:9)'
                                fd = AAsset_openFileDescriptor ((AAsset*)(((SDL_RWops*)handle)->hidden.androidio.asset), &outStart, &outLength);
                                                                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^`

The compile checks for the merge request are reported as successful but the logs are no longer available and can't confirm there. Any idea what might be causing the error? Are we missing part of the code that is not in the merge request?

player-03:
Closing because #1531 does it better.

player-03:
@inovento My best guess is that 17f516584d6a312f470d025508689bf2caab9ddd failed to take. In other words, if you ran `git rev-parse HEAD` inside the SDL submodule, it would have said `be23e5934706347af00f4e4caf5e0a9bcdc369ca` instead of `dd6b884fce6cb0efca489c25d5f0fd714b14c549`. If so, you'd want to try `git submodule update` from outside the submodule, or `git checkout dd6b884fce6cb0efca489c25d5f0fd714b14c549` from inside.

Issue #1502, opened 2021-11-17
`BackgroundWorker` edge cases


player-03:
As it stands, [`BackgroundWorker`](https://github.com/haxelime/lime/blob/1188b3aa42b18992ef50241050217304b28c1ae5/src/lime/system/BackgroundWorker.hx) has several problems.

How it's designed to work
===================
The first problem is the lack of documentation, so let's do a quick run-through of a typical use case. For this example, suppose we have `function generateItems(type:ItemType, count:Int):Array<Item>`. It's gotten slower as the game got bigger, so we want to make it asynchronous.

```haxe
private var bgWorker:BackgroundWorker;

//Background functions must take one argument and return nothing.
private static function generateItems(params:{ type:ItemType, count:Int }):Void {
    if(params.count < 1)
        throw "Must always generate at least one item!";
    
    var items:Array<Item> = [];
    
    //Glossing over how they're actually made. Just assume it takes time.
    for(i in 0...count) {
        /* ... */
    }
    
    //Instead of returning the array, dispatch an event.
    bgWorker.sendComplete(items);
}

public function new() {
    //First things first.
    bgWorker = new BackgroundWorker();
    
    //To run a function on the other thread, add it as a `doWork` listener.
    bgWorker.doWork.add(generateItems);
    
    //Prepare the argument.
    var itemTypeAndCount = { type: ItemType.SHIELD, count: 30 };
    
    //To receive the result, add a completion listener.
    bgWorker.onComplete(function(shields:Array<Item>) {
        list = shields;
    });
    
    //Add an error listener too, just in case.
    bgWorker.onError.add(handleError);
    
    //Create the new thread and dispatch `doWork` on that thread, passing
    //the given argument to each listener.
    bgWorker.run(itemTypeAndCount);
}
```

In cases like these, `BackgroundWorker` works well.

Edge case: "completed" race conditions
==============================
Let's start with something minor. The `sendComplete()` function is run on the background thread, and sets `completed = true`. `completed` is typically checked on the main thread, which can in rare cases lead to crashes and other bugs.

```haxe
private function onEnterFrame(e:Event) {
    if(shields.bgWorker.completed)
        equipShield();
    else
        hideBehindCover();
}

private function equipShield() {
    if(!shields.bgWorker.completed)
        throw "bgWorker hasn't finished loading shields yet!";
    
    if(activeShield == null) {
        activeShield = shields.list[Std.int(Math.random() * shields.length)];
        
[...]

Issue #1081, opened 2017-10-02
Add asynchronous worker support for the HTML5 target


jgranick:
`BackgroundWorker` as well as `ThreadPool` use a synchronous mode on HTML5, but these could be made to use web workers on HTML5

https://stackoverflow.com/a/37156020

Depending on the performance of web workers, it may be preferable to consider whether one or both of the above APIs should also have a way to use them synchronously by choice, if web workers end up being too expensive

player-03:
`ThreadPool` won't work in HTML5, because [you can't pass functions as a message](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm). This behavior is required by [Future](https://github.com/haxelime/lime/blob/8c9808b038ea06ea1f8c8c1ffb4a6d01d43c7972/src/lime/app/Future.hx#L82) at the very least.

(Yeah, I know only Firefox uses the `structuredClone()` function, but the error also happens in Chrome, so whatever function it uses imposes the same restriction.)

Maybe we can pass a string and then call `Function()`...

barisyild:
HTML5 workers does not support shared memory usage.
I know that it is possible to send byte data only with SharedArrayBuffer.

Note: wasm has thread support but unfortunately does not share memory like worker.

player-03:
Oh yeah, forgot to mention that, but the lack of shared memory is the other problem with `ThreadPool`. The current implementation relies on a shared `Deque` - the main thread adds a task to this queue, and whichever thread becomes available next gets to handle it.

Workaround is obvious but annoying: the main thread will have to handle all scheduling. Keep track of which threads are busy and which are done, and assign incoming tasks to available threads.

player-03:
After significant work, [I've concluded that web workers are incompatible with Lime's mission](https://player03.com/development/on-starting-over/). (Long post, so [click here if you just care about the reasoning](https://player03.com/development/on-starting-over/#web-worker-assumption).)

As an alternative, I propose a single-threaded async mode. Details to come.

player-03:
This didn't get closed automatically? Whoops.

PR #1503, opened 2021-11-17
Add HTML5 support to `BackgroundWorker` and `ThreadPool`

player-03:
This request fulfills #1081, addresses the issues in #1502, and adds documentation.

It will also break existing code, because `js.html.Worker` runs its functions in complete isolation. Many functions that work fine with `sys.thread.Thread` will fail because they can no longer access outside variables and functions. This will break someone's code if (and only if) they relied on the "fall back to synchronous execution" behavior. Fortunately, they can restore the old behavior using `<haxedef name="force_synchronous" if="js" />`.

### DIFF
--- a/docs/hxml/html5.hxml
+++ b/docs/hxml/html5.hxml
@@ -4,6 +4,7 @@
 -D html5
 -D doc_gen
 -D lime-doc-gen
+-D force_synchronous
 ImportAll
 -lib lime
 --no-output
--- a/include.xml
+++ b/include.xml
@@ -59,6 +59,7 @@
 
 	<haxedef name="force_synchronous" if="emscripten || display" />
 	<haxedef name="force_synchronous" if="html5" unless="lime-web-workers" />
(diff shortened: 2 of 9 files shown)


player-03:
This will two browser features:
 
- [Web workers](https://caniuse.com/webworkers), which are almost universally available.
- [The URL API](https://caniuse.com/url), which isn't supported by IE, which I've decided to ignore. There are workarounds for IE, but they aren't worth it.

player-03:
Turns out I didn't abandon this! I was just waiting to push until I finished all the features.

player-03:
I wrote a blog post explaining this: https://player03.com/openfl/web-workers/

player-03:
As discussed [here](https://player03.com/development/on-starting-over/#web-worker-assumption), I've concluded that this is not the correct way to approach this problem. I'll be submitting a new - and far simpler - pull request soonish.

Issue #1511, opened 2021-12-29
Audio length problems by changing the pitch


Cheemsandfriends:
Hello,
There's a problem with lime's way to finish it's length. To be specific, when you use the normal form to change a pitch, the audio will pitch, yeah. The problem is that the duration "remains the same" by default. Making that if you have a high pitch of the audio, the audio will have to wait a few seconds to trigger the onComplete or if you use a low pitch, the "length" will cut of the audio.

PR #1512, opened 2021-12-29
Make a Consistent length when you're changing the pitch

Cheemsandfriends:
So I made some rough code with the Native Audio Backend and made the looping go smooth whatever is the pitch on. This was tested on Hashlink so I guess that it should work on different platforms... Right?

(This Closes Issue Number #1511 And this comes with the Pull Request Number #1510 with the Review notes included)

### DIFF
--- a/src/lime/_internal/backend/flash/FlashAudioSource.hx
+++ b/src/lime/_internal/backend/flash/FlashAudioSource.hx
@@ -116,6 +116,15 @@ class FlashAudioSource
 	{
 		return loops = value;
 	}
+	public function getPitch():Float
+	{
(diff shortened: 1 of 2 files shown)


player-03:
I copied these changes over to #1510, with adjustments. I don't think there's a need to run the check every frame, but there definitely _is_ a need to update `setLength()` and `setCurrentTime()`, or it'll reset each loop.

Cheemsandfriends:
I made it on each frame because it was like the best to check on when the pitch was changed.
The thing that you made on the PR is good, but I think it's only made once. Should test it out tho

PR #1514, opened 2022-01-08
Add a `if (__backend != null)` statement

Cheemsandfriends:
Hello,
So while trying to add a pitch setting more nicely in the PR number #1510, It seems that when you get and set the pitch, it crashes. So basically meant that the other ones, like the loop, gain etc. won't work because the pitch didn't work without it.

This adds the `if (__backend != null)` in case you play the music at the beginning of everything and you want to set some things while it loads and stuff

### DIFF
--- a/src/lime/media/AudioSource.hx
+++ b/src/lime/media/AudioSource.hx
@@ -42,82 +42,137 @@ class AudioSource
 		}
 	}
 
+	
 	public function dispose():Void
 	{
-		__backend.dispose();
+		if (__backend != null)
+			__backend.dispose();
 	}
 
 	@:noCompletion private function init():Void
 	{
-		__backend.init();
+		if (__backend != null)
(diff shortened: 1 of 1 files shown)


Cheemsandfriends:
Srry if this is a shitty solution.
I actually don't know why the backend brings null and the game crashes if you set something so I made the easiest solution imo. even tho it isn't optimal

Cheemsandfriends:
I closed it cos of the comment of @player-03 lol

player-03:
Ah, so it really does crash. (I thought it was just hypothetical.) Anyway, as I said in that comment, we should figure out why it's null, not just hide the error.

Can you post your test code?

Cheemsandfriends:
@player-03, Nvm, I think that I know why it crashed before. I'll post A thingy with a review post and why I think it crashes

Issue #1516, opened 2022-01-22
[HL] Invalid function signature for `lime_key_event_manager_register`


player-03:
4595541 produces this error on HashLink on Linux:

```text
FATAL ERROR : Invalid signature for function lime@lime_key_event_manager_register : PP_vOdiii__v required but PP_vOiiii__v found in hdll
```

The difference seems to be the third argument, which is indeed the one that was changed.

I tried `lime rebuild tools -clean`, but that didn't work. The solution probably requires changing a C++ function. (Or conditional compilation as a last resort.)

Shouldn't be hard, but I don't have time to track it down today, so I'm making this issue to help me remember it later.

ShaharMS:
I made a PR that fixed that error, and it was merged in a some weeks ago. The error was produced because the keyCode argument in the event manager was changed from `Int` to `Float` to fix some neko issues. I used some very minimal one-liner conditional to fix that:
```haxe
var keyCode: #if neko Float #else Int
```

PR #1517, opened 2022-01-24
Update hashlink.

Apprentice-Alchemist:
Update hashlink to latest, and add all the hdll's required for the haxe std.
Also change the names of exported functions in lime.hdll, in the interest of making future HL/C support easier.
Also fix an issue introduced by https://github.com/haxelime/lime/commit/45955412eb685ee2f760d54850181e4127b57030.

Tested on Linux.
Note that this does not update that HL_PATH thing in the windows tooling, someone else will have to make that copy *everything* instead of only `ssl.hdll`.


project/include/hl.h
```diff
@@ -786,7 +799,7 @@ typedef struct {
 	int length;
 } vstring;
 
-// #define DEFINE_PRIM(t,name,args)						DEFINE_PRIM_WITH_NAME(t,name,args,name)
+// the is macro has been renamed from DEFINE_PRIM to DEFINE_HL_PRIM, because DEFINE_PRIM is already used
```
player-03, line 802: "the is"

player-03:
> Also change the names of exported functions in lime.hdll, in the interest of making future HL/C support easier.

I'm unsure how this makes things easier. Are apps going to have to load from both lime.hdll and lime.ndll, causing name conflicts? Are you saying it's clearer to have unique names?

---

The fact that this is all one commit makes it hard to follow. If it was me, I'd break it up into three commits:

1. Fix #1516. (That's just line 3950 of ExternalInterface.cpp plus line 24 of KeyEvent.h, right?)
2. Update HashLink. (That's just hl.h, as far as I can tell.)
3. Rename the functions. (Everything else.)

Even with simple commit messages, this will make it much easier to see why each change was made. (Though for the third commit I'd recommend a sentence or two of explanation.)

Apprentice-Alchemist:
> Fix # 1516. (That's just line 3950 of ExternalInterface.cpp plus line 24 of KeyEvent.h, right?)

Correct
> Update HashLink. (That's just hl.h, as far as I can tell.)

`hl.h` + the binaries

> I'm unsure how this makes things easier. Are apps going to have to load from both lime.hdll and lime.ndll, causing name conflicts? Are you saying it's clearer to have unique names?

It makes it easier by doing what the haxe compiler expects us to do.
When building to HL/C, apps would have to be linked with `lime.hdll`.
At the moment the functions exported in `lime.hdll` (those marked with `HL_PRIM`)  are named `hl_lime_foo`, whereas the C code generated by the haxe compiler would expect functions with named `lime_lime_foo`.

Now the functions are named `lime_hl_foo`, and the haxe code has been changed accordingly.

Apprentice-Alchemist:
I've split the commit up into three, as you suggested.

player-03:
> I've split the commit up into three, as you suggested.

Perfect!

> `hl.h` + the binaries

Ah, of course.

> At the moment the functions exported in `lime.hdll` (those marked with `HL_PRIM`) are named `hl_lime_foo`, whereas the C code generated by the haxe compiler would expect functions with named `lime_lime_foo`.

So you're saying [HashLink/C](https://haxe.org/manual/target-hl-c-compilation.html) was broken, and this was never detected because we'd only tested HashLink bytecode. (It wouldn't be the first time something like that happened...)

Just to make sure I can understand, wouldn't it have required fewer changes to switch to `lime_lime_foo`? You would still have needed to update all the .cpp files, but it wouldn't have required so many changes on the Haxe side of things. If you went to the extra effort, you must prefer `lime_hl_foo`. Is this just for clarity?

Apprentice-Alchemist:
> So you're saying HashLink/C was broken, and this was never detected because we'd only tested HashLink bytecode.

Yes.
> Just to make sure I can understand, wouldn't it have required fewer changes to switch to lime_lime_foo? You would still have needed to update all the .cpp files, but it wouldn't have required so many changes on the Haxe side of things. If you went to the extra effort, you must prefer lime_hl_foo. Is this just for clarity?

Personally I think `lime_hl_foo` is clearer than `lime_lime_foo`.
And the only changes on the Haxe side are to the `@:hlNative` metadata's.
The function names on the haxe side are still exactly the same as before.

Apprentice-Alchemist:
I cannot reproduce the Windows build failure on my machine, so I don't know what's wrong.

As for the Android failure, you're using API level 16, whereas `uchar.h` was added in API level 21.
(API level 16 is Android 4.1, which was released in 2012)
(API level 21 is Android 5.0, which was released in 2014)
According to a quick search, Android versions higher than 5.0 make up more than 96% of the Android market, so I believe it would be okay to use API level 21.

Alternatively I can try patching Hashlink, but I would really like to avoid that.

player-03:
> As for the Android failure, Lime seems to be using API level 16, whereas `uchar.h` was added in API level 21.

Huh, well that led me down a rabbit hole.

Lime targets API level 28 but loads header files from the `android-16` folder. According to 6dce7ad, the folder's number has to match the minimum SDK version, which is why that's also 16.

Lime does already use `android-21` for 64-bit builds, because previous ones didn't support it. Everything else uses `android-16` to maximize backwards compatibility. Interestingly, the NDK we're using goes as high as `android-26`, but since Lime is all about backwards compatibility, that isn't likely to happen.

If you want to update the minimum SDK version - and I agree you have a good case for it - please submit it as a separate pull request, so we can discuss it separately.

inovento:
We can repro the build error on Windows namely the halt at JPEG.cpp. How can we help to track the root cause down?

player-03:
Does it still fail if you combine this with #1519?

inovento:
That’s the only way we tried and it fails. Haven’t tried without that commit.

player-03:
I'm surprised it's the same error. I don't know what's up with that, but I'll take a closer look later.

Apprentice-Alchemist:
I can also reproduce the Windows error now, I think I forgot to checkout my branch when I tested it last time.

The issue is the same as here: <https://github.com/native-toolkit/jpeg/pull/2>.

Apprentice-Alchemist:
Or to merge the pull request I linked (or even to move to libjpeg-turbo).

Personally I would prefer to keep our bundled hashlink headers as close as possible to upstream.
The less patches, the easier to update.

(Ideally we could even add upstream hashlink as a git submodule and make `lime rebuild hl` rebuild that too, making it a lot easier to update hashlink, it would just be a question of doing `git -C path/to/hl pull origin master`.)

Apprentice-Alchemist:
The only modification lime makes at the moment is renaming the `DEFINE_PRIM` macro to `DEFINE_HL_PRIM` (or something like that).
There are two solutions for that:
1. rename the other DEFINE_PRIM
2. add a bunch of `#ifdef` so that hashlink and neko/hxcpp things are not included at the same time.

Apprentice-Alchemist:
So, hashlink is now a submodule.
In theory CI should pass now (at least for Linux and macOS).
Windows requires https://github.com/native-toolkit/jpeg/pull/2.

Apprentice-Alchemist:
And of course I forget to install the bloody packages.
And macOS is being difficult, as usual.
Oh well, I'll deal with that tommorow.

Apprentice-Alchemist:
I think the iOS failure is because `char16_t` and `char32_t` are defined in C++ (but not C, at least on iOS).
`hl.h` tries to define those types, since it's included in a C++ files, we get an error.

Apprentice-Alchemist:
Looks like everything is working now.

Apprentice-Alchemist:
I've updated the PR to use the latest version of hashlink (1.12, which is now released).

ShaharMS:
Is there any reason why this isn't merged?

player-03:
It seem to work on my machine. At least, I can rebuild lime.hdll and test DisplayingABitmap.

Looking at BuildHashlink.xml (and `LinuxPlatform`), it looks like lime.hdll doesn't incorporate any of the other libraries (Cairo, SDL, etc.). And that would explain why you compile lime.ndll at the same time: Hashlink apps require both. Is that right?

Yet somehow lime.hdll is slightly larger than lime.ndll... Maybe that's just because Hashlink includes its own copies of several of the same libraries (inconveniently spread across [two](https://github.com/HaxeFoundation/hashlink/tree/68daa125ad115a1f1d1db65d38e6045858d55013/include) [folders](https://github.com/HaxeFoundation/hashlink/tree/68daa125ad115a1f1d1db65d38e6045858d55013/libs)).

Apprentice-Alchemist:
`BuildHashlink.xml` builds `hl(.exe)`, `libhl.so/dylib/exe`, `ssl.hdll`, `fmt.hdll`, `mysql.hdll`, `ui.hdll` and `uv.hdll` (these hdlls are needed for functionality of the Haxe std, eg `sys.ssl` depends on `ssl.hdll`).
On Windows these indeed use hashlink's copies of the libraries, on Linux/macOS they use system packages.

`lime.hdll` is still built by `Build.xml` and includes all of lime's dependencies (SDL, Harfbuzz, etc).

Hashlink apps do not depend on `lime.ndll`, only on `lime.hdll` (and the hashlink runtime of course)

> (inconveniently spread across [two](https://github.com/HaxeFoundation/hashlink/tree/68daa125ad115a1f1d1db65d38e6045858d55013/include) [folders](https://github.com/HaxeFoundation/hashlink/tree/68daa125ad115a1f1d1db65d38e6045858d55013/libs)).

`/include` contains the copies of the libraries, `/libs` contains the wrapper code.

player-03:
Looking back through open issues, I came across #1524, which still needs resolution. Is fmt.hdll one of the binaries you added?

Apprentice-Alchemist:
Yes, all the binaries needed for things in the Haxe standard library are added.
(See previous message.)

player-03:
Oh hey, you did mention it. Whoops.

ShaharMS:
lets gooo this got merged!

joshtynjala:
If anyone has issues rebuilding after this pull request was merged, don't forget to update submodules!

I got the following error on macOS:

> In file included from ./src/ui/TextEvent.cpp:1:
> include/system/CFFI.h:6:10: fatal error: 'hl.h' file not found
> #include <hl.h>
>          ^~~~~~
> 1 error generated.

But it was an easy fix:

```sh
git submodule update --recursive
```

joshtynjala:
Getting the following error when I try `lime test hl` on macOS:

> dyld[29362]: Library not loaded: @executable_path/libhl.dylib
>  Referenced from: /Users/joshtynjala/Development/feathersui/feathersui-openfl/samples/components-explorer/bin/hl/bin/Components.app/Contents/MacOS/Components
>  Reason: tried: '/Users/joshtynjala/Development/feathersui/feathersui-openfl/samples/components-explorer/bin/hl/bin/Components.app/Contents/MacOS/libhl.dylib' (no such file), '/usr/local/lib/libhl.dylib' (no such file), '/usr/lib/libhl.dylib' (no such file)

It appears to be looking for libhl.dylib in the wrong location. When I "Show Package Contents" in the _.app_ file, it is actually located here:

/Users/joshtynjala/Development/feathersui/feathersui-openfl/samples/components-explorer/bin/hl/bin/Components.app/libhl.dylib

Apprentice-Alchemist:
Hmm... I'll have to look again at were the tooling copies things.
I'd have expected it to copy the libhl library to the same place were it puts the executable (although that's not the right place for libraries iirc).

player-03:
> If anyone has issues rebuilding after this pull request was merged, don't forget to update submodules!

That's going to come up a bunch in the near future.

> tried: '/Users/joshtynjala/Development/feathersui/feathersui-openfl/samples/components-explorer/bin/hl/bin/Components.app/Contents/MacOS/libhl.dylib' (no such file)

A search for "Contents/MacOS" turns up MacPlatform.hx and a bunch of SDL files, meaning the problem lies in `MacPlatform`. Some code that's using [`executableDirectory`](https://github.com/openfl/lime/blob/ad70700a14ae7fd70630e263604afbf9bf4cbe90/tools/platforms/MacPlatform.hx#L163) should instead use `applicationDirectory`, or vice versa. Maybe [this line](https://github.com/openfl/lime/blob/ad70700a14ae7fd70630e263604afbf9bf4cbe90/tools/platforms/MacPlatform.hx#L209) is copying the files into the app root when it should be copying them into `executableDirectory`?

joshtynjala:
> I'd have expected it to copy the libhl library to the same place were it puts the executable (although that's not the right place for libraries iirc).

Just in case it helps, I should point out that I saw that most of the _.hdll_ files are in the root of the _.app_ directory with the _libhl.dylib_ too. _lime.hdll_ in in _Contents/MacOS_, though.

player-03:
[Found the line responsible for libhl.dylib in particular](https://github.com/openfl/lime/blob/ad70700a14ae7fd70630e263604afbf9bf4cbe90/src/lime/tools/HashlinkHelper.hx#L52).

```haxe
System.copyFile(Path.combine(hlPath, "libhl.dylib"), Path.combine(applicationDirectory, "libhl.dylib"));
```

And [here's what that line was replacing](https://github.com/openfl/lime/commit/cb3b0240de91f0102d400b375fca3e6c59d546d5#diff-73e3006270e360d32e981eff483614cc5c589f1f83b969bf8754dbb24fec618eL223-R224).

player-03:
@joshtynjala Since I'm not set up for Mac testing, can you be the one to change [line 209](https://github.com/openfl/lime/blob/ad70700a14ae7fd70630e263604afbf9bf4cbe90/tools/platforms/MacPlatform.hx#L209) and push the commit?

joshtynjala:
Changing from `applicationDirectory` to `executableDirectory` on Line 209 seems to be the correct solution there. The _.dylib_ and _.hdll_ files are now placed in _Contents/MacOS_, and `lime test hl` is working for me.

PR #1518, opened 2022-02-01
Add virtual threads and web workers.

### LINKED ISSUE #1081: Add asynchronous worker support for the HTML5 target

### LINKED ISSUE #1502: `BackgroundWorker` edge cases

### LINKED ISSUE #1580: (ThreadPool) Enable multithreading support for HashLink target?
Presumably you just need to change `#if (cpp || neko)` to `#if (cpp || neko || hl)`.

player-03:
This pull request adds [virtual threads](https://player03.com/openfl/threads-guide/#virtual-threads) (thread-like code that runs entirely on the main thread), adds web workers (HTML5 threads), merges `BackgroundWorker` into `ThreadPool`, and helps programmers reduce their risk of race conditions, all while maintaining backwards compatibility. I've written [a detailed tutorial](https://player03.com/openfl/threads-guide/) to go with the new code, including embedded HTML5 examples.

This request supersedes #1503, resolves #1081, and resolves #1502.


player-03:
It would be nice to support web workers (as proposed in #1081), but it proved impossible to add them without breaking backwards compatibility. I tried making them opt-in, but even then they required the user to jump through so many hurdles that I concluded it wasn't worth it. Hurdles were just piling up.

- If you want to pass a function (like `FutureWork` does), you need to convert it to `ThreadFunction` first. It's possible to do this incorrectly, making the function unusable, resulting in a 400+ character error with instructions to fix it. (And that's the short version!)
- If you want to be able to access a class, you have to add it to the worker's header. I made a convenient function to copy the whole thing, so at least that wasn't hard. Then you'd have to add its parent classes, and any classes these classes referenced, and repeat.
- When you pass a message, by default it makes a copy. If you're passing a large object and don't want to go through the slow process of copying it, you need to pass it as part of a second argument to `sendComplete()`. (But it also has to be in the first argument. It's a bit confusing.)
- If you pass a class instance to or from a web worker, it will lose all its instance methods. You can fix this by calling `restoreInstanceMethods()`... but only partially. It will only restore the instance methods known at compile time. For instance, if you pass a `Sprite` but the compiler only knows it's a `DisplayObject`, it will become a `DisplayObject`, without any of the overridden methods.

It was at about that point I gave up on adding web workers to Lime. They just don't belong here. They belong in a separate library that can properly document them. (And I do plan to make that library.)

player-03:
Correction: web workers are in fact viable! Thanks to Jonas Nyström and the others who wrote [this article](https://code.haxe.org/category/javascript/javascript-inline-workers.html) for the suggestion to reuse the existing script file.

I chose to sacrifice performance in places to provide a seamless user experience, but it works relatively well considering the number of restrictions imposed by web workers.

The fact that users are forced to use static functions actually makes it easier on them, as they don't have to deal with the confusion around instance functions. (That's assuming they opt in to use web workers. If not, they can use whatever functions they like.)

I've also encountered a strange issue with transferable objects. Namely, they don't seem to improve performance at all. Sending a 6 MB buffer seems roughly as long whether or not you mark it as transferable. It becomes inaccessible as expected if marked, but that seems to be the only difference.

player-03:
Finally. After... how long has it been now?

Finally, after _four months_ (!!) of work, I have uncovered the truth about `BackgroundWorker` and `ThreadPool`.

The truth is, the differences between these classes are entirely artificial. I tried for so long to enforce those differences, to give each one its own use case, but in the end I failed.

I tried to restrict `BackgroundWorker` to a single thread. However, I needed to account for the possibility that the user would call `run()` multiple times, something the old version overlooked. Throwing an error seemed inappropriate, as did exiting silently, so the only viable option was to cancel the old thread and start a new one. And guess what? That meant `BackgroundWorker` was using more than one thread. Perhaps they weren't so different after all.

The two classes became more and more similar over the ensuing months, until only a single meaningful difference remained. That difference was that `ThreadPool` went out of its way to force you to use the same function for every job. There was no performance benefit to enforcing this restriction, nor did it make the code simpler. All it did was preserve a distinction between classes.

And this brings us to the present moment:

The truth about BackgroundWorker

(Before anyone asks, yes, they're both still backwards-compatible.)

ShaharMS:
Any updates on this?

player-03:
I think this should go in 8.1.0, but since there will be a couple more releases before then, we just have to wait.

Though @Dimensionscape mentioned at one point that he wanted to make different branches for different upcoming releases. Perhaps we could make a v8.1.0 branch and merge into that?

player-03:
To anyone currently using my branch: please switch to 8.2.0-Dev at your earliest convenience. I'll leave my branch online for the time being, but not forever. Also, any fixes will now go in 8.2.0-Dev.

joshtynjala:
Hmm... OpenFL 9.2 is not currently compatible with Lime 8.2.0-Dev.

> Error: { add : (Dynamic -> Void) -> Void } has no field cancel

https://github.com/openfl/openfl/blob/9.2.1/src/openfl/filesystem/FileStream.hx#L300

player-03:
You can safely remove that line.

In fact, as long as `__disposeFileStreamWorker()` is called from the main thread, you can remove `onProgress.cancel()` and `onComplete.cancel()` too. (If `__disposeFileStreamWorker()` is called from anywhere other than the main thread, that's a potential problem and should change ASAP.)

joshtynjala:
> You can safely remove that line.

Are you saying that you intend for this to be a breaking change? OpenFL 9.2 is already released. We can't remove it from that version.

joshtynjala:
To preserve backwards compatibility, you could add a `cancel` field that is simply an empty function.

player-03:
In the original `BackgroundWorker` class, the intent was for you to add exactly one `doWork` listener (see the "multiple listeners" section in #1502). `Event.cancel()` only affects listeners after the currently-active one, meaning `doWork.cancel()` doesn't do anything.

That's what I mean by you can safely remove it. It never had an effect, even with the old version of `BackgroundWorker`.

joshtynjala:
I think you are missing what I am trying to tell you.

1) This change has resulted in a compiler error if you try to build an app with OpenFL 9.2 and Lime 8.2.0-Dev
2) It is impossible for me to remove a line from OpenFL 9.2 because it has already been submitted to Haxelib

It doesn't matter that it had no effect. That's fine, and we can safely remove the line in OpenFL 9.3 (or 9.2.2).  However, the call was there in OpenFL 9.2.0 and 9.2.1, so either this PR needs to go into 9.0.0-Dev instead, or you need to polyfill the `cancel` method to remain compatible with existing code that cannot be changed.

Dimensionscape:
Hm I thought @player-03 and I spoke about this already. So BackgroundWorker typedefs to ThreadPool, but we need to bring back the BackgroundWorker implementation, include a depreciation warning for BackgroundWorker and then phase it out in lieu of ThreadPool.

Dimensionscape:
@joshtynjala 
https://github.com/openfl/lime/commit/4ffc4422b2165249de994f2e82f359bd454b6fec should resolve it for now.

player-03:
Ok, see #1643.

> 1. This change has resulted in a compiler error if you try to build an app with OpenFL 9.2 and Lime 8.2.0-Dev

And specifically use the `FileStream` class. Which is probably why this wasn't noticed sooner.

> Hm I thought @player-03 and I spoke about this already.

Not this specific error, no. Last I heard you were going to do some more debugging to figure out where your error actually came from.

> either this PR needs to go into 9.0.0-Dev instead

I mean, I don't see why. There are still multiple revisions necessary for 

> we need to bring back the BackgroundWorker implementation

That shouldn't be necessary, no. Not over a single missing function.

dimensionscape:
Eh, I've already reverted it just in case there are any other oversights. 

It doesn't make sense to have both Classes. I've added the deprecation warning so that we can get rid of it all together in a future update.

dimensionscape:
If you really want to typedef it, go ahead and change it back, but include the deprecation warning and make sure there are no more backwards compatibility issues.

player-03:
Yeah, that was my plan.

Dimensionscape:
Hmm, I'm having issues with a multithreaded network application since switching to 8.2.0. Client is crashing intermittently on cpp without any error and I suspect the thread changes since it works fine with the old BackgroundWorker impl...

Will try to narrow down, reproduce and see why.

PR #1527, opened 2022-04-28
prevent clipboard being set as null

junsred:


player-03:
IMO, that order of operations isn't intuitive. I had to read the code twice to figure out why it was there. Perhaps one of these?

A.

```haxe
var text = Clipboard.text;
if (text == null)
{
	text = "";
}
event.clipboardData.setData("text/plain", text);
```

B.

```haxe
event.clipboardData.setData("text/plain", Clipboard.text != null ? Clipboard.text : "");
```

joshtynjala:
I agree that flipping the logic around makes it more clear. Plus, it potentially avoids the allocation of an empty string on platforms where that might matter.

Personally, I like option A. If that change is made, I think it looks good and can be merged.

junsred:
@player-03 @joshtynjala thanks for the comments, I've updated it accordingly.

PR #1528, opened 2022-04-28
focus back to textInput after setting clipBoard

junsred:


junsred:
@jgranick please check this pr

junsred:
> What is the importance of the Timer here?
@Dimensionscape 

without timer it doesn't copy + timer makes sure game doesn't stop when player does spam paste

junsred:
@player-03 how about this commit?

junsred:
Without this code, copying selected text in textInput causes textInput to lose focus and client loses ability to input.

player-03:
I wonder if the function name could be shorter.

- `refocusTextInput`
- `restoreFocusAfterDelay`
- `restoreInputFocus`

But there's always a trade-off. Longer names are explicit, shorter names are easier to remember. Personally, I err on the side of shorter names, adding a comment if I need to explain further.

For example:

```haxe
var __focusPending = false;

/**
	Restores focus to `textInput`, making sure not to change it too many times in succession. Changing focus multiple times
	per frame can sometimes interfere with copying or pasting text, as browsers may ignore all but the final focus target.
**/
public function focusTextInput():Void
{
	if (__focusPending) return;
	__focusPending = true;

	Timer.delay(function()
	{
		__focusPending = false;
		if (textInputEnabled) textInput.focus();
	}, 20);
}
```

The comment doesn't have to mention the delay, it just has to explain why the delay exists. (Same with `__focusPending`.) A good comment explains the problem being solved, so if someone finds a better solution, the explanation still applies.

junsred:
@joshtynjala could you check this pr? Right now client is unable to input after copying text. This pr solves the problem.

player-03 (approved):

PR #1529, opened 2022-05-04
Fix `Main.cpp` template for `-static` `-debug` builds

arm32x:
When building a Lime application in debug mode on Windows, the console subsystem is used and Windows looks for a `main` function. However, the `Main.cpp` file used when linking statically always defines a `WinMain` function regardless of whether the application is being built in debug mode.

This pull request adds an additional check in `Main.cpp` to define a `main` function instead of `WinMain` when building in debug mode.

### DIFF
--- a/templates/cpp/static/Main.cpp
+++ b/templates/cpp/static/Main.cpp
@@ -1,6 +1,6 @@
 #include <stdio.h>
 
-#ifdef HX_WINDOWS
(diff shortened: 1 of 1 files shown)


player-03:
Good explanation, and the changes look straightforward.

Unfortunately I don't test on Windows, so I can't test this myself. Hopefully someone else can come along and give it a spin.

PR #1531, opened 2022-05-13
Restructure and update submodules

### LINKED ISSUE #1324: Update MojoAL
Opening this issue at the request of a friend.

Apparently there's a bug in the version of MojoAL that OpenFL is currently shipping with. Something to do with Spinlocks, or so I'm told.

In any case, the latest version of MojoAL is supposed to solve this, and I **believe** this is the latest & correct version (but I might be wrong):
https://hg.icculus.org/icculus/mojoAL/file/tip/mojoal.c

### LINKED ISSUE #1368: Rebuilding Android library
I'm trying to fix up a few things but I'm not familiar with Haxe at all, neither with lime or the ecosystem in general.

Rebuilding lime works, but trying to rebuild `ndll/Android/liblime-v7.so` with, I presume, `lime rebuild android -release -armv7` gives the following:

```
x86 Platform: 16
arm Platform: 16
x86_64 Platform: 21
arm_64 Platform: 21
Frameworks should set the minSdkVersion for each APK to these values.

Compiling group: lime
clang++ -Iinclude -Ilib/cairo/include -Ilib/cairo/include/configs/default/ -DLIME_CAIRO -DHAVE_CONFIG_H -DCAIRO_HAS_FT_FONT -Ilib/curl/include/ -DLIME_CURL -DCURL_STATICLIB -Ilib/freetype/include -DLIME_FREETYPE -Ilib/harfbuzz/src -DLIME_HARFBUZZ -DLIBHL_EXPORTS -Ilib/jpeg/ -DLIME_JPEG -Ilib/lzma/src -DLIME_LZMA -Ilib/ogg/include/ -Ilib/vorbis/include/ -DLIME_OGG -Ilib/openal/include/ -DLIME_OPENAL -DLIME_OPENALSOFT -DAL_LIBTYPE_STATIC=1 -DAL_ALEXT_PROTOTYPES -DLIME_OPENGL -Ilib/png/ -Ilib/zlib/ -DLIME_PNG -Ilib/sdl/include/ -Ilib/sdl/include/configs/default/ -DHAVE_LIBC -DLIME_SDL -Ilib/ogg/include/ -Ilib/vorbis/include/ -DLIME_VORBIS -DLIME_ZLIB --target= -DHXCPP_VISIT_ALLOCS(haxe) -DHXCPP_API_LEVEL=0(haxe) -I/usr/local/lib/haxe/lib/hxcpp/4,0,64/include -Iinclude -DANDROID=ANDROID -DHX_ANDROID -DHXCPP_ANDROID_PLATFORM= -fvisibility=hidden -ffunction-sections -fstack-protector -fexceptions -c -fpic -Wno-invalid-offsetof -Wno-return-type-c-linkage -Wno-parentheses ... tags=[haxe,static]
 - src/system/ValuePointer.cpp 
 - src/math/ColorMatrix.cpp 
 - src/ui/TextEvent.cpp 
 - src/media/codecs/vorbis/VorbisBindings.cpp 
 - src/graphics/cairo/CairoBindings.cpp 
 - src/ui/GamepadEvent.cpp 
 - src/text/harfbuzz/HarfbuzzBindings.cpp 
 - src/media/containers/WAV.cpp 
Error: error: unknown target triple 'unknown', please use -triple or -arch
Error: error: unknown target triple 'unknown', please use -triple or -arch
```

I've dug through the code but I can't find much - it seems that Android building is disabled in CI too. How do I go about this?

### LINKED ISSUE #1462: SDL out of date

### LINKED ISSUE #1482: lime window alert
**CODE**
If I do this 
`openfl.Lib.current.stage.window.alert("tetetet","hiahiohaohaa");`

**ERROR**
The window will appear but when I click on "ok", 
I have this error :  XIO:  fatal IO error 11 (Ressource temporairement non disponible) on X server ":0"

**URGENCY**
None. I was just playing with Lime and I noticed this error.
.

### LINKED ISSUE #1530: libjpeg forks
Wikipedia lists [two well-known forks of libjpeg](https://en.wikipedia.org/wiki/Libjpeg#Forks): [libjpeg-turbo](https://github.com/libjpeg-turbo/libjpeg-turbo) is optimized for speed, while [mozjpeg](https://github.com/mozilla/mozjpeg) is optimized for filesize. Should Lime use one of these instead of the original? If so, which?

(Bonus: both are Git repos, meaning they'll be easier to keep up to date.)

player-03:
I'm not a fan of how Lime's submodules have been structured up until now. They were created by hand-copying files, and so in order to update them, you have to hand-copy all the new files. However, almost all of the source projects are public Git repos. This pull request takes advantage of that, pointing directly to these repos. Now, updating is simply a matter of checking out a more recent commit, and perhaps updating files.xml.

- I added a readme file to introduce people to the project folder, Lime's C++ code, and the submodules.
- I switched to libjpeg-turbo, since it has a public Git repo, and is also faster.
- We can (and must) now rebuild for Android using more recent NDKs.
- liblzma is untouched, as it doesn't have a public Git repo.
- While tiny file dialogs has a public repo, it gives an error if you try to use anything other than the most recent commit as a submodule. Because I don't want to have to update the submodule every time they commit, we're going to use a GitHub mirror of that too.

---

Because it's so much easier to update submodules, I went ahead and updated almost all of them to the latest version. ([Excluding OpenAL](https://github.com/player-03/lime/commit/e027811906f70aec28d1046b0c7adee5addf3bf3) until hxcpp receives an update, and excluding LZMA since there's no official public Git repo.)

Of particular note: updating zlib fixes a significant security issue ([CVE-2018-25032](https://github.com/advisories/GHSA-jc36-42cf-vqwj)).

- [ ] In order to make OpenAL work, I included ~4MB of generated files. This could be avoided by using CMake to generate them from source. It adds another dependency, but it's probably worth it.

---

Because some changes had to be reverted in order to pull this off, I've compiled a list of differences. Hopefully, we can find some other way to resolve these issues, now that directly editing the source files is off the table.

Two of the differences are warnings we chose to suppress. They're now back, but is that really a big deal?

- [x] https://github.com/openfl/libpng/commit/601ea3ce9b4af315bed7f371348305b28b3a231c
- [ ] https://github.com/openfl/libopenal/commit/e5b5a559075eff71b899094c19d887aec96e1ff3

Then there are some more specific changes:

- [ ] [Removing a block of code related to font scaling](https://github.com/openfl/libcairo/commit/70d5bd67f3fc3b904f2ce730f61e7403feb62b05).
- [x] [Declaring a variable as `assign` instead of `retain`](https://github.com/openfl/libsdl/commit/370d05558c31570e34c642303705a66a9d0d1547).
- [ ] [Including libkern/OSAtomic.h](https://github.com/openfl/libsdl/commit/fa51fcb56b152ba6419b6decfef5dad48b075ac2#diff-e6af188bc44dbb9466a11886afaca34949ef416c76c102f0fae0fba653951223R33-R34).

None of these are explained in the commit history. CI builds seem to work fine without them, but I notice that the latter two are related to the `HAVE_GCC_ATOMICS` macro. Maybe the CI build defines it, hiding an error. Or maybe it'll work fine now that we updated SDL.


player-03:
I tried updating to NDK r23, but it turns out both [hxcpp](https://github.com/HaxeFoundation/hxcpp/pull/992) and [Pixman](https://gitlab.freedesktop.org/pixman/pixman/-/issues/46) are incompatible with it. Until they're updated, we're stuck with NDK r21e. (But hey, it's better than being stuck with r15c.)

player-03:
Almost forgot to mention: since this pull request switches to libjpeg-turbo, it resolves #1530.

player-03:
Surprise bugfixes, yay!

player-03:
Great! That's one explained.

justin-espedal:
I've tested this targeting iOS/macOS/Android/Hashlink so far and it all seems to work.

Is there anything specific you'd like help with to push this forward?

player-03:
The main question is whether we want this to be part of the big 8.0.0 release. With a change this big, there's always the chance it'll break people's builds. I'd hate to release a year's worth of bug fixes along with a new bug that prevents them from updating. OTOH, there's the argument that this is what major version releases are for.

Besides that, it'd be nice to get a response about what the other unexplained changes were for, but at this point I'm not holding my breath.

justin-espedal:
Regarding the OpenAL issue, maybe requiring cmake isn't too terrible. After all, this is for people who are building lime from source. They already have git and a native toolchain anyway.

justin-espedal:
Alternatively, they could just be part of the git repository. Although I don't prefer such an approach myself, there are lots of other things in the lime git repo that I normally wouldn't include in a source code repository either (like all the minified js in the dependencies folder and all the executables in the templates folders).

player-03:
We're using the version of Clang that comes bundled with the Android NDK. I don't think any other version would work.

Sirox228:
> CI builds seem to work fine without them, but I notice that the latter two are related to the HAVE_GCC_ATOMICS macro. Maybe the CI build defines it, hiding an error. Or maybe it'll work fine now that we updated SDL.

it seems that CI builds work fine, but the app in armv7 is dieing with sdl error: `cannot locate symbol "__atomic_fetch_sub_4" referenced by "(path)....../lib/arm/liblime.so"`
what changes can cause this?

Sirox228:
i can't say why, but recently, this started to happen in this lime version:
```
error: Server does not allow request for unadvertised object 2681e426ddaebc8e2764a7823b4b9d69564d1684
fatal: Fetched in submodule path 'project/lib/tinyfiledialogs', but it did not contain 2681e426ddaebc8e2764a7823b4b9d69564d1684. Direct fetching of that commit failed.
```

player-03:
That's annoying. I guess the tinyfiledialog settings mean we can only fetch the latest commit? Two days ago they made a new commit, which is probably what broke it.

I believe my change will fix it for you, but [`actions/checkout`](https://github.com/marketplace/actions/checkout) isn't so forgiving, hence the errors.

player-03:
Good news: the latest NDK is working! Bad news: only if you use my HXCPP pull request.

flashultra:
What is the current status of the pull request ? Using the latest submodules in official release looks great.

player-03:
Glad to hear it! Can always use more eyes on a large change like this.

This won't be released until Lime 8.1.0 or 8.2.0. I don't think any of the TODOs would block it, but it would certainly be nice to check them off.

player-03:
@flashultra reported the following error with libpng:

> ```bash
> From https://github.com/freedesktop/pixman
> * branch            244383bf9f3493c014985de46876e40fd5db43f3 -> FETCH_HEAD
> Submodule path 'project/lib/pixman': checked out '244383bf9f3493c014985de46876e40fd5db43f3'
> error: Server does not allow request for unadvertised object a40189cf881e9f0db80511c382292a5604c3c3d1
> Fetched in submodule path 'project/lib/png', but it did not contain a40189cf881e9f0db80511c382292a5604c3c3d1. Direct fetching of that commit failed.
> ```

This is similar to the error we saw with tinyfiledialogs, and I was never able to find a direct solution. Instead I switched to a GitHub mirror (ae25706), which is probably my plan here too.

flashultra:
The last change works for me . libpng in github looks official for me.
My understanding about  the error is that:
As github modules contain SHA1 information for a specific commit some servers refuse to directly serve
that SHA1. You can get only branches and tags.
Maybe one possible solution will be to add branch, not to get the HEAD 
At example you need to specify the branch ( not tested) :
```
[submodule "project/lib/tinyfiledialogs"]
	path = project/lib/tinyfiledialogs
	url = https://git.code.sf.net/p/tinyfiledialogs/code
	branch = master
```
Some other suggestions: 
```
git submodule sync --recursive
git submodule update
```
and here is also small discussion : https://buildroot.uclibc.narkive.com/32IZXomj/v3-01-13-core-pkg-download-change-all-helpers-to-use-common-options#post15

player-03:
> Maybe one possible solution will be to add branch, not to get the HEAD

According to [a highly-upvoted StackOverflow answer](https://stackoverflow.com/a/18797720), even if you specify a branch, you still have to point at a specific commit. And even if that wasn't the case, we would still want control over which version we're pointing at. Specifying a tag would be ideal but doesn't seem possible.

> ```bash
> git submodule sync --recursive
> git submodule update
> ```

There are certainly ways to make it work from the command line, but it also needs to work in CI. Unfortunately, [actions/checkout](https://github.com/marketplace/actions/checkout) offers only two options: `submodules: true` and `submodules: false`. Not a lot of room for configuration.

flashultra:
I'm trying to use TextField , but i've got the error ```TextField.hx:1753: characters 17-33 : lime.ui.Window has no field setTextInputRect```
Maybe it's need to be merged with the latest master branch ?

player-03:
Merging wouldn't help. [Judging from TextField.hx](https://github.com/openfl/openfl/blob/9.2.0/src/openfl/text/TextField.hx#L1739-L1754), the problem is that Haxelib isn't reporting the correct version of Lime. For whatever reason, Haxelib judges library version based off the folder name, not haxelib.json.

I'd say the easiest workaround is to rename the folder:

```cmd
> cd [haxelib folder]/lime
> mv git 8,1,0
> haxelib set lime 8.1.0
```

flashultra:
I'm usng lime from submodules branch and there is no method setTextInputRect 
It exist here : https://github.com/openfl/lime/blob/develop/src/lime/ui/Window.hx#L547
but not in submodule branch:  https://github.com/player-03/lime/blob/submodules/src/lime/ui/Window.hx#L542
Renaming my submodules folder to 8,1,0 did not help.

player-03:
I must have misread the code. I thought the first `#end` said `#else`, meaning it would only run `setTextInputRect()` if Lime was older than 8.0.0.

That also means Haxelib might have fixed version number detection. Neat!

player-03:
Ugh, that's probably from merging the upstream changes. I'll check it out.

player-03:
To anyone currently using my branch: please switch to 8.2.0-Dev at your earliest convenience. I'll leave my branch online for the time being, but not forever. Also, any fixes will now go in 8.2.0-Dev.

gepatto:
great!
Just checked out 8.2.0 -Dev on the raspberry pi and it compiles.
I did have to install libpulse-dev tot satisfy openal.

jgranick:
I would like to chime in here briefly.

https://github.com/openfl/libpng/commit/601ea3ce9b4af315bed7f371348305b28b3a231c
https://github.com/openfl/libopenal/commit/e5b5a559075eff71b899094c19d887aec96e1ff3

Both of these messages will print to the console. I found these inappropriate for release builds, as well providing noise to developers. I think I may have also received issues where developers thought they had an error. We ask libpng to decode a PNG for us, and do not need to know if the PNG is "improper." We just need it parsed 🙃 

Similarly, OpenAL was providing a warning (I think on shutdown) related to closing the device. There's a chance that with a different shutdown procedure in Lime, or with current source versions, that we will no longer have this warning. However, it would be great to not have this message at the end of every log for every application running OpenAL.

I wish I knew the exact case where Cairo's font scaling appeared to be incorrect. This may have been related to our older use of the Cairo text ("toy") API, which was meant for demos and not intended for production use. We have since switched to the smarter, more advanced APIs using Harfbuzz and rendering characters individually. Ultimately, the issue would have been that a certain font under certain circumstances rendered or measured at the wrong size.

I recently cherry-picked a commit in Cairo to fix an issue with WebAssembly. This can be ignored if a newer version of Cairo is pulled.

player-03:
Turns out I already made the custom pnglibconf.h file, so it was easy to toggle the warnings. I guess I just missed that option back in 2022. Oops!

As for OpenAL, they seem to have rewritten all of the shutdown code, since I can't find either that warning or the function it was in.

Thanks for the clarifications!

PR #1533, opened 2022-05-26
Add `exported="true"` to AndroidManifest.xml

pozirk:
Otherwise it throws the following error on build:
"Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined. See https://developer.android.com/guide/topics/manifest/activity-element#exported for details."

To publish apps on Google Play everyone needs to target SDK 31+, which is Android 12+, so this is like a must now.
Works fine with older version of SDKs. (I've checked 28.)

If "true", the activity is accessible to any app, and is launchable by its exact class name.
If "false", the activity can be launched only by components of the same application, applications with the same user ID, or privileged system components. This is the default value when there are no intent filters.

### DIFF
--- a/templates/android/template/app/src/main/AndroidManifest.xml
+++ b/templates/android/template/app/src/main/AndroidManifest.xml
@@ -19,7 +19,7 @@
 		<meta-data android:name="SDL_ENV.SDL_IOS_ORIENTATIONS" android:value=  "LandscapeLeft LandscapeRight" />
 		::end::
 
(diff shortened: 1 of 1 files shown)


player-03:
Oh yeah, I remember this one. I fixed it in my own manifest file and never thought about it again.

The fix looks good. Unless CI somehow finds a problem, I'll go ahead and merge.

PR #1535, opened 2022-06-04
Only call `buildBuffer()` when needed.

player-03:
Turns out, [`HTTPRequest`'s call](https://github.com/openfl/lime/blob/02617a854d02777ebb0e9534c1d59642938e8aef/src/lime/net/HTTPRequest.hx#L114) is only needed for internet requests. Files on the local machine will be [loaded directly into `bytes`](https://github.com/openfl/lime/blob/02617a854d02777ebb0e9534c1d59642938e8aef/src/lime/_internal/backend/native/NativeHTTPRequest.hx#L427), leaving `buffer` empty. Calling `buildBuffer()` will then delete the data.


Dimensionscape:
Perfect, this is the correct fix. Thank you!

PR #1538, opened 2022-06-07
Allow running native apps from other directories.

### LINKED ISSUE #1284: Linux target doesn't start unless working directory matches executable
Target:
* Lime 7.2.1
* lime-samples 7.0.0
* openfl 8.8.0
* haxe 3.4.7
* Arch Linux

To reproduce:
* Get BunnyMark from lime-samples
* build `lime build linux`
* `cd Export/linux/bin`
* `./BunnyMark` - observe that it runs
* `cd ..`
* `./bin/BunnyMark` - observe error 

> Error : [lime.utils.Preloader] ERROR: There is no asset library with an ID of "default"

Expected behavior:
Executable starts correctly regardless of working directory

player-03:
It isn't always safe to assume `./` is the app directory, and removing that assumption opens up options.

`Sys.programPath()` requires at least Haxe 3.4, but I don't think Lime supports 3.3 anyway.

Resolves #1284.

### DIFF
--- a/src/lime/system/CFFI.hx
+++ b/src/lime/system/CFFI.hx
@@ -1,6 +1,7 @@
 package lime.system;
 
 #if (!lime_doc_gen || lime_cffi)
+import haxe.io.Path;
 import lime._internal.macros.CFFIMacro;
 #if (sys && !macro)
 import sys.io.Process;
@@ -135,11 +136,16 @@ class CFFI
 
 			__moduleNames.set(library, library);
 
-			result = __tryLoad("./" + library, library, method, args);
+			var programPath:String = ".";
+			#if sys
+			programPath = Path.directory(Sys.programPath());
+			#end
+
+			result = __tryLoad(programPath + "/" + library, library, method, args);
 
 			if (result == null)
 			{
-				result = __tryLoad(".\\" + library, library, method, args);
+				result = __tryLoad(programPath + "\\" + library, library, method, args);
 			}
 
 			if (result == null)
--- a/templates/haxe/ManifestResources.hx
+++ b/templates/haxe/ManifestResources.hx
@@ -3,6 +3,7 @@ package;
 
 
 import haxe.io.Bytes;
+import haxe.io.Path;
 import lime.utils.AssetBundle;
 import lime.utils.AssetLibrary;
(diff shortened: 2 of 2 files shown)


joshtynjala:
> Sys.programPath() requires at least Haxe 3.4, but I don't think Lime supports 3.3 anyway.

I'm okay with Haxe 3.4 (or even 3.4.7 specifically) being the minimum requirement for Lime/OpenFL. 3.4.7 is the only Haxe 3 build that we actually test with, as far as I know.

joshtynjala:
This is breaking asset loading on macOS. The directory returned by `Path.directory(Sys.programPath())` is not correct.

On macOS, `Path.directory(Sys.programPath())` resolves to _MyProject.app/Contents/MacOS_, but assets are placed in _MyProject.app/Contents/Resources_.

I tried `lime.system.System.applicationDirectory`, and that seems to be working. So maybe something like this?

```haxe
#elseif (mac || console)
rootPath = lime.system.System.applicationDirectory;
#elseif sys
rootPath = Path.directory(Sys.programPath()) + "/";
#else
rootPath = "./";
#end
```

Or maybe `lime.system.System.applicationDirectory` is correct on _all_ sys targets?

player-03:
> Or maybe `lime.system.System.applicationDirectory` is correct on _all_ sys targets?

I mean, it seems like it ought to be. (If only it was documented...)

> _MyProject.app/Contents/MacOS_

Is that still the correct place to look for NDLLs, or do we need to update CFFI.hx as well?

player-03:
I also discovered that HashLink still requires you to run it from Export/hl/bin/. This is true whether you use `applicationDirectory` or `programPath()`, so I assume the error happens elsewhere.

It doesn't give a useful error message, it just prints basic instructions on how to run a HashLink app from command line.

Hardly a showstopper, as "running from another directory" is a niche use case, but a little bit annoying. Personally, I'm just going to keep using `lime run hl` and not worry about it.

joshtynjala:
> MyProject.app/Contents/MacOS
> Is that still the correct place to look for NDLLs, or do we need to update CFFI.hx as well?

Yes, that's still where we copy the ndll/hdll files.

Issue #1464, opened 2021-05-20
Linux ( CPP & Neko)  Clipboard doesn't work


Shallowmallow:
**Describe the bug**
If you copy some text from elsewhere, you  should be able to paste it in a textfield
( or lime.system.Clipboard.text  should be different than null ) 

**To Reproduce**
1. copy text somewhere else ( not from lime/openfl program, if it's from the program it will work)
2. trace(lime.system.Clipboard.text)
3. it is null

**Expected behavior**
It should be the copied text

**Targets**
Linux  CPP  and NEKO tested
Hashlink not tested
X11

** Idea to correct  ??**
This library seems to work https://github.com/waneck/systools  on Linux  (  Neko crashes but it  could not link to to neko)

That's their c  https://github.com/waneck/systools/blob/master/project/src/linux/clipboard.c
Maybe you could get some inspiration ?

loudoweb:
I may be wrong but I think lime clipboard works only inside textfield. Can you test to copy/paste from external program inside Input Textfield?

Anyway Lime relies on SDL to use the Clipboard and it is very limited (only text). I'm currently working on an advanced Clipboard (only Windows for now) to copy/paste html, svg and images : https://github.com/loudoweb/native-clipboard

Shallowmallow:
It works when you copy from the openfl textfield to your other openfl textfeld ( from the same program) .
If you open twice the openfl program, you cannot copy from one program to the other.
You cannot copy from an external program to the textfield. ( I wouldn't have discovered  lime.system.Clipboard.text  otherwise ;) ) 

Nice project ! Indeed it could be a useful extension to plenty of people.

PR #1540, opened 2022-06-07
Work around unreliable clipboard on Linux.

### LINKED ISSUE #1298: Clipboard only updates on change
https://github.com/openfl/lime/commit/c49a90d451ebbe2fa5ef6e525e2d1d5d31cb872f re-broke #1196 as far as I've tested. `update()` is called when the clipboard text *changes*, but it is not called once at the beginning of execution. So my previously copied text is not seen by Lime, and re-copying the same text does nothing since it did not *change*.

Calling `update` once at the beginning somewhere would probably work.

player-03:
As reported in multiple issues, the Linux clipboard isn't very reliable. Based on my testing, it only starts working properly after `set_text()` is called, so I figured the easiest option was to ensure that that gets called without actually changing the text.

Downside is that it'll turn any rich text into plaintext, but that's why I modified `get_text()` rather than making an `__init__()` function. That way, we only overwrite the clipboard if the app actually needs it.

Closes #1298 and #1464.

### DIFF
--- a/src/lime/system/Clipboard.hx
+++ b/src/lime/system/Clipboard.hx
@@ -52,16 +52,20 @@ class Clipboard
 	// Get & Set Methods
 	private static function get_text():String
 	{
-		// Native clipboard calls __update when clipboard changes
+		// Native clipboard (except Xorg) calls __update when clipboard changes.
 
 		#if (flash || js || html5)
 		__update();
 		#elseif linux
-		// Hack: SDL won't start calling __update until set_text()
-		// is called once.
+		// Xorg won't call __update until we call set_text at least once.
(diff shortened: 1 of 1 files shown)


player-03:
I spent the time to track down the C++ code that makes it start working, to see if there was a better way. And there isn't really. We'd have to duplicate most of the `X11_SetClipboardText` function (everything except the one `X11_XChangeProperty()` call), which is longer than what I did here, and also less cross-platform compatible.

I also tried updating SDL, but [SDL_x11clipboard.c has barely been touched in years](https://github.com/libsdl-org/SDL/commits/main/src/video/x11/SDL_x11clipboard.c), so of course it didn't make any difference.

PR #1542, opened 2022-06-10
Display preloader progress for packed asset libraries.

justin-espedal:
Before this change, packed asset libraries don't incrementally update the preloader progress bar. This can make it seem like loading has stalled, especially on online targets with large asset packs being delivered over slow connections.

With this change, the download progress of the asset pack is tracked. Additionally, the unpacking progress of each individual asset in the pack is also tracked, though it's assumed to be a much smaller portion of the total time spent loading the asset pack.

Since audio files are not included in packet asset libraries, audio files are loaded separately from other asset types, the same way they would be loaded in a non-packed asset library.

### DIFF
--- a/src/lime/utils/PackedAssetLibrary.hx
+++ b/src/lime/utils/PackedAssetLibrary.hx
@@ -168,17 +168,36 @@ import flash.media.Sound;
 		if (promise == null)
 		{
 			promise = new Promise<AssetLibrary>();
+			bytesLoadedCache = new Map();
 
 			// TODO: Handle `preload` for individual assets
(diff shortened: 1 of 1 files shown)


player-03:
~~Can you update this request to target 8.1.0-Dev?~~

player-03:
The code looks good at a glance, but I have no actual experience with packed asset libraries, so I can't verify it.

joshtynjala:
> The code looks good at a glance, but I have no actual experience with packed asset libraries, so I can't verify it.

Same

zanzlanz:
Anecdotally, I applied these changes to my environment, and it works as I was hoping. It allows me to track preload progress when using packed assets, whereas before it was not possible.

Would love to see this merged.

PR #1543, opened 2022-06-10
Use the newer Xcode build system introduced in Xcode 10

justin-espedal:
Up until now the legacy Xcode build system has been being used.
As of Xcode 14, the legacy build system has been removed ([source](https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes)) :
> The legacy build system has been removed. (90801041)

In order to keep products being built to a known location local to
the project folder, the `derivedDataPath` is passed to all calls to
`xcodebuild`. This also requires the desired `scheme` to be passed.
Previously, `scheme` was only passed for archive operations, but now
it's a common argument for all calls to `xcodebuild`.

Without specifying the desired `derivedDataPath`, it would end up being
"`~/Library/Developer/Xcode/DerivedData/{generated-project-id}`", and
`xcodebuild` would need to be queried for the exact path.

_I **have not** yet actually confirmed for myself that the previous code no longer
works with Xcode 14. I also haven't tested the tvOS changes at all. I modeled
those changed after the iOS changes, which work as expected. The iOS changes
have been tested with Xcode 14 and Xcode 13._


joshtynjala:
> I also haven't tested the tvOS changes at all. I modeled those changed after the iOS changes, which work as expected.

This is fine. Lime's tvOS support is currently broken (has been for a while now). Hopefully, we can fix it up eventually.

player-03:
A couple months back, I was still able to use the old build system. I think we should treat this as a new feature, and merge it into 8.1.0-Dev.

PR #1545, opened 2022-06-12
Suppress `uiMode` configuration changes.

### LINKED ISSUE #1532: Android app exits on theme change
There is a problem, that most likely related to SDL.
If I change Android theme from light to dark and vice versa, OpenFL app closes.
It is not really crashing, as I couldn't find much in logs.
Happens to any app, here is PiratePig video of the problem:
https://youtube.com/shorts/hZSR0HdVa2Y

It wouldn't be a big deal, but Samsung AppStore suspended my game because of this. :confused:

Anyone has any idea on how to fix that?
Thank you!

player-03:
When the user switches between dark and light themes, Android destroys and recreates the activity. However, Lime activities can't be recreated like this, and just close instead.

Since we don't actually use the system theme, there's no reason for Android to do this. Simplest solution is to suppress it. The only question is, should we expose [`onConfigurationChanged()`](https://developer.android.com/reference/android/app/Activity#onConfigurationChanged\(android.content.res.Configuration\)) to extensions, or is this too niche to be worth it?

References:
https://developer.android.com/guide/topics/resources/runtime-changes
https://developer.android.com/guide/topics/ui/look-and-feel/darktheme#config-changes

Closes #1532.

### DIFF
--- a/templates/android/template/app/src/main/AndroidManifest.xml
+++ b/templates/android/template/app/src/main/AndroidManifest.xml
@@ -19,7 +19,7 @@
 		<meta-data android:name="SDL_ENV.SDL_IOS_ORIENTATIONS" android:value=  "LandscapeLeft LandscapeRight" />
 		::end::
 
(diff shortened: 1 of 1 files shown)


player-03:
We suppress other events too, and it isn't like we expose `onConfigurationChanged()` for those. I think this is fine as-is.

Issue #1546, opened 2022-06-12
Stale output from previous builds can cause errors


justin-espedal:
Lime tends to avoid unnecessary file operations where possible, which is good. Avoiding unnecessary file IO is faster, and some external build systems rely on file timestamps to determine whether a file has been modified. However, since the output folder is not cleaned before each build, this can lead to problems due to stale files remaining in the build output.

For a real-world example:
1. Take a lime project that has [Stencyl's iap extension](https://github.com/Stencyl/extension-purchases) enabled.
2. Build the project for Android. The export folder now contains Java source files from the extension.
3. Update the extension past [this commit](https://github.com/Stencyl/extension-purchases/commit/6bba0ed463868b3bec96920d7b993a71b29f9a6b#diff-f80e7ef4daa7f4dfccfe16d99e1d9e3822ae4aaf048fe07cf2dfb8f8ade9271a) (notice that a number of Java source files have been deleted).
4. Build the project for Android again. The lime output / gradle input folder now contains a mixture of the old Java source files and the new ones, causing a compilation error during the gradle build.

Of course the easiest way to get a correct build in this situation is to just do a clean build. But I'd like to explore options that allow us to have a correct build without sacrificing the cache, even in cases like this. There should be some mechanism in place that allows the build process to clean out stale files from old builds.

I've prepared one potential approach, and have put it to use in the Android platform, which I'll post soon.

player-03:
Also worth noting that the vast majority of an Android/iOS build is spent compiling C++ code, which is cached in the obj folder. Deleting the bin folder wouldn't slow compilation that much, and would also fix this. I'm still happy to talk about the automated approach, but should Lime also add a `-clear-templates` flag? (Or some other name?)

player-03:
I do think there's a better way. While looking at the code, I realized a few key things.

- There's no need to call `markFileAsTouched()` if the file is overwritten. Instead, we can check the timestamp.
- All non-template assets are included in an asset manifest. We can read the manifest for an up-to-date list.
   - It is possible for asset manifests themselves to be stale, but we can tell by checking the timestamp.
- All templates are ultimately handled by `hxp.System.copyFile()`.
   - Custom source files like the ones in Stencyl's iap extension are all processed as templates.

That last one is extra important, because it means we don't even need to worry about ordinary assets. We could clean them up if we had to, but they aren't going to break anything.

Pull requests pending.

player-03:
Ok, I submitted my pull requests. What do you think?

[...]

PR #1547, opened 2022-06-12
Ensure removal of stale files from previous android asset updates

justin-espedal:
This is the lime part of a potential solution to [openfl/lime#1546 (Stale output from previous builds can cause errors)](https://github.com/openfl/lime/issues/1546), applied only to the android platform. This relies on https://github.com/openfl/hxp/pull/27.

`System.markFileAsTouched` is called as appropriate in AssetHelper, IconHelper, and ProjectHelper.

`app/src/main/jniLibs` is ignored because that folders isn't modified as part of the `update` step, but later on as part of the `build` step. `app/build` is ignored because that's the gradle build folder. We don't directly touch it ourselves, and gradle can hopefully be trusted to keep that in order as long as we provide the correct inputs to the gradle build.

### DIFF
--- a/src/lime/tools/AssetHelper.hx
+++ b/src/lime/tools/AssetHelper.hx
@@ -43,6 +43,7 @@ class AssetHelper
 		}
 		else
 		{
+			System.markFileAsTouched(destination);
 			try
 			{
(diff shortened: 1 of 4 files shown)


player-03:
Closing this because #1790 was merged.

PR #1550, opened 2022-06-13
Delete stale templates after each build.

player-03:
Template files have a habit of sticking around in the output folder, even if the project no longer needs or wants them. Clearing old templates can improve app stability.

Like #1547, this is one half of a potential solution to #1546. (The other half is openfl/hxp#28.) And I'd argue it's a better solution, requiring less maintenance and not forcing users to update hxp at the same time as Lime.

### DIFF
--- a/src/lime/tools/PlatformTarget.hx
+++ b/src/lime/tools/PlatformTarget.hx
@@ -96,6 +96,10 @@ class PlatformTarget
 			// AssetHelper.processLibraries (project, targetDirectory);
 			// #end
 			update();
+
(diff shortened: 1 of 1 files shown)


player-03:
The conditional compilation block is going to prevent CI from noticing the change. (But if it weren't there, it would crash because CI still uses HXP 1.2.2.) If you want to test this yourself, make sure to pull my HXP branch, open HXP's haxelib.json file, and set `"version": "1.2.3"`.

player-03:
Update: I think the new pull request is a better approach.

PR #1551, opened 2022-06-15
Expose more information to project.xml users.

### LINKED ISSUE #1486: How to retrieve target architecture using defines?

player-03:
Previously, they could access `${app.file}` and `${config.android}`, but not `${project.platformType}` or `${config.android.target-sdk-version}`. Now they can!

I don't know if we'd want to phase out `${projectDirectory}`, but it's now an option.

### DIFF
--- a/src/lime/tools/ProjectHelper.hx
+++ b/src/lime/tools/ProjectHelper.hx
@@ -111,8 +111,10 @@ class ProjectHelper
 		{
 			return project.environment.get(string);
 		}
+		// TODO: Should we start phasing this out?
 		else if (string == "projectDirectory")
 		{
+			Log.info("", "Consider using ${project.workingDirectory} instead of ${projectDirectory}.");
 			return project.workingDirectory;
 		}
 		else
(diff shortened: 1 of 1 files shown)


player-03:
Some additions I've considered:

- Access map properties. I've already implemented this, still using the dot access syntax. It's a bit weird considering that isn't how Haxe works.
- Add a `parent` variable to `ProjectXMLParser` that points to the project above. Then you can use dot access to get `parent.parent.workingDirectory` or whatever.
- Make some adjustments so that you can always use the result value in conditionals. For instance, `if="${targetFlags.clean}"`.

player-03:
~~Pretty sure I should have used `Reflect.getProperty()` instead of `Reflect.field()`. But I'm not confident enough about how `Reflect` works to change it this close to release.~~ Edit: the documentation says it falls back to the same behavior, and it passes initial tests, so I did it.

PR #1557, opened 2022-07-04
[wip] Audio 3D position/panning changes (add panning, both panning and position off by default)

justin-espedal:
There are a few issues with sound position that this PR aims to improve.

- 3D position is used to simulate panning on non-Flash targets. This is ok for native, where we use OpenAL and there's no alternative for panning. On HTML5, however, howler has a stereo mode for playing sounds. That gives a better result than using the awkwardly simulated panning based on 3D audio.

- Using 3D audio at all on HTML5 can lead to serious degradation of audio quality ("muffled" sounds).
  [Stencyl Issue Tracker: Muffled Sounds On HTML5](https://community.stencyl.com/index.php?issue=1509.0)

- Using 3D audio positioning at all can lead to extra audio processing code being run, even if the position is (0, 0, 0), so it should be avoided if not desired.

This makes position a nullable value, null by default, which is a change to the AudioSource API.

### DIFF
--- a/src/lime/_internal/backend/flash/FlashAudioSource.hx
+++ b/src/lime/_internal/backend/flash/FlashAudioSource.hx
@@ -114,6 +114,15 @@ class FlashAudioSource
 		return loops = value;
 	}
 
+	public function setPan(value:Float):Float
+	{
+		var soundTransform = channel.soundTransform;
+		soundTransform.pan = value;
+		channel.soundTransform = soundTransform;
+
+		return value;
+	}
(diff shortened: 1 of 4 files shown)


DigiEggz:
As mentioned [here](https://github.com/openfl/openfl/pull/2570), these changes have vastly improved the sound quality for me. Paired with my recent [PR](https://github.com/openfl/lime/pull/1788/), sound in HTML5 is now comparable to native.

PR #1561, opened 2022-07-09
Added timestamp to KeyDown and KeyUp events.

EliteMasterEric:
I am looking into methods of retrieving a highly precise timestamp corresponding with a user's keypress. This is important for applications which need to know exactly when a given key was pressed (which, among other things, includes video games such as rhythm games or fighting games, which demand precise inputs).

If each key event includes a precise hardware timestamp, recorded before any of the processing required to call event listeners is done, then said processing can be compensated for by retrieving a new timestamp and calculating the difference.

I discovered in my research that the `SDL_KeyboardEvent` which Lime receives (on native) does include such a timestamp, however the value is not retained, so I made this pull request to include it.

The pull request is a draft pending the resolution of the following issues:
- The change to the method signature of `lime.app.Application.onKeyDown` is a breaking change which affects OpenFL, Flixel, and other dependent libraries, is there a more stable method to make this change?
- The value provided in the `SDL_KeyboardEvent` is millisecond accurate, and is the value of `SDL_GetTicks` function, which is based on the duration since SDL was initialized, but there is not currently an accessible method to retrieve this value in Lime to perform comparison with.
- HTML5 platforms also have a timestamp associated with keyboard events, but the timestamp value is an absolute epoch time instead. How should this discrepancy be handled?

### DIFF
--- a/project/include/ui/KeyEvent.h
+++ b/project/include/ui/KeyEvent.h
@@ -25,6 +25,7 @@ namespace lime {
 		int modifier;
 		KeyEventType type;
 		int windowID;
+		int timestamp;
 
 		static ValuePointer* callback;
 		static ValuePointer* eventObject;
--- a/project/src/backend/sdl/SDLApplication.cpp
+++ b/project/src/backend/sdl/SDLApplication.cpp
@@ -575,6 +575,7 @@ namespace lime {
 			keyEvent.keyCode = event->key.keysym.sym;
 			keyEvent.modifier = event->key.keysym.mod;
 			keyEvent.windowID = event->key.windowID;
+			keyEvent.timestamp = event->key.timestamp;
 
 			if (keyEvent.type == KEY_DOWN) {
 
--- a/project/src/ui/KeyEvent.cpp
+++ b/project/src/ui/KeyEvent.cpp
@@ -12,6 +12,7 @@ namespace lime {
 	static int id_modifier;
 	static int id_type;
 	static int id_windowID;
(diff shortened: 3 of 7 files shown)


player-03:
> The change to the method signature of `lime.app.Application.onKeyDown` is a breaking change which affects OpenFL, Flixel, and other dependent libraries, is there a more stable method to make this change?

For now we could use a compile-time define, like `#if lime_key_timestamps`. Change the method signature if and only if they set the define, and then users who don't opt in won't experience any changes.

> HTML5 platforms also have a timestamp associated with keyboard events, but the timestamp value is an absolute epoch time instead. How should this discrepancy be handled?

This has precedent in the `Timer.stamp()` function, so we could handle it the same way. Simply tell people that they should only look at differences between values, not at the individual values themselves.

> The value provided in the `SDL_KeyboardEvent` is millisecond accurate, and is the value of `SDL_GetTicks` function, which is based on the duration since SDL was initialized, but there is not currently an accessible method to retrieve this value in Lime to perform comparison with.

I wonder if this change would be better off as its own Haxelib. Not saying it would, necessarily, just that it's something to consider.

Pros:
- Whenever you found a new function you need to expose, you could just do it.
- You could focus on C++ in particular, instead of worrying about cross-platform consistency.
- You wouldn't have to worry about breaking changes.

Cons:
- It wouldn't be integrated into OpenFL/Lime, though that seems manageable. You didn't need these events to be tied to the display list, did you?

PR #1570, opened 2022-08-04
Implement window transparency.

Apprentice-Alchemist:
The FNF modders want transparent windows.
The implementation has not been tested.

### DIFF
--- a/src/lime/_internal/backend/native/NativeCFFI.hx
+++ b/src/lime/_internal/backend/native/NativeCFFI.hx
@@ -347,6 +347,10 @@ class NativeCFFI
 
 	@:cffi private static function lime_window_warp_mouse(handle:Dynamic, x:Int, y:Int):Void;
 
+	@:cffi private static function lime_window_get_opacity(handle:Dynamic):Float;
+
+	@:cffi private static function lime_window_set_opacity(handle:Dynamic, value:Float):Void;
+
 	@:cffi private static function lime_window_event_manager_register(callback:Dynamic, eventObject:Dynamic):Void;
 
 	@:cffi private static function lime_zlib_compress(data:Dynamic, bytes:Dynamic):Dynamic;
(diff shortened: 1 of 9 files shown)


Dimensionscape:
Just a side note, SDL's opacity affects the entire window and its content. I thought I would throw that out there since it might not be clear, and because some developers would possibly be interested in setting the actual window transparency(and its background) while leaving the window content opaque (as is already possible for the AIR target). 

Unfortunately SDL does not support this at the moment. 😢

src/lime/_internal/backend/native/NativeCFFI.hx
```diff
@@ -1357,6 +1367,10 @@ class NativeCFFI
 
 	@:hlNative("lime", "hl_window_warp_mouse") private static function lime_window_warp_mouse(handle:CFFIPointer, x:Int, y:Int):Void {}
 
+	@:hlNative("lime", "hl_window_get_opacity") private static function lime_window_get_opacity(handle:CFFIPointer):Float;
```
47rooks, line 1370: This function needs a body. It compiles the libs without it but a calling program cannot build. I changed it to :

	@:hlNative("lime", "hl_window_get_opacity") private static function lime_window_get_opacity(handle:CFFIPointer):Float {
		return 0;
	};
	
	For consistency with other functions I added an empty function body to set_opacity below too.

Jsa010:
Hilarious

Jsa010:
also how it works?

EyeDaleHim:
PR works well, saw it mentioned by some other dude
image

player-03:
Sorry for making a dozen different commits. The online UI only allows editing one file at a time, and I didn't feel like going to the effort of checking it out.

Most of the changes were sorting and formatting, but two are worth noting:

- [In SDLWindow.cpp, I removed the `if` statement`](https://github.com/openfl/lime/blob/42e345061429abe1c7e74d523b25659fe6859bb8/project/src/backend/sdl/SDLWindow.cpp#L592-L600). Looking at [the documentation](https://wiki.libsdl.org/SDL_GetWindowOpacity), it seems unlikely to ever actually return an error, given that Lime checks the handle before calling it. Plus that's how the rest of the file is written.
- I [replaced `Window.getOpacity()` and `Window.setOpacity()` with getters and setters](https://github.com/openfl/lime/pull/1570/commits/6b2724b147c636673d00738b45c1d6f25745184e), since that's how the rest of the `Window` class works.

And for future reference, FlashWindow.hx and NativeWindow.hx still aren't fully sorted. Something to do later, and not as part of this pull request.

player-03 (approved):
Works correctly on Linux, code looks good.

Raltyro (approved):
(was saying "fnf modders need this" necessary..?)

player-03:
> (was saying "fnf modders need this" necessary..?)

It's always good to know why a change is being suggested. Granted, this wasn't a very thorough explanation, but it's a simple change and apparently the modders in question are happy with the outcome, so why not?

Jsa010:
Omg it exist

PR #1579, opened 2022-09-13
Implement `clickCount` on HTML5 and Native.

ShaharMS:
Added missing implementation for counting click sequences. needs to be merged for (https://github.com/openfl/openfl/pull/2587)


ShaharMS:
The `FileDialog` update is in the commit log becuase of my last contribution, doesnt do anything though

ShaharMS:
woops, turns out i do need Flash & air implementations for clickCount (even if theyre very basic)

joshtynjala:
> The FileDialog update is in the commit log becuase of my last contribution, doesnt do anything though

You should put each PR into a different branch to avoid this. It'll make things easier to merge.

ShaharMS:
> > The FileDialog update is in the commit log becuase of my last contribution, doesnt do anything though
> 
> You should put each PR into a different branch to avoid this. It'll make things easier to merge.

Yeah, luckily this update has already been merged so it doesn't change anything. About flash & air, turns out I don't need to implement those

ShaharMS:
Looks like lime sample fails because it listens to `window.onMouseDown`, which now requires 4 arguments  instead of 3. Where are the `lime samples`?

player-03:
https://github.com/openfl/lime-samples

ShaharMS:
> https://github.com/openfl/lime-samples

Do I need to make a PR there too before this gets merged? (A pr to add a cluckCount argument to onMouseDown functions)

player-03:
I don't know if we're going to merge this at all, given that it breaks existing code. If we did merge it, then yes, someone would need to update the samples.

ShaharMS:
> I don't know if we're going to merge this at all, given that it breaks existing code. If we did merge it, then yes, someone would need to update the samples.

The reason I want it to get merged is because clickCount is needed for word & paragraph selection in TextField. There's no way for it to not break existing code  given that it changes the function signature in the c/c++ code, and the extra arguments needs to be present somewhere in mouse event management

ShaharMS:
The selection thing is actually the reason we discovered clickCount was unimplemented

player-03:
Tricky. If only things were implemented differently, maybe there would be a non-breaking way to do this. Like passing the `MouseEventInfo` object instead of individual arguments. But of course, at this point that would be an even bigger breaking change.

That said, `clickCount` is listed in the documentation, which means it definitely should be implemented, breaking change or no. It still isn't appropriate for a micro patch like 8.0.1, but maybe a minor patch like 8.1.0. We'll have to get some more opinions on that.

Dimensionscape:
We are definitely going to implement this. I was going to do it myself over the past weekend as I already mentioned but I didn't quite have enough time and you guys got a bit ahead of me. Slow down a bit and give me a chance to look at this more so I can provide some better input.

joshtynjala:
I think if we want to avoid breaking changes, we need to add a separate property for click count somewhere in Lime that a listener for `onMouseDown` can access.

Perhaps simply on `lime.ui.Window`.

```hx
public var onMouseDown(default, null) = new Event<Float->Float->MouseButton->Void>();
public var clickCount(default, null):Int = 0;
```

**It's not ideal**, but I don't see many other options, unless we're okay with a breaking change that requires Lime 9. I think it's too soon to jump to 9, so if we really need a breaking change, this PR should probably wait. Personally, I don't think that this PR should wait, so we need to consider this or other alternatives.

player-03:
Now that you mention it, [I did the same thing](https://github.com/player-03/lime/blob/a95f398ef4b409cc31814875c8a9766479d30872/src/lime/system/WorkOutput.hx#L76-L81) in this situation when rewriting the threading classes. I looked for other options but didn't find any good ones.

Dimensionscape:
@joshtynjala I think we should just handle it exclusively internal and use it temporarily until Lime 9 in OpenFL, similarly to how we discussed doing Process to fix the issue from hxcpp. There will be no noticeable changes in Lime for now. I think it provides a clean stop-gap measure that avoids any confusion between versions and gives us time to save any breaking changes for a major version release.

joshtynjala:
Yeah, making it an internal API that only OpenFL can access for now sounds fine.

ShaharMS:
I don't think I have any problem (time-wise) trying to extract clickCount from onMouseDown. Yhe question is how should I do that

ShaharMS:
If I manage to make this not a breaking change, would it be merged soon?

player-03:
I think the plan for now is

```haxe
public var onMouseDown(default, null) = new Event<Float->Float->MouseButton->Void>();
@:allow(openfl.display.Stage)
private var clickCount:Int = 0;
```

ShaharMS:
Why private? Isn't this info useful for lime users too?

I think going the extra mile by making it public and (default, null) is preferred

ShaharMS:
Where isouse related data usually stored? Just want to be consistent

ShaharMS:
Ok, this is what i think can be implemented, and it might be a nice structure to adopt temporarily to avoid breaking changes:

Ill create a new file called `MouseData`, in which ill insert a public, unsettable field called `clickCount`. other status changes/values can be passed onto that structure. when dispatching `MouseEvent.MOUSE_DOWN/MOUSE_UP/CLICK/DOUBLE_CLICK`, ill get the value from the `MouseData` class and dispatch it. its not a breaking change, and it welcomes other types of enrichment in the mouse/keyboard/joystick department:

```haxe
//MouseData will be in lime.data
package lime.data;

@:allow(openfl.display.Stage)
class MouseData {
    public var clickCount(default, null):Int;
    //MouseData can be easily expanded when needed

    function new() {
        clickCount = 0;
    }
}
```

```haxe
//in Stage.hx

MouseEvent.__create(..., mouseData.clickCount);
```

Im ofc open to other suggestions, i like this one because it allows expanding upon existing i/o features without any breaking changes to both OpenFL and Lime.

joshtynjala:
> Why private? Isn't this info useful for lime users too?

Yes, it is useful for Lime users, but we need to consider future maintenance and user impact as well.

We would prefer to change the signature of `onMouseDown`, but we can't until the next major version. So, as an alternative, we create a separate property for `clickCount`. That's where things get tricky. If we exposed `clickCount` publicly, do we keep it around if we later add it to `onMouseDown` too? Or does it get removed, since it will become redundant? It's either a breaking change or legacy code to be maintained. Both have downsides.

If we start out by making it private for OpenFL use, we avoid those downsides, while creating minimal impact on the Lime API. Sure, there's yet another downside that Lime-only users won't be able to access it, but they couldn't before either, so it's not like we're taking anything away. Moreover, it's technically possible to write code with the existing APIs to calculate click count manually.

In the future, as Lime 9 approaches, we can look again at how to expose the click count for Lime users too. We might add it to `onMouseDown`, as I mentioned. Or we might even keep the `clickCount` property, but change it from private to public, if our final decision is that it's too much of an impact on a core API like `onMouseDown`, even for a major version. We still have that option.

Finally, right now, I think that it's more important to get `clickCount` in OpenFL. Spending too much time worrying about how to expose the API in Lime today will delay that, when all of us have limited time to work on these projects.

ShaharMS:
ok, ill get to implementing rn.

what i think im gonna do is set the clickCount every time a `MouseEventInfo` object of type `MOUSE_UP` or `MOUSE_DOWN` is created. is that ok?

ShaharMS:
ive just removed the breaking change, can somebody run CI to check?

ShaharMS:
oh it runs automatically 😅

ShaharMS:
ive just realized the value isnt getting passed onto the function anymore 🤔 lemme fix this and then merge this

ShaharMS:
everything works now, tested on HL, neko and html

src/lime/_internal/backend/native/NativeApplication.hx
```diff
@@ -336,17 +337,18 @@ class NativeApplication
 			{
 				case MOUSE_DOWN:
 					window.onMouseDown.dispatch(mouseEventInfo.x, mouseEventInfo.y, mouseEventInfo.button);
+					window.clickCount = mouseEventInfo.clickCount;
 
 				case MOUSE_UP:
 					window.onMouseUp.dispatch(mouseEventInfo.x, mouseEventInfo.y, mouseEventInfo.button);
+					window.clickCount = mouseEventInfo.clickCount;
```
player-03, line 344: Shouldn't we set `window.clickCount` _before_ dispatching the event? OpenFL records the value before returning from `__onLimeMouseDown` or `__onLimeMouseUp`.

Also probably worth setting it back to 0 after the event is done, to minimize cross-contamination. (Though even that won't stop it from showing up in the pending `MOUSE_MOVE` event.)

src/lime/_internal/backend/native/NativeApplication.hx
```diff
@@ -1,5 +1,6 @@
 package lime._internal.backend.native;
 
+import haxe.Int64;
```
player-03, line 3: Why?

src/lime/_internal/backend/html5/HTML5Window.hx
```diff
@@ -839,6 +842,7 @@ class HTML5Window
 
 							if (touch == primaryTouch)
 							{
+								parent.clickCount = event.detail;
```
player-03, line 845: [According to the documentation](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail), `detail` is only available for `click`, `dblclick`, `mousedown`, and `mouseup` events. I'm guessing that since you can have multiple touches at once, they decided the idea of a "touch sequence" didn't make sense.

src/lime/_internal/backend/native/NativeApplication.hx
ShaharMS: We should, that's an oversight on my part. I'll fix that in the next commit. I didn't set it back to 0 for lime users (i wanted the value to still be accessible via window.clickCount when adding a listener to lime's mouse events)

src/lime/_internal/backend/native/NativeApplication.hx
ShaharMS: Sometimes, when you type a class name and want to auto-import it, it inserts the wrong class, and auto imports it. Probably didn't pay enough attention, my bad. Nice catch!

src/lime/_internal/backend/html5/HTML5Window.hx
ShaharMS: Makes sense, I'll fix that too. Another nice catch :D

ShaharMS:
can somebody run CI again? haxelib misbehaving...

src/lime/_internal/backend/html5/HTML5Window.hx
```diff
@@ -805,7 +805,10 @@ class HTML5Window
 
 				if (touch == primaryTouch)
 				{
-					parent.clickCount = event.detail;
+					// touches shouldn't raise clickCounts value
+					// but they also indicate the end of a click sequence 
+					// (you expect click count to reset after you stop clicking with the mouse)
+					parent.clickCount = 0;
```
player-03, line 811: When only using one finger, browsers will dispatch both a touch event (sans click count) and a click event (with click count). Proof: [this demo](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#result) works on your phone, as long as you use only one finger.

So this code is either doing nothing or overwriting useful data. I'd say touch events just shouldn't interact with `clickCount` at all.

src/lime/_internal/backend/native/NativeApplication.hx
player-03: > i wanted the value to still be accessible via window.clickCount when adding a listener to lime's mouse events

But `onMouseDown` and `onMouseUp` are Lime's event dispatchers. Once those return, the mouse event listeners are done.

src/lime/_internal/backend/html5/HTML5Window.hx
ShaharMS: Interesting! Didn't know that. Let me fix that real quick

player-03:
The main question I still have is cleanup. Here's how I see it:

If we could, we would have added another argument to the event. However, we can't do this because it would break backwards compatibility. Instead, our best option is to add a private variable. Because this private variable is replacing a function argument, it must exist for the duration of the function, and doesn't need to exist afterwards.

ShaharMS:
I'll do it your way (cleanup wise). I'll commit when I'm done. Thanks for the help :)

player-03 (approved):

ShaharMS:
Yeah, I requested by accident and for some reason didn't find a way to cancel the request 😆

PR #1583, opened 2022-09-22
Implement `onDropStart`, `onDropEnd` for file dropping

ShaharMS:
currently' lime only has the `onFileDrop` function for getting the name of the dropped file. ive added support for 3 more, but only implemented 2 of them:

- `onDropStart` - dispatches before any dropping of data
- `onDropEnd` - dispatches after any dropping of data
- ~~`onDropText`~~ - is implemeted on the native side, but only works for X11 on linux, so commented out on the haxe side.


ShaharMS:
thats so strange - CI fails when trying to install from haxelib. happened to me like 2 days ago too... what is going on?

Dimensionscape:
> thats so strange - CI fails when trying to install from haxelib. happened to me like 2 days ago too... what is going on?



> thats so strange - CI fails when trying to install from haxelib. happened to me like 2 days ago too... what is going on?



> thats so strange - CI fails when trying to install from haxelib. happened to me like 2 days ago too... what is going on?

No response from the haxelib server. Might be down.

ShaharMS:
it is working for windows tho 🤔 
can you rerun CI @Dimensionscape?

Dimensionscape:
Yeah, I'll try later, the haxelib server is just being slow/unresponsive at the moment.

player-03:
Huh. I was expecting `onDropStart` to be something like JavaScript's [`dragover` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dragover_event), letting you know that the user is currently holding a file over the app. I guess I didn't pay enough attention to the name.

For the record, here's what actually happens. No events are dispatched while the user holds onto the file(s). Once they let go, several events are dispatched in succession: `onDropStart`, followed by one or more `onDropFile`s, followed by `onDropEnd`.

Basically if you're collecting the files into an array, this lets you know where that array starts and ends. And you know what? That could be helpful for JS, which dispatches an array of file objects (despite the documentation saying the listener should expect a single `String`).

ShaharMS:
> Ok, [I went ahead and implemented the JS fix](https://github.com/ShaharMS/lime/pull/1), and it works pretty well. The question is, should it be part of this pull request? It's technically a breaking change, even though all it does is make JS match the documentation.

actually, idk... we should probably ask the rest of the core team. should i ping them?

player-03:
Nah, I already pinged Chris today. I'll just post it in Discord.

.vscode/settings.json
```diff
@@ -18,4 +18,7 @@
     },
     "files.trimTrailingWhitespace": true,
     "C_Cpp.autoAddFileAssociations": false,
+    "cSpell.enableFiletypes": [
+        "!cpp"
+    ],
```
player-03, line 23: Does this need to be added to Lime? It looks like something that belongs in user settings.

ShaharMS: It doesnt, looks like I used that extension while working on lime by accident. I'll remove this today (even right now if I can through GitHub...)

ShaharMS:
looks like CI failing on ios and getting stuck on android isn't something relate to my pr...
is there a problem with the branch's lime version compiling to mobile?

player-03:
Yeah, the latest hxcpp release broke those two specific tests.

player-03:
Switching over to 9.0.0 so we can include HTML5 support at the same time. I guess we don't have to do it this way, but it feels wrong for a new feature to not work on a major platform.

PR #1589, opened 2022-10-20
Fix openFile not working on Linux.

Vulpicula:
Thanks to LeotomasMC for some assistance with this.

I noticed and was quite confused by the fact that certain things in Flixel failed to work on Linux, such as FlxG.openURL, which returned "xdg-open: unexpected argument '&'". This appears to fix that and seems to work quite fine, with that segment of Flixel now working once more.

May be good to have someone more qualified than I check this out before er... merging, as I'm not exactly the best programmer around.

### DIFF
--- a/src/lime/system/System.hx
+++ b/src/lime/system/System.hx
@@ -298,7 +298,7 @@ class System
 			#elseif mac
 			Sys.command("/usr/bin/open", [path]);
 			#elseif linux
-			Sys.command("/usr/bin/xdg-open", [path, "&"]);
+			Sys.command("/usr/bin/xdg-open", [path]);
 			#elseif (js && html5)
 			Browser.window.open(path, "_blank");
 			#elseif flash


player-03:
Looks like someone wanted [`Sys.command()`](https://api.haxe.org/Sys.html#command) not to block. In a shell, you can do this by adding `&` after all the arguments, but that's a special message to the shell, outside the argument list. Makes sense that it didn't work here.

We could use [`Process`](https://api.haxe.org/sys/io/Process.html) instead, but I figure `xdg-open` is quick enough that it's probably fine.

player-03:
Just checked and we also don't include the ampersand for `<prebuild open="file" />` or `<postbuild open="file" />`. Almost no one uses those, but then again, how many people use this? Lime sure doesn't; Lime only calls it from `WindowsPlatform`, which obviously uses the Windows-specific command.

Pretty sure it never worked and just never got noticed until now, making this a straightforward bug fix.

PR #1599, opened 2022-11-16
Update to the latest version of Gradle

MAJigsaw77:
Updating `Gradle` to the latest version can bring us support for `Java 18`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -215,7 +215,7 @@ jobs:
 
       - uses: actions/setup-java@v1
         with:
-          java-version: 8
+          java-version: 11
 
       - uses: krdlab/setup-haxe@v1
         with:
@@ -460,7 +460,7 @@ jobs:
 
       - uses: actions/setup-java@v1
         with:
-          java-version: 8
+          java-version: 11
 
       - uses: krdlab/setup-haxe@v1
         with:
--- a/tools/platforms/AndroidPlatform.hx
+++ b/tools/platforms/AndroidPlatform.hx
@@ -459,8 +459,8 @@ class AndroidPlatform extends PlatformTarget
 			"android.permission.VIBRATE",
 			"android.permission.ACCESS_NETWORK_STATE"
 		]);
-		context.ANDROID_GRADLE_VERSION = project.config.getString("android.gradle-version", "5.6.3");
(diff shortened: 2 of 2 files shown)


player-03:
Note that you can also configure this on a per-project basis:

```xml
<config:android gradle-version="7.5.1" gradle-plugin="7.3.1" />
```

MAJigsaw77:
> Note that you can also configure this on a per-project basis:
> 
> ```
> <config:android gradle-version="7.5.1" gradle-plugin="7.3.1" />
> ```

Ik this but will be better to update on `lime` itself, will be better for all the people, not only me in this case.

MAJigsaw77:
The latest gradle is better then the old one

flashultra:
Before this update maybe is good to check the compability with HaxeExtension projects ( https://github.com/HaxeExtension ) .
It is possible that most (if not all) of them will stop working.

MAJigsaw77:
> Before this update maybe is good to check the compability with HaxeExtension projects ( https://github.com/HaxeExtension ) . It is possible that most (if not all) of them will stop working.

Idk, the gradle from here is from 2019, I think we must update to something new tough.

MAJigsaw77:
From what I've saw, from all their `build.gradle` files must replace `compile` option with `api` or `implementation`, that's all.

flashultra:
I think there was also a dependency from AndroidX libraries and maybe something else.  I'm not against such update, but I'm just warning about possible problems in the future with extensions . If anyone want could make a check .

MAJigsaw77:
> I think there was also a dependency from AndroidX libraries and maybe something else. I'm not against such update, but I'm just warning about possible problems in the future with extensions . If anyone want could make a check .

I used `AndroidX` with this gradle, there was no ploblems.

player-03:
Plus AndroidX is related to the target SDK version, not the Gradle version. If you're still targeting an old SDK, then you don't need it.

MAJigsaw77:
I still think we must update `Gradle`, the current version of gradle that lime uses is from 2019, it's pearty old.

player-03:
I knew this felt familiar! Closing this because we already have #1500.

MAJigsaw77:
Um?

MAJigsaw77:
Nvm

PR #1601, opened 2022-11-18
Use macOS 11 for testing.

player-03:
GitHub has officially begun to [drop macOS 10.15 support](https://github.com/actions/runner-images/issues/5583), causing workflows to fail.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,7 +76,7 @@ jobs:
           if-no-files-found: error
 
   macos-ndll:
-    runs-on: macos-10.15
+    runs-on: macos-11
     steps:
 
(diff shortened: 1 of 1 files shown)


joshtynjala:
Should we consider macos-11 instead instead of macos-12, to help ensure that our NDLL binaries will remain compatible with the oldest possible macOS version? I recall that's why we were chose macos-10.15 instead of going the easier route with macos-latest.

player-03:
Guess I'm out of the loop on this. Sure, we can do that.

player-03:
Wait, didn't they officially drop support at the beginning of the month? How have workflows been working all this time? In any case, it's time to merge.

PR #1604, opened 2022-11-25
Use `getDisplayMetrics()` to find screen DPI on Android.

MAJigsaw77:
Title

### DIFF
--- a/templates/android/template/app/src/main/java/org/haxe/lime/GameActivity.java
+++ b/templates/android/template/app/src/main/java/org/haxe/lime/GameActivity.java
@@ -38,7 +38,11 @@ public class GameActivity extends SDLActivity {
 		if (metrics == null) {
 
 			metrics = new DisplayMetrics ();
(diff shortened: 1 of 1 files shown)


player-03:
I'd say it's safe to use `getRealMetrics()` without checking. `JELLY_BEAN_MR1` corresponds to SDK 17, and as of #1519, Lime is no longer compatible with SDKs below 21. Plus, Google dropped support for that version in September 2021.

MAJigsaw77:
A ok

MAJigsaw77:
@player-03 done

MAJigsaw77:
I thought this won't be pushed, that's why I closed it .___.

Dimensionscape:
I haven't looked at it personally yet, but sometimes it just takes us forever to completely vet some prs. If you're getting feedback, its typically a good sign.

MAJigsaw77:
Ok

PR #1615, opened 2023-01-05
Allow using Eval to run Lime tools.

### LINKED ISSUE #1307: alloc.c(177) : max_string_size reached

player-03:
Similar to #1614, but using [a different target](https://haxe.org/blog/eval/). This version leaves Neko as the default, but adds Eval as an option. To enable, pass `-eval` as an argument.

Pros:

- Eval fixes several of Neko's quirks. For instance, it should resolve #1307.
- There's no need to run `lime rebuild tools` after each change.

Cons:

- There's a ~2 second pause every time Eval starts up. (I included a message right before the pause to make it easier to measure.)
- CFFI is disabled, which might make some features not work. (But do the tools really use audio and fonts and so on?)
- If someone was already using the `-eval` flag for their own purposes, this will shadow that. Seems unlikely, given that the usual convention is to use two dashes for any flag longer than a letter.


player-03:
@AdamHarte Give this one a try, see if it fixes your issue.

PR #1623, opened 2023-01-21
Assets open from wherever

Shallowmallow:
Before if you did 
./build/openfl/linux/bin/App

it would do
Error : [lime.utils.Preloader] ERROR: There is no asset library with an ID of "default"

### DIFF
--- a/src/lime/utils/AssetManifest.hx
+++ b/src/lime/utils/AssetManifest.hx
@@ -167,6 +167,8 @@ class AssetManifest
 		{
 			basePath = basePath.substr(0, basePath.length - 1);
 		}
+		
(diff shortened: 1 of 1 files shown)


player-03:
This is at least partially a duplicate of #1538. Please rewrite based on the [8.1.0-Dev](https://github.com/openfl/lime/tree/8.1.0-Dev) branch.

Shallowmallow:
Oh cool, your version seems better :)

player-03:
Can you confirm it works for your use case?

Shallowmallow:
Yes, just tested your commit works for me :)

PR #1625, opened 2023-02-05
Fix null pointer error on RPi.

player-03:
The error might not even be RPi-specific. As far as I can tell, the issue is that due to [`@:generic`](https://haxe.org/manual/type-system-generic.html), `Promise<Dynamic>` is its own class, distinct from whatever type got created earlier, so we can't cast between the two. And static targets have some kind of enforcement mechanism, so when we try to store the other promise as a `Promise<Dynamic>`, it just stores null.

I don't think the `@:generic` metadata is needed, but I'm not about to remove it without knowing why it was added. So instead, I chose to store the specific functions we need, since those shouldn't have any trouble with casting.

### DIFF
--- a/src/lime/app/Future.hx
+++ b/src/lime/app/Future.hx
@@ -346,8 +346,10 @@ import lime.utils.Log;
 	private static var singleThreadPool:ThreadPool;
 	#if lime_threads
 	private static var multiThreadPool:ThreadPool;
-	// It isn't safe to pass a promise object to a web worker.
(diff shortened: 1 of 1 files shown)


player-03:
@gepatto, can you confirm this works?

gepatto:
Yes it works on the Pi. 
No more need to ignore the @:generic meta in Promise.hx now.

player-03:
Perfect! Let's go with this, then.

PR #1626, opened 2023-02-06
Allow Assets.registerLibrary() for default library

dpomier:


player-03:
Couldn't you just pass the string "default"?

dpomier:
The only motivation is consistency, "" is handled as an alias for "default" in other functions (`unloadLibrary`, `getLibrary`, etc.). It eventually handles `Assets.registerLibrary("", Assets.loadLibrary(""))`, but has a low chance to happen in real code.

PR #1633, opened 2023-02-15
Functionality for Future Implementations

Just-Feeshy:
An essential step towards ensuring that a software project remains flexible and adaptable to changing requirements. By adding this functionality, developers can build a solid foundation for future implementations, saving time and resources in the long run. I have a framework that I'm working on called [spoopy](https://github.com/Project-Ever-Changing/spoopy), which is based on OpenFL but uses both Vulkan and Metal instead. I would love to see and help Lime implement Vulkan since OpenGL is deprecated, and Vulkan could offer more optimization to achieve optimal performance for Haxe games.


Dimensionscape:
Hey! We definitely welcome efforts to future proof the frameworks and tools! If you'd like to join us on the official discord channel, I'd love to talk about the situation. A native Vulkan renderer would be really interesting, although we potentially have options like Googles ANGLE to translate GLSL calls to Metal or Vulkan :D

https://github.com/google/angle

Just-Feeshy:
Sure, I would love to join the discord! We indeed have to rely on either Googles ANGLE, GLSANG, or SPIRV-Cross to translate GLSL shaders.

PR #1635, opened 2023-02-17
undefine legacy pi driver

gepatto:
The legacy pi driver is preventing a successful rebuild of lime for the raspberry pi running a recent piOS (bullseye)
because it will try to link to files that are no longer present on bullseye 
_This driver wasn't very good anyway because it didn't have proper x11 support, wasn't able to set display resolutions and mouse support was buggy_

As pi support is very new I don't think backwards compatibility is an issue here.


player-03:
Huh, looks like [you defined it in the first place](https://github.com/openfl/libsdl/commit/581f53d0c77d67e915a9fc344630559da115b93d#diff-2130a90366e832f38c21d938e2b6bc75ef6e603e733c54bc4fa1bcc98228c82bR291). You've been working on this for a while!

player-03:
I don't know why that convention exists, but I do prefer consistency. Just checking: does this still work?

gepatto:
I have been trying to have support both the legacy and the new KMSDRM driver.
My main piOS has been updated through versions. 
Therefore some files existed on my OS that are missing from a clean OS.
So my builds WITH the old driver were still successful.

When I tried out a clean piOS install I noticed the files that are needed for the legacy driver are no longer there
and compiling lime would fail.  ( the old driver uses DISPMANX )

As for the #undef
It might be ok to use the   */ #undef    style  for consistency
I'll try that first.

Ao

gepatto:
I'm closing this pull request, I'm doing some more research first

PR #1636, opened 2023-02-17
Pass `-rpi` flag to svg.n.

player-03:
This hopefully fixes the issue @gepatto reported on Discord:

> so in the chat channel we discovered that when you build html5 on a Pi, 
> SVGExport fails because it runs neko in a regular linux context in stead of rpi .
> SVGExport is looking for a -rpi flag but that flag isn't appended when ImageHelper.hx calls 'neko process svg.n'


player-03:
> This hopefully fixes the issue

Conclusion: no it doesn't.

PR #1639, opened 2023-02-25
fix wierd audio stuttering on windows

Vortex2Oblivion:
after updating to lime 8.0.1 yesterday, i was experiencing weird audio issues where it would stutter for a tiny bit at random points, reverting  #1558  *seems* to fix it
After some testing with this change I haven't really had any issues

### DIFF
--- a/src/lime/_internal/backend/native/NativeAudioSource.hx
+++ b/src/lime/_internal/backend/native/NativeAudioSource.hx
@@ -186,6 +186,8 @@ class NativeAudioSource
 		else
 		{
 			var time = completed ? 0 : getCurrentTime();
+			
+			AL.sourcePlay(handle);
 
 			setCurrentTime(time);
 		}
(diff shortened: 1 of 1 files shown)


player-03:
@justin-espedal Can you take a look at this?

justin-espedal:
This isn't exactly helpful if it's just going back to the previous behavior, which was also buggy.

The only platform the previous behavior was definitively wrong for was iOS. If somebody can't figure out exactly what's going wrong here, perhaps some platform-specific code could be used as a crutch?

What platform are you experiencing issues on, @Vortex2Oblivion?

Vortex2Oblivion:
im on windows c++

player-03:
Did you mean to close this?

player-03:
@Vortex2Oblivion Should we assume there's no more need to revert #1558?

PR #1641, opened 2023-02-28
remove define for legacy SDL_VIDEO_DRIVER_RPI

gepatto:
I did some extra research.

The files that are needed for compiling the legacy SDL_VIDEO_DRIVER_RPI
are indeed no longer present in the current piOS (raspbian bullseye). 
So we need to comment out this define to be able to rebuild lime.
I stuck to the proper syntax style this time.

So the default will be compiling the X11 and the KMSDRM-driver for raspberry pi for running in x11 and commandline respectively.

_I didn't catch this earlier because my own Pi was upgraded from a previous os version, so I still had the dispmanx related files that are needed for the legacy driver.
A clean recent version of piOS, which most people will be using, doesn't have the files ._


player-03 (approved):

gepatto:
I see CI failing on hashlink-samples and neko-samples on ubuntu but I don’t see how it’s related to the change I made to a Pi specific file.

joshtynjala:
The failure did not seem related to your changes. I guess someone re-ran the job, and it worked.

Looks good to me.

Issue #1640, opened 2023-02-27
Building lime.ndll on Apple Silicon Mac


tobil4sk:
Created this to keep track of the status of native Apple Silicon builds of the lime.ndll.

Todo:

- [x] SDL submodule fix (see https://github.com/openfl/libsdl/commit/fa51fcb56b152ba6419b6decfef5dad48b075ac2). `develop` branch is missing this patch, but this is applied in `Dev-8.2.0`, in f200ecd9cae9ed45a1f8c23f3229fac74a608d9e
- [x] #1642
  This PR includes the following submodule patches, for the Dev-8.2.0 branch:
  - https://github.com/openfl/libpixman/pull/2
  - https://github.com/openfl/libopenal/pull/1
- [x] Hxcpp patch is also required for the ndll name to be correct: https://github.com/HaxeFoundation/hxcpp/pull/1088 
- [x] Update lime to load the ndll from the correct path: #1753
  - [x] Update swf to load the ndll from the correct path: https://github.com/openfl/swf/pull/30 
- [x] #1752
- [ ] Add `/opt/homebrew/lib` as an rpath for arm64 mac builds of lime.ndll: https://github.com/openfl/lime/issues/1750

After applying these patches, it is possible to build the ndll and build/run projects as expected.

tobil4sk:
Created a new branch based on 8.2.0 as suggested. The linker error was due to a problem with the libpixman patch, which I have explained in more detail here: https://github.com/openfl/libpixman/pull/1. 

I've created a new PR with the 8.2.0 based branch, which allows for a successful build of the lime.ndll on arm64 Mac with `haxelib run lime rebuild mac -DHXCPP_ARM64`: #1642. This also requires the most recent development version of hxcpp. I've pasted the installation instructions in case anyone is interested:

<details>
<summary>Expand for detailed installation instructions</summary>

```sh
# install updated hxcpp with arm64 building support
haxelib git hxcpp https://github.com/HaxeFoundation/hxcpp.git

# clone forked version of lime from PR #1642
git clone https://github.com/soccertutor/lime.git --branch fix/m1-build-ndll --recursive

# set lime library path
haxelib dev lime ./lime

# rebuild lime
haxelib run lime rebuild mac -DHXCPP_ARM64
```
</details>

Since the ndll is still loaded from the `Mac64` path for now, we had to move the new arm64 binary:

```sh
cd $(haxelib libpath lime)
rm ndll/Mac64/lime.ndll
[...]

PR #1642, opened 2023-03-01
Fix lime.ndll build for Apple Silicon

tobil4sk:
Applies the patch from https://github.com/openfl/libopenal/pull/1, along with some other fixes to get the openal config up to date with the version used in `8.2.0-Dev`. Also applies fixes for pixman build.

These patches allow for successful building of the lime ndll on arm64 Mac systems.

See #1640.

### DIFF
--- a/project/lib/custom/openal/include/config-macos-arm64.h
+++ b/project/lib/custom/openal/include/config-macos-arm64.h
@@ -14,6 +14,9 @@
 /* Define a built-in call indicating an aligned data pointer */
 #define ASSUME_ALIGNED(x, y) x
 
+/* Define a restrict macro for non-aliased pointers */
+#define RESTRICT __restrict
+
 /* Define if HRTF data is embedded in the library */
 #define ALSOFT_EMBED_HRTF_DATA
 
--- a/project/lib/custom/openal/include/config-macos-arm64.h
+++ b/project/lib/custom/openal/include/config-macos-arm64.h
@@ -2,26 +2,11 @@
 #define AL_API  __attribute__((visibility("default")))
 #define ALC_API __attribute__((visibility("default")))
 
-#ifdef IN_IDE_PARSER
-/* KDevelop's parser doesn't recognize the C99-standard restrict keyword, but
- * recent versions (at least 4.5.1) do recognize GCC's __restrict. */
-#define restrict __restrict
-#endif
-
- /* Define any available alignment declaration */
-#define ALIGN(x) __attribute__((aligned(x)))
-
-/* Define a built-in call indicating an aligned data pointer */
-#define ASSUME_ALIGNED(x, y) x
-
 /* Define a restrict macro for non-aliased pointers */
 #define RESTRICT __restrict
 
 /* Define if HRTF data is embedded in the library */
 #define ALSOFT_EMBED_HRTF_DATA
-
-/* Define if we have the sysconf function */
-/* #undef HAVE_SYSCONF */
 
 /* Define if we have the C11 aligned_alloc function */
 /* #undef HAVE_ALIGNED_ALLOC */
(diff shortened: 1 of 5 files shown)


tobil4sk:
On Apple Silicon it still builds the x86_64 ndll by default unless `-DHXCPP_ARM64` is specified. Perhaps something needs to be updated in: https://github.com/openfl/lime/blob/8.2.0-Dev/project/Build.xml

player-03:
That's odd. I _thought_ that was set in the platform class, but [`MacPlatform` only sets one architecture at a time](https://github.com/openfl/lime/blob/8.2.0-Dev/tools/platforms/MacPlatform.hx#L96-L108).

joshtynjala:
MacPlatform seems like the right place to me. What are you referring to when you mention multiple architectures? "Universal" apps that support both Apple Silicon and Intel?

joshtynjala:
I was looking at LinuxPlatform a bit today in relation to Raspberry Pi stuff. I realized that hxp's `System.hostArchitecture` doesn't have an ARM64 enum value yet. That's going to need to be updated for Lime tools to be able to detect that architecture.

tobil4sk:
With 55ca39da44a9f399b8aeff6d947be1e732ebd875, it would be good to double check that the pixman/openal changes here won't break anything for raspberry pi.

> I realized that hxp's System.hostArchitecture doesn't have an ARM64 enum value yet

Created an issue here to keep track of the hxp changes: https://github.com/openfl/hxp/issues/29

joshtynjala:
Sorry for the noise. I didn’t see any discussion about multiple ndlls being built. I guess I mistakenly thought the issue was that it built the wrong architecture by default, but still built only a single ndll.

tobil4sk:
>  I guess I mistakenly thought the issue was that it built the wrong architecture by default, but still built only a single ndll.

This is correct, by default it built `Mac64`. `MacArm64` was only built with the `-DHXCPP_ARM64` flag set.

tobil4sk:
> I realized that hxp's System.hostArchitecture doesn't have an ARM64 enum value yet. That's going to need to be updated for Lime tools to be able to detect that architecture.

I've opened a PR to add this: https://github.com/openfl/hxp/pull/30. I assume we'll have to wait for a new hxp release to make use of this?

joshtynjala:
> I assume we'll have to wait for a new hxp release to make use of this?

Yeah.

jgranick:
iOS, Android and other platform targets use multiple architectures in a single build. If this is needed or desired for future macOS builds, that functionality could easily be copied and modified from those targets.

JonnycatMeow:
i tried to build the arm64 ndll for macos but it gave me some errors 
Screen Shot 2023-07-24 at 7 04 19 AM

JonnycatMeow:
and i used the fix/m1-build-ndll

JonnycatMeow:
and my machine is a x86_64 but i have xcode

joshtynjala:
> Inline variable initialization must be a constant value

I recall that this one is caused by using the no-inline option with Haxe 4.3. You may need to temporarily downgrade to Haxe 4.2. I don't think our Haxe 4.3 fixes have been merged into the 8.2.0-Dev branch yet, which is what the PR is based on.

JonnycatMeow:
oh ok

tobil4sk:
>  I don't think our Haxe 4.3 fixes have been merged into the 8.2.0-Dev branch yet

Actually, it looks like 8.2.0-Dev does have most of these patches now but my branch was still out of date. I've updated it now, hopefully this should solve those 4.3 errors.

JonnycatMeow:
it builds good now thank you

JonnycatMeow:
nvm got another error. 
Screen Shot 2023-07-26 at 3 28 25 PM

JonnycatMeow:
i used the branch fix/m1-build-dll

tobil4sk:
> nvm got another error.

@JonnycatMeow Did you manage to resolve this? It's possible that your submodules were not updated after checking out to this branch. The location of the header files is different in the version of pixman used in this branch (because it is based on 8.2.0-Dev rather than develop).

See: https://github.com/openfl/lime/blob/8.2.0-Dev/project/README.md#build-troubleshooting

tobil4sk:
This PR just fixes compilation issues with the submodules, the rest of the apple silicon related changes are handled in #1642. This won't affect rosetta. Would it be possible to merge this?

PR #1643, opened 2023-03-02
Improve `doWork`'s backwards compatibility.

player-03:
This provides the no-op function expected by [`FileStream.__disposeFileStreamWorker()`](https://github.com/openfl/openfl/blob/9.2.1/src/openfl/filesystem/FileStream.hx#L300), and potentially others.


joshtynjala:
Thanks for restoring the public API for backwards compatibility!

player-03:
Just to be certain, I went through every public field in `BackgroundWorker`. Fields are identical except where noted.

- `public var canceled(default, null):Bool;` now has a default value.
- `public var completed(default, null):Bool;` now has a default value.
- `public var doWork = new Event<Dynamic->Void>();` is now `(get, never)` and has type `PseudoEvent`. `PseudoEvent` closely matches `Event`'s public API, except `__listeners` and `__repeat` are both `(get, never)`.
- `public var onComplete = new Event<Dynamic->Void>();` is now `(default, null)`.
- `public var onError = new Event<Dynamic->Void>();` is now `(default, null)`.
- `public var onProgress = new Event<Dynamic->Void>();` is now `(default, null)`.
- `public function new() {}` now has optional arguments.
- `public function cancel():Void` now has an optional argument.
- `public function run(message:Dynamic = null):Void` now has two optional arguments, with `message` being second. The first argument is a very specific type that did not previously exist, so Haxe will normally be able to tell that the user's message is meant as the second. The only failure case I can think of is if the user passes a `Dynamic` value.
- `public function sendComplete(message:Dynamic = null):Void` now has a second optional argument.
- `public function sendError(message:Dynamic = null):Void` now has a second optional argument.
- `public function sendProgress(message:Dynamic = null):Void` now has a second optional argument.

player-03:
> Thanks for restoring the public API for backwards compatibility!

I'm pretty sure there was an old commit with the full API too, but I removed it before submitting the PR. Since I couldn't find it quickly, I decided it was quicker to rewrite it. Plus side, I think the new approach is cleaner.

player-03:
As long as I'm here, I might as well go through `ThreadPool`'s old fields.

- `public var currentThreads(default, null):Int;` is now `(get, never)`.
- `public var doWork = new Event<Dynamic->Void>();` is now (get, never) and has type PseudoEvent. PseudoEvent closely matches Event's public API, except __listeners and __repeat are both (get, never).
- `public var maxThreads:Int;` is unchanged.
- `public var minThreads:Int;` is unchanged.
- `public var onComplete = new Event<Dynamic->Void>();` is now `(default, null)`.
- `public var onError = new Event<Dynamic->Void>();` is now `(default, null)`.
- `public var onProgress = new Event<Dynamic->Void>();` is now `(default, null)`.
- `public var onRun = new Event<Dynamic->Void>();` is now `(default, null)`.
- `public function new(minThreads:Int = 0, maxThreads:Int = 1)` now has more optional arguments.
- `public function queue(state:Dynamic = null):Void` now has two optional arguments, with `state` being second. The first argument is a very specific type that did not previously exist, so Haxe will normally be able to tell that the user's state is meant as the second. The only failure case I can think of is if the user passes a `Dynamic` value.
- `public function sendComplete(message:Dynamic = null):Void` now has a second optional argument.
- `public function sendError(message:Dynamic = null):Void` now has a second optional argument.
- `public function sendProgress(message:Dynamic = null):Void` now has a second optional argument.

Yep, pretty much exactly the same.

Issue #1009, opened 2017-07-21
Bug: Looping audio will "hiccup" and restart twice, at slightly different positions


jaredsain:
Digging into  _backend/native/NativeAudioSource.hx it looks like there's an issue with getting the correct AL.BYTE_OFFSET, perhaps due to a race condition. On looped audio this is the series of function calls I'm interested in:

```
private function timer_onRun ():Void {
	playing = false;
	if (loops > 0) {
		loops--;
		setCurrentTime (0);
		play ();
		return;	
	} 
}
```

The play() function also calls getCurrentTime and setCurrentTime:

```
public function play ():Void {
		
	
	playing = true;
		
	if (stream) {
			
	...
			
	} else {
		var time = getCurrentTime ();
			
		AL.sourcePlay (handle);
		setCurrentTime (time);
			
	}
		
}
```
So here's what it *looks* like is happening (to me):
In timer_onRun(), setCurrentTime(0) is called before play(), and it is supposed to set the playhead information to 0. To do that, it calls
`AL.sourcei (handle, AL.BYTE_OFFSET, totalOffset);`
and totalOffset is at value 0.

Then, back in timer_onRun, play() is called, and we get to this line:
`var time = getCurrentTime ();`
which *should* return 0, since we just set the current time manually to 0. In getCurrentTime the return value is calculated based on this variable:
`var offset :Int = cast AL.getSourcei (handle, AL.BYTE_OFFSET);`
That's the offset we just set to 0! Except here the value is NOT 0, it is a completely different value, which I believe is related to the length of the audio file being played. So, we start playing the audio file again, using AL.sourcePlay(handle), which is unaffected by the offset value.

However, we then call 
`setCurrentTime(time)`
And "time" SHOULD be 0, but it is a very different value. And in setCurrentTime, the "playing" flag is now true, so we run the piece of code that creates the loop timer, but it is given an incorrect time value, so the timer is very short, because it thinks the audio file is about to reach its end, even though it just started over. A frame or two later we get a second timer_onRun() call, there is an audible glitchy sound when the audio file restarts for a second time. This time, though, the offset is retrieved correctly at value 0, and the audio plays unobstructed until its next loop.

[...]

PR #1646, opened 2023-03-09
Tweak audio behavior on iOS and Windows.

### LINKED ISSUE #1638: audio issues in lime 8.0.1
I've been experiencing an issue where audio will cut out for like half a second and then come back on with like a static noise in between
i downgraded my lime to 8.0.0 and it seemed to fix the issue.
anyone else having this problem?

player-03:
#1558 made some changes to fix an audio issue on iOS. However, in #1638, @Vortex2Oblivion reported that this caused a new error on Windows. This pull request simply reverts #1558 on non-iOS platforms. 

This isn't a perfect solution, but it should work until someone figures out the root cause of one or both errors.

### DIFF
--- a/src/lime/_internal/backend/native/NativeAudioSource.hx
+++ b/src/lime/_internal/backend/native/NativeAudioSource.hx
@@ -417,7 +417,7 @@ class NativeAudioSource
 			else if (parent.buffer != null)
 			{
 				AL.sourceRewind(handle);
-				#if !ios if (playing) AL.sourcePlay(handle) #end
+				#if !ios if (playing) AL.sourcePlay(handle); #end
 				// AL.sourcef (handle, AL.SEC_OFFSET, (value + parent.offset) / 1000);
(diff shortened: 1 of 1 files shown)


player-03:
I wonder if #1009 is related to all this, but am not prepared to spend the time to figure it out right now.

player-03:
I wonder, does anyone else experience audio issues (static, cutting out, etc.) when targeting Windows C++? So far only one person reported this behavior, and he silently withdrew the issue.

joshtynjala:
I keep meaning to try to reproduce the original issue and test the PR, but I haven't had the time yet.

joshtynjala:
I tried various ogg and wav files that I had laying around with Windows c++ today, but I could not reproduce the issue.

player-03:
The original reporter long since closed their issue, and no one else reported it after nine months, so I think it should be good as-is.

PR #1654, opened 2023-04-01
Use forward slashes to load lime.ndll.

player-03:
`substr(7)` returns the substring _starting_ at character 7, which will never be "windows". The author meant to type `substr(0, 7)` instead, to get the substring ending at 7.

We could easily make this change, but given that the code has been successfully using forward slashes for several years, I prefer to simplify.

### DIFF
--- a/src/lime/system/CFFI.hx
+++ b/src/lime/system/CFFI.hx
@@ -149,16 +149,15 @@ class CFFI
 
 			if (result == null)
 			{
-				var slash = (__sysName().substr(7).toLowerCase() == "windows") ? "\\" : "/";
 				var haxelib = __findHaxelib("lime");
 
 				if (haxelib != "")
 				{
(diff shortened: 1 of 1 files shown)


player-03:
Actually I'm not sure if this piece of code is ever run. It's the "else" case at the end of a long if-else chain, so I think it could only come up for users who made their own DLLs.

Also, testing reveals that `findHaxelib()` returns a string with forward slashes, even on Windows, so we're getting forward slashes no matter what.

Also also, after delving into how Neko loads DLLs on Windows, I'm not sure the passed string is actually used. That sounds wrong, but Neko only passes the string to `dlopen()`, which doesn't exist on Windows. Windows uses `GetModuleHandle()` instead, and Neko only ever calls `GetModuleHandle(NULL)`.

It's all a big mess, and I barely understand any of it. But one thing I do understand is that `substr(7)` is wrong.

PR #1659, opened 2023-04-10
Test using NDK r21e.

player-03:
hxcpp 4.3.2 introduced an error when using GCC to compile projects. By switching to an NDK >= 20, we can make it use Clang instead, and hopefully avoid the error.

We might want to do this anyway, since NDK 15 is so old.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -215,7 +215,7 @@ jobs:
         uses: nttld/setup-ndk@v1
         id: setup-ndk
         with:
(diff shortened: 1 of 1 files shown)


player-03:
Oh whoops, this requires updates from 8.2.0-Dev, including 4d096d03e22c9377dfa8cef13d39af37071123e6 at minimum.

player-03:
...And of course 8.2.0-Dev is already using r21e. Guess there isn't much to do here.

PR #1661, opened 2023-04-12
Allow cross compiling Windows projects with MinGW

tobil4sk:
If the `-mingw` flag is set when compiling the `windows` target from another host OS, it will now use MinGW instead of neko.

### DIFF
--- a/src/lime/system/System.hx
+++ b/src/lime/system/System.hx
@@ -32,9 +32,9 @@ import sys.io.Process;
 @:access(lime._internal.backend.native.NativeCFFI)
 @:access(lime.system.Display)
 @:access(lime.system.DisplayMode)
-#if (cpp && windows && !HXCPP_MINGW && !lime_disable_gpu_hint)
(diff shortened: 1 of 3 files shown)


DigiEggz:
I've tested this and it's working well.

tobil4sk:
@player-03 Thanks for merging! As we discussed in #1610, we will also have to clean up the logic for deciding which target to use for a cross compilation build (and make sure everything works properly in non xml projects).

EDIT: I see you have made a commit to address this issue: 6725825469b8cb3113090daeb837f7a72bdc6d9e. Thanks!

ahmutkadim:
Hey, is this pull request merged with a main release yet? If not when it will be?

joshtynjala:
@ahmutkadim This PR was merged into the 8.2.0-Dev branch. Lime 8.2 won't be released until some point after that branch has been merged into the develop branch. It's coming soon-ish, I think, but it'll still be a bit of time while we work to stabilize that release.

PR #1663, opened 2023-04-17
Fixed JS Image issue when using the genes library to compile ES modules

andresa88:
Need to explicitly use:

new window.Image() 

otherwise it will create a lime Image object.

### DIFF
--- a/src/lime/_internal/backend/html5/HTML5HTTPRequest.hx
+++ b/src/lime/_internal/backend/html5/HTML5HTTPRequest.hx
@@ -444,11 +444,7 @@ class HTML5HTTPRequest
 
 	private static function __loadImage(uri:String, promise:Promise<Image>, options:Int):Void
 	{
-		#if (openfljs || genes)
-		var image:JSImage = untyped #if haxe4 js.Syntax.code #else __js__ #end ('new window.Image ()');
-		#else
-		var image = new JSImage();
-		#end
+		var image:JSImage = #if haxe4 js.Syntax.code #else __js__ #end ('new window.Image ()');
 
 		if (!__isSameOrigin(uri))
 		{
--- a/src/lime/_internal/backend/html5/HTML5HTTPRequest.hx
+++ b/src/lime/_internal/backend/html5/HTML5HTTPRequest.hx
@@ -444,7 +444,11 @@ class HTML5HTTPRequest
 
 	private static function __loadImage(uri:String, promise:Promise<Image>, options:Int):Void
 	{
+		#if (openfljs || genes)
+		var image:JSImage = untyped #if haxe4 js.Syntax.code #else __js__ #end ('new window.Image ()');
+		#else
 		var image = new JSImage();
+		#end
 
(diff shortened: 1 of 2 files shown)


joshtynjala:
Are you sure that this change is necessary? The original code is this:

```haxe
var image = new JSImage();
```

And `JSImage` is explicitly set as an alias of `js.html.Image` in the imports.

```haxe
import js.html.Image as JSImage;
```

It should be impossible for the generated JS to use `lime.graphics.Image` in this context instead. There would be a conflict between `js.html.Image` and `lime.graphics.Image` here only if the constructor call was `new Image()` instead of `new JSImage()`.

This seems more like a bug in genes.

player-03:
I think Andres might be right. Looking at the output, [`new js.html.Image()` in Haxe becomes `new Image()` in JavaScript](https://try.haxe.org/#EC27F340). Normally, this creates a standard image element.

```haxe
class Test {
  static function main() {
    trace(new js.html.Image()); //[object HTMLImageElement]
  }
}
```

But as soon as we define our own `Image` class, [it shadows the standard one](https://try.haxe.org/#23232f8b).

```haxe
class Test {
  static function main() {
    trace(new js.html.Image()); //[not HTMLImageElement]
  }
}

@:keep
class Image {
  public inline function new() {
  }
  
  public inline function toString():String {
		return "[not HTMLImageElement]";
  }
}
```

There might still be a better workaround. And if there is, we should apply that in the [two other locations](https://github.com/openfl/lime/commit/d806b27b1b0189bc4467915c453cf0ca60c870b0) this code is used.

joshtynjala:
Oh yes, I see. It slipped by mind that the fully-qualified `js.html.Image` wouldn't persist in JS. If there's another unqualified `Image` symbol in scope (which there seems to be, since `lime.graphics.Image` is imported), it will shadow the native JS API, as you said.

This seems like the correct solution to me.

player-03:
Is this correct, though? `HTML5HTTPRequest` always imports `lime.graphics.Image`, so why do we call `new JSImage()` under any circumstances? Why not always call `new window.JSImage()`?

player-03:
Also, I don't think classes are scoped in the JS output. Pretty sure that if `lime.graphics.Image` is compiled at all (i.e., not removed by DCE), it's available at the top level.

joshtynjala:
If I remember correctly, `new JSImage()` works correctly with the Haxe compiler's default JS output because it adds more qualification than the alternative JS output generated by genes (and probably hxgenjs too). I think the Haxe compiler generates an identifier for `lime.graphics.Image`, like `lime_graphics_Image`, so it doesn't conflict.

player-03:
Oh right, it probably does. So as long as you provide a package (unlike try.haxe.org), you can't shadow the type.

player-03:
> I'd still like to get rid of the conditional compilation.

Done. (Took a couple tries to get it right...)

joshtynjala:
Looks good to me!

andresa88:
Thank you both! The additional commits work for me.

PR #1664, opened 2023-04-17
Temporarily revert to hxcpp 4.2.1.

player-03:
hxcpp 4.3.2 breaks CI builds. Until the fix makes its way to Haxelib, the easiest workaround is just to use an older version.


player-03:
Oh wait, Josh had the exact same idea an hour earlier.

Issue #1622, opened 2023-01-20
Invalid unicode string from file dialogue on non-Windows system


tobil4sk:
When using the save file dialogue on non-Windows systems, any non-ASCII characters in the file name result in an invalid string. This is because the string is returned as a `wstring`: https://github.com/openfl/lime/blob/develop/project/src/ExternalInterface.cpp#L902

While this works fine on Windows, this is problematic on Linux where wstring is not standard. As a result, if the string ends up being passed to hxcpp's system functions, it is considered invalid and we end up with confusing error messages (like `Allocating from a GC-free thread` which comes from a failed allocation of an exception string to describe this issue).

It's possible to workaround this issue by adding an extra conversion back to a regular string: https://github.com/tobil4sk/lime/tree/fix/unix-unicode-file

However, it would probably be better to refactor the api a bit to remove the need for wstring on non Windows systems altogether, because right now there are redundant conversions between string types just to fit [this api](https://github.com/openfl/lime/blob/d75b9f9cfaafbca58cafadeaf91475ea36362c8a/project/include/ui/FileDialog.h#L16-L19):

Conversion to wstring:
https://github.com/openfl/lime/blob/d75b9f9cfaafbca58cafadeaf91475ea36362c8a/project/src/ExternalInterface.cpp#L748-L750
Conversion from wstring to string:
https://github.com/openfl/lime/blob/d75b9f9cfaafbca58cafadeaf91475ea36362c8a/project/src/ui/FileDialog.cpp#L108-L110

joshtynjala:
It looks like the 8.2.0-Dev branch has updated tinyfiledialogs to a newer version, and tinyfiledialogs has switched the default char encoding on Windows to UTF-8, instead of MBCS. If I'm understanding things correctly, it means that we shouldn't need to use the std::wstring APIs on Windows anymore, and Lime's FileDialog can just use std::string everywhere.

joshtynjala:
Interestingly, with the following code and Lime 8.0.1, I tried opening a file named _åéîøü.txt_, and it outputs an incorrect string **on Windows!**,

```haxe
var dialog = new FileDialog();
dialog.onSelect.add(value -> {
	trace("select file: " + value);
});
dialog.browse(OPEN);
```

> src/Main.hx:26: open file: C:\Users\josht\Desktop\├Ñ├⌐├«├╕├╝.txt

It seems like, if we were using `std::wstring` to support Unicode on Windows, it was never actually working, or it broke at some point.

[...]

PR #1665, opened 2023-04-18
Remove redundant file dialog string conversions

tobil4sk:
Currently, there are a lot of redundant string conversions when using the file dialog api. This was mentioned briefly in #1622.

Here is a rough overview of the current situation:

Linux/macOS:
- C++: `hxstring` -> `std::wstring` -> `std::string`, passed into tinyfiledialogs utf-8 api, then `std::string` -> `std::wstring` -> `hxstring`
- HL: `hl_vstring` -> `std::wstring` -> `std::string`, passed into tinyfiledialogs utf-8 api, then `std::string` -> `std::wstring` -> `utf-8` bytes -> native `utf-16` hashlink string (on Haxe side via `String.fromUtf8()`)

Windows:
- C++: `hxstring` -> `std::wstring`, passed into tinyfiledialogs utf-16 api, then `std::wstring` -> `hxstring`
- HL: `hl_vstring` -> `std::wstring`, passed into tinyfiledialogs utf-16 api, then `std::wstring` -> `utf-8` bytes -> native `utf-16` hashlink string (on Haxe side via `String.fromUtf8()`)

This cleans things up to avoid unnecessary conversions, so now it looks like this in most cases:

Linux/macOS:
- C++: `hxstring` -> `utf-8`[0], passed into tinyfiledialogs utf-8 api, then `utf-8` -> `hxstring`
- HL: `hl_vstring` -> `utf-8`, passed into tinyfiledialogs utf-8 api, then `utf-8` -> `utf-16` hashlink string (on Haxe side via `String.fromUtf8()`)

Windows:
- C++: `hxstring` -> `utf-16`[0], passed into tinyfiledialogs utf-16 api, then `utf-16` -> `hxstring`
- HL: `hl_vstring`, passed into tinyfiledialogs utf-16 api, then `utf-16` -> `utf8` -> `utf-16` hashlink string (on Haxe side via `String.fromUtf8()`)

[0] hxcpp can use ASCII (which is compatible with utf-8) or utf-16 depending on the string, so these conversions are avoided in some cases.

We can improve Hashlink further (for Windows), by returning utf-16 strings from the apis, however, I'm not sure if this would be considered a breaking change (new lime.hdlls would be incompatible with old lime haxe files).

In #1622, we briefly discussed using the utf-8 tinyfiledialog functions on Windows to unify the api, however, I looked into this and on Windows these functions just convert back to utf-16, so there would still be extra conversions. So I think this is the cleanest solution.

I've tested on Linux and Windows with the code from #1622, and everything works well.

### DIFF
--- a/project/include/ui/FileDialog.h
+++ b/project/include/ui/FileDialog.h
@@ -13,10 +13,30 @@ namespace lime {
 
 		public:
 
-			static std::wstring* OpenDirectory (std::wstring* title = 0, std::wstring* filter = 0, std::wstring* defaultPath = 0);
-			static std::wstring* OpenFile (std::wstring* title = 0, std::wstring* filter = 0, std::wstring* defaultPath = 0);
-			static void OpenFiles (std::vector<std::wstring*>* files, std::wstring* title = 0, std::wstring* filter = 0, std::wstring* defaultPath = 0);
-			static std::wstring* SaveFile (std::wstring* title = 0, std::wstring* filter = 0, std::wstring* defaultPath = 0);
+#ifdef HX_WINDOWS
+			using char_t = wchar_t;
+#else
+			using char_t = char;
+#endif
+
+			static const char_t* OpenDirectory(const char_t* title = nullptr, const char_t* filter = nullptr, const char_t* defaultPath = nullptr);
+			static const char_t* OpenFile(const char_t* title = nullptr, const char_t* filter = nullptr, const char_t* defaultPath = nullptr);
+
+			// Use actual string view once we can use c++17
+			class string_view {
+			public:
+				string_view(const char_t* c_str, size_t length) : c_str(c_str), length(length) {}
+
+				const char_t* data() const { return c_str; }
+				const size_t size() const { return length; }
+
+			private:
+				const char_t* c_str;
+				size_t length;
+			};
+
+			static std::vector<string_view> OpenFiles(const char_t* title = nullptr, const char_t* filter = nullptr, const char_t* defaultPath = nullptr);
+			static const char_t* SaveFile(const char_t* title = nullptr, const char_t* filter = nullptr, const char_t* defaultPath = nullptr);
 
 	};
 
--- a/project/include/ui/FileDialog.h
+++ b/project/include/ui/FileDialog.h
@@ -30,7 +30,7 @@ namespace lime {
 				string_view(const char_t* c_str, size_t length) : c_str(c_str), length(length) {}
 
 				const char_t* data() const { return c_str; }
-				const size_t size() const { return length; }
+				size_t size() const { return length; }
 
 			private:
 				const char_t* c_str;
--- a/project/include/ui/FileDialog.h
+++ b/project/include/ui/FileDialog.h
@@ -9,6 +9,8 @@
 namespace lime {
 
 
+	// Functions in this class return pointers to externally owned memory, which will be reused automatically
+	// for future dialogs and should not be freed.
 	class FileDialog {
 
 		public:
--- a/project/src/ui/FileDialog.cpp
+++ b/project/src/ui/FileDialog.cpp
@@ -6,297 +6,143 @@
 
 #include <tinyfiledialogs.h>
 
+#ifdef HX_WINDOWS
+#define tinyfd_selectFolderDialog tinyfd_selectFolderDialogW
+#define tinyfd_openFileDialog tinyfd_openFileDialogW
+#define tinyfd_saveFileDialog tinyfd_saveFileDialogW
+
+#define FD_STR(str) L##str
+#else
+#define FD_STR(str) str
+#endif
+
+using char_t = lime::FileDialog::char_t;
+using string_t = std::basic_string<char_t>;
 
 namespace lime {
 
-
-	std::string* wstring_to_string (std::wstring* source) {
-
-		if (!source) return NULL;
-
-		int size = std::wcslen (source->c_str ());
-		char* temp = (char*)malloc (size + 1);
-		std::wcstombs (temp, source->c_str (), size);
-		temp[size] = '\0';
-
-		std::string* data = new std::string (temp);
-		free (temp);
-		return data;
-
+	static bool is_null_or_empty (const char_t* str) {
+		return str == nullptr || str[0] == 0;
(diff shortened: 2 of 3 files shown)


nixbody:
Just a little correction here, HXCPP (with smart strings enabled) never uses UTF-8 for strings. It uses ASCII and when any character in a string doesn't fit within ASCII range then the entire string is converted to UTF-16. HXCPP, however, provides function to convert it's strings to UTF-8.

tobil4sk:
Yes, that's correct. However, an ASCII string is valid utf-8, so when using hxs_utf8 on an hxcpp string which is ASCII encoded, it just returns a pointer to the ASCII string without any conversion being necessary.

nixbody:
I was wondering if you point that out :) I thought you might know, my apologies for pointing out the obvious.

tobil4sk:
No worries, it's worth making the distinction :)

project/src/ui/FileDialog.cpp
```diff
@@ -9,6 +9,11 @@
 
 namespace lime {
 
+	template<typename T>
+	static bool is_empty (const T* str) {
+		return str[0] == 0;
+	}
+
```
nixbody, line 16: Wouldn't `std::wcslen()` did pretty much the same thing?

nixbody: Never mind, I get it.

nixbody (approved):

nixbody (approved):

project/src/ui/FileDialog.cpp
nixbody: Ok, I apologise for my previous useless comments but you still probably check if `str` isn't `nullptr` before dereferencing it.

tobil4sk: I moved the null check into the function in the latest commit.

nixbody (approved):

nixbody (approved):

tobil4sk:
There are a few more places where wstrings are converted to utf8 still. These include most of the system functions, though these are windows only, where wchar is equal to char16_t so `hl_to_utf8`/`alloc_wstring` can be used there. For `System::GetDirectory`, it can be modified to avoid wstrings.

The bigger issue is `Font::GetFamilyName()` in `text/Font.cpp`. Currently it assumes that wchar is 16 bits, which is only the case on Windows. Even if it used char16_t, the `alloc_hxs_utf16()` function returns an HxString, and I can't seem to figure out to turn that into a `value`.

project/include/ui/FileDialog.h
```diff
@@ -13,10 +13,23 @@ namespace lime {
...
+			static const wchar_t* SaveFile (const wchar_t* title = 0, const wchar_t* filter = 0, const wchar_t* defaultPath = 0);
+		#else
+			// Path separator used by OpenFiles
+			static const char PATH_SEPARATOR = '|';
+
+			static const char* OpenDirectory (const char* title = 0, const char* filter = 0, const char* defaultPath = 0);
+			static const char* OpenFile (const char* title = 0, const char* filter = 0, const char* defaultPath = 0);
+			static const char* OpenFiles (const char* title = 0, const char* filter = 0, const char* defaultPath = 0);
+			static const char* SaveFile (const char* title = 0, const char* filter = 0, const char* defaultPath = 0);
+		#endif
```
player-03: On the one hand, I'm pretty sure this is what typedefs are for. It would be really nice not to have this kind of duplicate code.

On the other hand, if we aren't careful, new typedefs would just mean new instances where conditional compilation is needed. For instance, a type that mapped to `wchar_t` on Windows and `char` elsewhere doesn't help when Windows code needs to use `char`, so we'd still need `#ifdef HX_WINDOWS`.

tobil4sk: I've updated the PR with a cleaner version using typedefs, which avoids having multiple declarations/definitions of the functions

tobil4sk:
I fixed conflicts in the PR and updated it to clean things up and reduce duplicate code. I've tested on linux and windows and it works as expected.

Aside from cleaning things up and simplicity, it is also a good idea to make this change because the `std::wstring` currently being created on non-windows systems are actually utf8 with widened characters, which is a bit confusing and doesn't play nicely with apis (e.g. hxcpp's `alloc_wstring`). This confusion has caused issues in the past like #1622.

player-03 (approved):

player-03:
I was going to ask about memory leaks, since some `delete` statements were removed. But then I tracked everything down in _tinyfiledialogs.c_ and found that it only returns static variables. If it's called again, it reuses the variable, overwriting the old value. And that's fine because we copy the value into a Haxe string before calling again. As far as I can tell, everything works as it should.

Not that I don't trust you to check, I just wanted to understand it myself. Plus, I learned some basics of gdb, which is nice.

tobil4sk:
> I was going to ask about memory leaks, since some delete statements were removed. But then I tracked everything down in tinyfiledialogs.c and found that it only returns static variables. 

This is something I found confusing too so perhaps a comment might be useful. I can't remember if tfd has some documentation explaining this or if i just had to check the source?

Also, I suppose the static string might also not be thread safe, though that is regardless of this PR. Though, given it is file dialogs it might be unlikely to cause issues in practice.

> Not that I don't trust you to check, I just wanted to understand it myself.

I agree, we want code that can be easily verified/understood by anyone reading it, rather than just assuming it works based on trust in one contributor.

player-03:
> I can't remember if tfd has some documentation explaining this or if i just had to check the source?

I checked the source when I wrote that, but it turns out they also [mention it in the readme](https://sourceforge.net/p/tinyfiledialogs/code/ci/3f79fd513b8eaa760424ce959699431a792eac9a/tree/README.txt#l168): "String memory is preallocated statically for all the returned values."

> Also, I suppose the static string might also not be thread safe, though that is regardless of this PR. Though, given it is file dialogs it might be unlikely to cause issues in practice.

Agreed on all counts. I doubt most OSes allow opening dialogs from non-UI threads, but if they did, both threads would read/write that same memory.

tobil4sk:
> This is something I found confusing too so perhaps a comment might be useful.

I added a comment in the `ui/FileDialog.h` header, does this seem ok to you now? @player-03

PR #1680, opened 2023-05-07
Fix `WebGL2RenderContext` formatting.

player-03:
Apparently the formatter doesn't account for conditional compilation, so when it sees a block of code like this...

```haxe
#if !lime_webgl
public function bufferData(target:Int, srcData:ArrayBufferView, usage:Int, srcOffset:Int = 0, length:Int = 0):Void
{
#else
public inline function bufferData(target:Int, srcData:Dynamic, usage:Int, ?srcOffset:Int, ?length:Int):Void
{
#end
```

...it indents once per curly brace. So instead of duplicating these curly braces, we should put them after `#end`.

### DIFF
--- a/src/lime/graphics/WebGL2RenderContext.hx
+++ b/src/lime/graphics/WebGL2RenderContext.hx
@@ -3772,79 +3772,79 @@ abstract WebGL2RenderContext(Dynamic) from Dynamic to Dynamic
 		}
 
 	#if !lime_webgl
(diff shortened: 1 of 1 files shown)


player-03:
Yes, this does have to be a bunch of separate commits. Otherwise, Git would show that 1245 lines were removed and replaced by 1245 new lines, not recognizing that any of those lines match, and it would be much harder to verify.

I'm not saying the "files changed" tab shows a very good diff, but it's a lot better than if I'd submitted this PR as a single commit. At least the lines _almost_ still line up, rather than appearing hundreds of lines down.

To see good-quality diffs, look at the individual commits.

PR #1683, opened 2023-05-12
Removed Joystick trackball support.

EliteMasterEric:
This API is only supported on Linux, with a single piece of hardware. They are set to be removed with SDL3 (see libsdl-org/SDL#6766).

They should be removed from Lime as well, since no application was actually using them (or could have, since the relevant hardware is no longer available).


project/include/ui/JoystickEvent.h
```diff
@@ -13,7 +13,7 @@ namespace lime {
 
 		JOYSTICK_AXIS_MOVE,
 		JOYSTICK_HAT_MOVE,
-		JOYSTICK_TRACKBALL_MOVE,
+		// JOYSTICK_TRACKBALL_MOVE,
```
player-03: Any reason for commenting this, rather than deleting?

src/lime/_internal/backend/native/NativeApplication.hx
```diff
@@ -746,7 +742,7 @@ class NativeApplication
 {
 	var AXIS_MOVE = 0;
 	var HAT_MOVE = 1;
-	var TRACKBALL_MOVE = 2;
+	// var TRACKBALL_MOVE = 2;
```
player-03: Also this.

project/include/ui/JoystickEvent.h
EliteMasterEric: Fixed by setting the enum values to constants so existing applications don't break.

src/lime/_internal/backend/native/NativeApplication.hx
EliteMasterEric: Since these values are already pegged to constants, there is no reason to keep this here. I made a commit to remove it.

player-03: I guess having a comment could explain why 2 is skipped, but there doesn't need to be. Skipping numbers is normal in enums, and usually unexplained.

PR #1684, opened 2023-05-12
Migrate backend from SDL 2 to SDL 3

EliteMasterEric:
This pull request performs the following:

- Updates the project from SDL 2 to the latest version of SDL 3.
    - This lays the groundwork for later commits which provide new features based on these changes.
- Fixed a rendering issue caused by changes in how SDL 3.0 handles high-DPI displays.
- Window frame now displays in either black or white, depending on the user's system theme (dark/light).
    - Functionality to configure this, or retrieve the current system theme, are planned for later PRs.

PR #1688, opened 2023-05-15
Implemented precise key events

EliteMasterEric:
This feature PR adds two new key press event, one for pressing a key and one for releasing a key. The distinction is that the precise key press events include the SDL timestamp of the keypress. By calling `SDL_GetTicks()` (which this PR also includes an interface for), you can calculate the number of milliseconds since the key was pressed. This is useful for mitigating input latency.

Notes:

- The previous PR (#1561) is now deprecated and has been closed.
- This PR creates a new, otherwise-redundant key press event to prevent breaking changes from modifying the existing key press event.
- The event has been made to be compatible with both Native and HTML5.
- The timestamp provided by the keypress event is only suitable for comparision, and is not an absolute/Unix timestamp.
  - This behavior probably needs to change somehow, for example making a Timestamp object with various `compare` functions available.
  - On native, compare to `lime._internal.backend.native.NativeCFFI.lime_sdl_get_ticks()`.
  - On HTML5, compare to `js.Browser.window.performance.now()`.
- The old key press event can be deprecated and eventually removed in a future update.
- Once #1684 is tested, debugged, and implemented, the event can be updated to use nanosecond timings instead.


EliteMasterEric:
Since I made this PR I also had to tweak these code changes to make a similar addition to gamepads.

[That commit is public](https://github.com/openfl/lime/commit/f2f038ff0dda7bf7ba3bac277c44f0f5d7a2f98e).

EliteMasterEric:
Redone in #1802 to include the gamepad stuff and exclude extra commits.

PR #1698, opened 2023-07-10
Add `vsync` property to windows for native targets

Starmapo:
Currently, setting a window's VSync value is only possible from the Project.xml file. This pull request adds the `vsync` property to `lime.ui.Window`, which allows for getting and setting the VSync value during runtime. Useful if you want to let the user decide if they want VSync enabled.

This is only implemented for native targets, and doesn't work for Flash or HTML5 applications, though I don't know if they support VSync.


dimensionscape:
Which platforms have you tested this on?

Starmapo:
> Which platforms have you tested this on?

Windows.

dimensionscape:
Thank you. This is a nice addition.

Sword352 (approved):

SomeGuyWhoLovesCoding:
> Currently, setting a window's VSync value is only possible from the Project.xml file. This pull request adds the `vsync` property to `lime.ui.Window`, which allows for getting and setting the VSync value during runtime. Useful if you want to let the user decide if they want VSync enabled.
> 
> This is only implemented for native targets, and doesn't work for Flash or HTML5 applications, though I don't know if they support VSync.

HTML5 has vsync enabled by dwfault anyway, and for flash, hmmmmm... We don't know yet

dimensionscape:
https://github.com/openfl/lime/commit/a5d2a49dc1ff8279be31689e2fc94ddfe09acc17

PR #1702, opened 2023-08-01
Fix warnings in Haxe 4.3.

player-03:
Looks like Haxe is dropping support for `@:enum abstract`, and we'll have to use `enum abstract` instead. (Same with `@:extern` → `extern`.)

For now, all it does is generate warnings, but it can generate them after build-stopping errors, making it hard to find what went wrong.

### DIFF
--- a/src/lime/_internal/backend/native/NativeApplication.hx
+++ b/src/lime/_internal/backend/native/NativeApplication.hx
@@ -637,7 +637,7 @@ class NativeApplication
 	}
 }
 
-@:enum private abstract ApplicationEventType(Int)
+#if haxe4 private enum #else @:enum private #end abstract ApplicationEventType(Int)
 {
 	var UPDATE = 0;
 	var EXIT = 1;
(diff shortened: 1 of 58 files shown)


joshtynjala:
These warnings should be fixed in develop/8.1.0-Dev already. It should just need a merge from 8.1.0-Dev.

player-03:
Oh ok. I didn't check for that.

player-03:
I do still like `#if haxe4` as shorthand for `#if (haxe_ver >= 4.0)`, but whatever.

(Yes, `#if haxe4` will remain true in versions greater than 4. After all, `#if haxe3` is still true.)

joshtynjala:
Yeah, I understand. I think I made some of the Flash externs use `if (haxe_ver >= 4.3)` so that some of the changes wouldn't be disruptive for those who haven't upgraded to 4.3 yet. So I probably just copied that and changed to `4.0` instead of switching to `#if haxe4`.

PR #1716, opened 2023-10-10
Sound Length Fix Part 2

### LINKED ISSUE #2666: Sounds over a certain length do not play properly.
I originally posted this on the HaxeFlixel repository, but it turned out to be an issue with OpenFL and lime.

Sounds over a certain length, seemingly around 25 minutes, do not play properly. An acquaintance discovered the details regarding the reason for this, and how to fix it:

image

Read below for additional details:
https://github.com/HaxeFlixel/flixel/issues/2899

Jaduplex:
Fixes https://github.com/openfl/openfl/issues/2666. Must be pulled along with the pull request to OpenFL, as the fix only works when made in both places.

The OpenFL pull request: https://github.com/openfl/openfl/pull/2667

### DIFF
--- a/src/lime/_internal/backend/native/NativeAudioSource.hx
+++ b/src/lime/_internal/backend/native/NativeAudioSource.hx
@@ -136,7 +136,7 @@ class NativeAudioSource
 			}
 		}
 
-		samples = Std.int((dataLength * 8) / (parent.buffer.channels * parent.buffer.bitsPerSample));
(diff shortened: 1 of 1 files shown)


Jaduplex:
I don't really wanna deal with this. If there's any objections or changes to this, I'm sure there's some other way to fix this somewhere. I just wanted to throw this out randomly because it's a fix I found useful, but apparently it's not as helpful as I first thought.

PR #1720, opened 2023-10-22
Fix incorrectly escaped file path on Windows.

player-03:
Specifically, `C:` becomes `C\:`, which means Gradle can no longer find the NDK. Presumably escaping any other colon is appropriate, but absolutely not that one.


player-03:
Issue was fixed by #1721.

PR #1721, opened 2023-10-24
Only escape backslashes in Android paths.

player-03:
Neither [.properties](https://en.wikipedia.org/wiki/.properties#Format) nor [.gradle](https://groovy-lang.org/syntax.html#_escaping_special_characters) files require anything else to be escaped, and in .gradle files, escaping anything else is incorrect.

I think most of these weren't well-tested in the first place. On Linux* I was able to use the ` #!=` characters without any trouble at all, while colons caused hxcpp to be unable to find the g++ executable. (For the latter, it makes no difference whether we do or don't escape it, because hxcpp doesn't get the file path from Lime.)

*And it's clear this wasn't well-tested on Windows either, since it escapes the colon in "C:\".

Supersedes #1720, or vice versa.

### DIFF
--- a/tools/platforms/AndroidPlatform.hx
+++ b/tools/platforms/AndroidPlatform.hx
@@ -490,9 +490,8 @@ class AndroidPlatform extends PlatformTarget
 			context.ANDROID_BUILD_TOOLS_VERSION = AndroidHelper.getBuildToolsVersion(project);
 		}
 
-		var escaped = ~/([ #!=\\:])/g;
-		context.ANDROID_SDK_ESCAPED = escaped.replace(context.ENV_ANDROID_SDK, "\\$1");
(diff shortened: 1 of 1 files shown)


player-03:
So, funny story. I went to track down whoever added that character-escaping code and ask why it was there.

...

...Turns out? [That was me.](https://github.com/openfl/lime/commit/22d07a5c759808b3905e68a8e87a464fce05b220)

(Makes sense in retrospect. I might be the only Lime dev who regularly uses regexes.)

I don't recall doing it, but after reviewing some documentation for .properties files, it's clear I was trying to escape characters with special meanings in .properties files. Of the six characters in the regex, two indicate comments, three are used as delimiters, and backslash escapes characters. However, comments only work at the beginning of the line, and only the first delimiter on a line has any effect, so we never needed to escape those characters. And I guess .properties files ignore extraneous backslashes, so it still worked even though it wasn't correct.

PR #1724, opened 2023-11-20
Added `android:appCategory`.

MAJigsaw77:
This pr is basically making an app able to be detected as a game on `android`.

### DIFF
--- a/templates/android/template/app/src/main/AndroidManifest.xml
+++ b/templates/android/template/app/src/main/AndroidManifest.xml
@@ -7,7 +7,7 @@
 	::if (ANDROID_PERMISSIONS != null)::::foreach ANDROID_PERMISSIONS::<uses-permission android:name="::__current__::" />
 	::end::::end::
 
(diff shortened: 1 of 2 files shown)


MAJigsaw77:
Is there anything wrong with the pr?

player-03:
The implementation looks fine, but it seems [`isGame` is deprecated](https://developer.android.com/guide/topics/manifest/application-element#isGame).

MAJigsaw77:
> The implementation looks fine, but it seems [`isGame` is deprecated](https://developer.android.com/guide/topics/manifest/application-element#isGame).

Maybe i should implement that instead

MAJigsaw77:
@player-03 what should be as default value?

joshtynjala:
> what should be as default value?

The Android docs say this:

> Only define this value for apps that fit well into one of the specific categories.

With that in mind, by default, you shouldn't set it to any value. Just skip it entirely.

MAJigsaw77:
Is ok?

MAJigsaw77:
Will this be pushed?

player-03:
What do you think of #1735 as a solution?

MAJigsaw77:
> What do you think of #1735 as a solution?

It's better then this implementation

MAJigsaw77:
You can close this if you want

player-03:
Might leave this open until we get more feedback on my version. Edit: merged.

PR #1726, opened 2023-11-30
[HTML5] Preloader does not reflect progress of a packed asset library

scanline:
When using:
`<library name="default" type="zip" />`

to wrap individual assets like images into a single **default.pak** file, the preloader
(indirectly) does not take the size of the file into account. So if we have a packed library of e.g.
1MB and no other assets, the preloader bar jumps directly from 0 to 100% as soon as the file
finished loading.

This issue has been mentioned years ago in this thread: [[Preloader Progress Doesn’t Update When Using “default” Library](https://community.openfl.org/t/preloader-progress-doesnt-update-when-using-default-library/10688)](https://community.openfl.org/t/preloader-progress-doesnt-update-when-using-default-library/10688/)

The findings by singmajesty did not get rid of the issue though. The problem is that
**PackedAssetLibrary.hx** does not monitor the loading of default.pak itself.

This pull request will take care of it.

By the way - HTML5 is surely not the only affected target but it's where I've noticed it.

### DIFF
--- a/src/lime/utils/PackedAssetLibrary.hx
+++ b/src/lime/utils/PackedAssetLibrary.hx
@@ -168,6 +168,7 @@ import flash.media.Sound;
 		if (promise == null)
 		{
 			promise = new Promise<AssetLibrary>();
+			bytesLoadedCache = new Map();
 
 			// TODO: Handle `preload` for individual assets
 			// TODO: Do not preload bytes on native, if we can read from it instead (all non-Android targets?)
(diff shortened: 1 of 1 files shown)


justin-espedal:
Looks like this tackles the same problem as https://github.com/openfl/lime/pull/1542, right?

I missed the request there to rebase on the latest development branch, but if desired I can still do that. If this PR is gone with on the other hand, feel free to close that one. :)

PR #1728, opened 2023-11-30
Prevents stuck if we don't have access to gamepad

### LINKED ISSUE #1727: Failed to execute 'getGamepads' on 'Navigator': Access to the feature "gamepad" is disallowed by permissions policy.
Latest Chrome Instant games gives issue:
`Failed to execute 'getGamepads' on 'Navigator': Access to the feature "gamepad" is disallowed by permissions policy.`

Possible fix in `src/lime/ui/Joystick.hx`
```
	#if (js && html5)
	@:noCompletion private static function __getDeviceData():Array<Dynamic>
	{
		var res:Dynamic = null;
		
		try 
		{
			res = (untyped navigator.getGamepads) ? untyped navigator.getGamepads() : (untyped navigator.webkitGetGamepads) ? untyped navigator.webkitGetGamepads() : null;
		}
		catch (err:Dynamic)
		{
			trace("Gamepad access issue");
		}
		
		return res;
	}
	#end
```

Or maybe also add new flag to prevent call `updateGameDevices()` each frame in `src/lime/_internal/backend/html5/HTML5Application.hx`
if we don't need this feature?

andrew-git:
Original issue: Failed to execute 'getGamepads' on 'Navigator': Access to the feature "gamepad" is disallowed by permissions policy. #1727

### DIFF
--- a/src/lime/ui/Joystick.hx
+++ b/src/lime/ui/Joystick.hx
@@ -63,7 +63,8 @@ class Joystick
 		}
 		catch (err:Dynamic)
 		{
-			trace("Gamepad access issue");
+			// if something went wrong, treat it the same as when navigator.getGamepads doesn't exist
+			// we probably don't have permission to use this feature
 		}
 		
 		return res;
--- a/src/lime/ui/Joystick.hx
+++ b/src/lime/ui/Joystick.hx
@@ -55,8 +55,18 @@ class Joystick
 	#if (js && html5)
 	@:noCompletion private static function __getDeviceData():Array<Dynamic>
 	{
-		return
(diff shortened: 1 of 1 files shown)


andrew-git:
```
// if something went wrong, treat it the same as when navigator.getGamepads doesn't exist
// we probably don't have permission to use this feature
```
But we anyway will continue to try get gamepads each frame
`https://github.com/openfl/lime/blob/develop/src/lime/_internal/backend/html5/HTML5Application.hx#L362`

@joshtynjala Maybe add additional fixes or add flag to disable calls of `updateGameDevices()` if we don't need gamepads/joysticks in game?

joshtynjala:
> But we anyway will continue to try get gamepads each frame

This should be fine. It should behave the same as an older browser where navigator.getGamepads doesn't exist: updateGameDevices() will see null and return early.

PR #1734, opened 2023-12-20
Remove armv7 from default iOS target architectures

### LINKED ISSUE #1733: `lime rebuild ios` fails on M1 due to multiple architecture flags
By default, when running `lime rebuild ios`, both armv7 and arm64 architecture flags are active. See: https://github.com/openfl/lime/blob/e6205bf3aaace23c7e2458841da4435ff9879eaf/tools/platforms/IOSPlatform.hx#L478-L482

This means that `HXCPP_ARMV7` and `HXCPP_ARM64` are set at once:
https://github.com/openfl/lime/blob/e6205bf3aaace23c7e2458841da4435ff9879eaf/tools/platforms/IOSPlatform.hx#L491-L493

The hxcpp code does not allow for these flags to be set at once (see https://github.com/HaxeFoundation/hxcpp/blob/18f87b949464ddcf24c6a83a2f7b8f76233540bc/toolchain/iphoneos-toolchain.xml#L43-L47).

The compiler receives `-arch armv7 arm64`, so only armv7 is treated as an architecture flag, and the `arm64` is treated as a file to be compiled, resulting in the error:
```
Error: clang: error: no such file or directory: 'arm64'
```

Perhaps it's fine to just no longer set `HXCPP_ARMV7` by default? Considering the commit message here: https://github.com/openfl/lime/commit/f34b595e70785758d3517030466a954af17027bb.

tobil4sk:
~~Fixes~~ #1733. Nevermind.

~~Avoids the error `no such file or directory: 'arm64'` due to multiple architectures being passed in.~~

### DIFF
--- a/tools/platforms/IOSPlatform.hx
+++ b/tools/platforms/IOSPlatform.hx
@@ -475,8 +475,7 @@ class IOSPlatform extends PlatformTarget
 	public override function rebuild():Void
 	{
 		var armv6 = (project.architectures.indexOf(Architecture.ARMV6) > -1 && !project.targetFlags.exists("simulator"));
-		var armv7 = (command == "rebuild"
-			|| (project.architectures.indexOf(Architecture.ARMV7) > -1 && !project.targetFlags.exists("simulator")));
+		var armv7 = (project.architectures.indexOf(Architecture.ARMV7) > -1 && !project.targetFlags.exists("simulator"));
 		var armv7s = (project.architectures.indexOf(Architecture.ARMV7S) > -1 && !project.targetFlags.exists("simulator"));
 		var arm64 = (command == "rebuild"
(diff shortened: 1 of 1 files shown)


player-03:
I mean, we could remove it from the defaults for other reasons, but that's unrelated to the rebuild error.

tobil4sk:
The original error was caused by a [bug in hxcpp](https://github.com/HaxeFoundation/hxcpp/pull/1079), but maybe it makes sense to remove armv7 from the defaults anyway:
https://community.openfl.org/t/ios-armv7-architecture-is-deprecated/13698

player-03:
That thread seems to be talking about the `build` step, not `rebuild`, and I think they already removed it by default for that step. So rebuilding for armv7 isn't breaking anything, it just performs a bunch of unnecessary work.

If you undelete the branch, I can reopen the pull request.

tobil4sk:
The branch is restored.

PR #1735, opened 2023-12-22
Make it easier to configure AndroidManifest.xml from project.xml.

player-03:
These changes allow the user to configure their [`<application />`](https://developer.android.com/guide/topics/manifest/application-element) and [`<activity />`](https://developer.android.com/guide/topics/manifest/activity-element) tags without ever leaving project.xml.

```xml
<config:android>
	<application android:appCategory="game" android:supportsRtl="true" />
	<activity android:immersive="false" />
	<activity android:persistableMode="persistAcrossReboots" />
</config:android>

<config:android>
	<!-- If an attribute is repeated, it will overwrite the old value. -->
	<activity android:persistableMode="persistNever" />
	
	<!-- Passing an empty string removes that attribute. -->
	<application android:appCategory="" />
	
	<!-- It's even possible to override Lime's defaults. -->
	<application android:theme="@android:style/Theme.Fullscreen" />
</config:android>
```

Previously, the user would have needed to make a template and then maintain that template indefinitely. Now, they can change the one thing they need and leave the maintenance to Lime.


ghost (approved):

player-03:
Note to self: remember to [add documentation](https://github.com/openfl/lime.openfl.org/blob/master/_docs/project-files/xml-format.md) when this gets closer to release.

Update: draft pull request created.

PR #1740, opened 2024-01-03
force disable code signing in iOS (probably needed better coding)

ghost:
due I'm thinking about what if there's people using code signing??

if u guys want me change smth u can add me in discord (cuz I can't sometimes see gh notifications and username mc.9)

### DIFF
--- a/src/lime/tools/IOSHelper.hx
+++ b/src/lime/tools/IOSHelper.hx
@@ -29,7 +29,12 @@ class IOSHelper
 		}
 		else
 		{
-			commands.push("build");
+			commands.push("clean");
+                        commands.push("build");
+                        commands.push("CODE_SIGN_IDENTITY=\"\"");
+                        commands.push("CODE_SIGNING_REQUIRED=\"NO\"");
+                        commands.push("CODE_SIGN_ENTITLEMENTS=\"\"");
+                        commands.push("CODE_SIGNING_ALLOWED=\"NO\"");
 		}
 
 		if (additionalArguments != null)


KarimAkra (approved):

player-03:
What if someone wants to sign their code [using `<certificate team-id="..." />` in project.xml](https://lime.openfl.org/docs/project-files/xml-format/#certificate)? This has been in place for at least 6 years.

I'd say try project.xml first, and if you're certain it doesn't work for you, [IOSPlatform.hx](https://github.com/openfl/lime/blob/f34b595e70785758d3517030466a954af17027bb/tools/platforms/IOSPlatform.hx#L128-L131) is the correct place to make the change.

ghost:
> What if someone wants to sign their code [using `<certificate team-id="..." />` in project.xml](https://lime.openfl.org/docs/project-files/xml-format/#certificate)? This has been in place for at least 6 years.
> 
> I'd say try project.xml first, and if you're certain it doesn't work for you, [IOSPlatform.hx](https://github.com/openfl/lime/blob/f34b595e70785758d3517030466a954af17027bb/tools/platforms/IOSPlatform.hx#L128-L131) is the correct place to make the change.

that's what I meant to say before but making them empty (in Project.xml) doesn't do anything, many users can't or don't know how to get DUNS number and gives "code signing required by (app name)" at actions

player-03:
> making them empty (in Project.xml) doesn't do anything

I doubt an empty string is a valid team ID, so I suppose we could skip `IOSHelper.sign()` in that case.

> many users can't or don't know how to get DUNS number

...But you don't need a DUNS number, you need an Apple-assigned certificate ID. And last I checked, Apple requires all developers to get one, no matter what kind of app they're making.

Unless you're saying there's a GitHub Action that accesses your Apple account and does that for you?

ghost:
> ...But you don't need a DUNS number, you need an Apple-assigned certificate ID. And last I checked, Apple requires all developers to get one, no matter what kind of app they're making.

I know and tried in actions with my account didn't work (doesn't have any DUNS number)

> Unless you're saying there's a GitHub Action that accesses your Apple account and does that for you?

[There's something here for make code signing](https://docs.github.com/en/actions/deployment/deploying-xcode-applications/installing-an-apple-certificate-on-macos-runners-for-xcode-development) but many users I talk says "I don't trust" idk really

player-03:
> [There's something here for make code signing](https://docs.github.com/en/actions/deployment/deploying-xcode-applications/installing-an-apple-certificate-on-macos-runners-for-xcode-development) but many users I talk says "I don't trust" idk really

Most of that looks correct, though I'm a bit wary of how they're using environment variables. Any program you ran afterwards would have access to the secrets, though presumably you'd run only trusted ones. Maybe it's fine?

But yeah, you're going to have to do something along those lines no matter what. And once you have the keystore set up, you ought to be able to use Lime's built-in code signing.

PR #1743, opened 2024-01-12
Optimize CI workflow.

player-03:
The `-eval` option means most of the CI tasks no longer have to wait for all the others. If we run more of them in parallel, users will be able to get much quicker feedback on their pull requests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -746,37 +746,39 @@ jobs:
           lime build SimpleAudio html5 -clean -release -verbose -nocolor -minify -terser -eval
 
   neko-samples:
-    needs: linux
-    runs-on: ${{ env.LINUX_VERSION }}
+    needs: package-haxelib
     strategy:
       matrix:
         haxe-version: [3.4.7, 4.0.5, 4.1.5, 4.2.5, 4.3.1]
+        os: [${{ env.WINDOWS_VERSION }}, ${{ env.LINUX_VERSION }}, ${{ env.MACOS_VERSION }}]
+    runs-on: ${{ matrix.os }}
     steps:
-
-      - uses: actions/checkout@v3
 
       - uses: krdlab/setup-haxe@v1
         with:
           haxe-version: ${{ matrix.haxe-version }}
 
-      - name: Set HAXEPATH
+      - name: Set HAXEPATH (Windows)
+        if: runner.os == 'Windows'
+        run: |
+          echo "HAXEPATH=$Env:HAXE_STD_PATH\.." >> $Env:GITHUB_ENV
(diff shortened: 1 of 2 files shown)


player-03:
At this point, it's possible to complete everything else while waiting for the `brew bundle` step of macos-ndll.

Tasks

player-03:
Credit to @Apprentice-Alchemist for figuring out why Homebrew was so slow.

player-03:
Workflow succeeded in 25 minutes, compared to ~1h45m before this PR.

joshtynjala:
Great to see the CI finishing faster! I'm concerned about one thing that you changed, though. The removal of this dependency from the samples builds:

```yml
needs: package-haxelib
```

I made the samples builds rely on package-haxelib on purpose. The idea was to ensure that the Haxelib .zip bundle that we're generating is valid, and can build projects for all supported targets.

player-03:
I see where you're coming from. On the one hand, with the brew issue fixed (for now), going back to the package-haxelib bottleneck shouldn't slow things down _too_ much. On the other hand, it has been packaging correctly for years, it would throw an error if any of the pieces were missing, and we're testing the pieces individually. I don't see how it could start going wrong now.

How about this as a compromise? We test each target as soon as the required ndlls become available, saving time. Later when package-haxelib finishes, we test its output by downloading the bundle on Windows/Mac/Linux, and building a sample in HashLink. (HashLink because all we're trying to do is prove that the bundled version of Lime works.)

joshtynjala:
> How about this as a compromise? We test each target as soon as the required ndlls become available, saving time. Later when package-haxelib finishes, we test its output by downloading the bundle on Windows/Mac/Linux, and building a sample in HashLink. (HashLink because all we're trying to do is prove that the bundled version of Lime works.)

Works for me!

player-03:
Ok, the variables gave me a fair bit of trouble, and I think the 25-minute run might have been a fluke, but it's working.

Workflow

player-03:
Hmm, some of the Neko runs take a while to download the artifact. Maybe there are just too many in parallel. Could scale it back, I guess.

Workflow

player-03:
Ugh, reducing the number of jobs didn't even help. All of the time saved was because the iOS builds happened to go a bit faster.

player-03:
As an experiment, let's try going back to hard-coded `runs-on` values. It's possible GitHub optimizes based on knowing in advance which machines to use, and that's how we got that 25-minute run.

player-03:
It's definitely faster with hard-coded machines to run on. Still not 25 minutes, but I think that was only possible because we didn't test lime-haxelib.zip. @joshtynjala, does this look good to merge?

joshtynjala (approved):

PR #1745, opened 2024-01-16
Don't compile `armv7` if the `architecture` excluded it.

MAJigsaw77:
Previously only the ndll was excluded from being added to the app, now that part of the compilation won't even happen anymore decreasing the compile time.

### DIFF
--- a/tools/platforms/AndroidPlatform.hx
+++ b/tools/platforms/AndroidPlatform.hx
@@ -145,7 +145,7 @@ class AndroidPlatform extends PlatformTarget
 		var architectures = [];
 
 		if (hasARMV5) architectures.push(Architecture.ARMV5);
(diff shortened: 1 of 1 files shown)


ghost (commented):
hopefully that does better job in ndk r26+

player-03:
Well now there's no guarantee that any architecture will be selected. Maybe we need something like this at the end?

```haxe
if (architectures.length == 0) architectures.push(Architecture.ARM64);
```

MAJigsaw77:
> Well now there's no guarantee that any architecture will be selected. Maybe we need something like this at the end?
> 
> ```hx
> if (architectures.length == 0) architectures.push(Architecture.ARM64);
> ```

Right

MAJigsaw77:
Maybe was better to also warn the user but ok

player-03:
That would be fine too. I don't have any reasons to prefer either.

MAJigsaw77:
I think is done

PR #1749, opened 2024-01-25
Update howler.min.js

ninjamuffin99:
simply updates howler to the latest version

lime currently uses 2.1.1 from 2018, which may seem old, but it's juuust one minor version away, and the ONE single breaking change (`xhrWithCredentials` replaced) doesn't seem to be touched in lime at all 

https://github.com/goldfire/howler.js/blob/master/CHANGELOG.md#breaking-changes
https://github.com/goldfire/howler.js/blob/master/CHANGELOG.md

aka... free fixes and profit

nabbed howler.min.js from the dist/ folder on github, just since i was browsing it! https://github.com/goldfire/howler.js/tree/master/dist


EliteMasterEric:
Free fixes AND free profit? Sign me up

PR #1753, opened 2024-01-27
Use correct ndll loader path on mac arm64

tobil4sk:
Also requires the same patch in swf: https://github.com/openfl/swf/pull/30


tools/platforms/MacPlatform.hx
```diff
@@ -535,6 +535,6 @@ class MacPlatform extends PlatformTarget
 
 	private inline function get_dirSuffix():String
 	{
-		return targetArchitecture == X64 ? "64" : "";
+		return targetArchitecture == X64 ? "64" : targetArchitecture == ARM64 ? "Arm64" : "";
```
tobil4sk, line 538: The ndll directory should be `MacArm64`. However, I'm not sure if this makes sense everywhere where `dirSuffix` is used, for example, the build directory would be `macArm64`, which looks odd with the upper case letter.

player-03: Addressed in 4ed893afbd3a9b16990b530fd3fcaca679fb6d12.

tobil4sk: Thank you!

tools/CommandLineTools.hx
```diff
@@ -485,15 +485,14 @@ class CommandLineTools
 				}
 
 			case MAC:
-				// if (System.hostArchitecture == X64) {
-
-				untyped $loader.path = $array(path + "Mac64/", $loader.path);
```
player-03, line 490: This used to run every time. Is it safe to replace it with an `if` block with no final `else`?

tools/SVGExport.hx
```diff
@@ -67,8 +67,14 @@ class SVGExport
 			// }
 
 			case MAC:
-				untyped $loader.path = $array(path + "Mac/", $loader.path);
-				untyped $loader.path = $array(path + "Mac64/", $loader.path);
```
player-03, line 71: Same issue here. Previously ran every time, now only runs on 64-bit systems.

tools/CommandLineTools.hx
tobil4sk: I don't think there is much point in adding Mac64 for non x86_64 bit systems. The architecture will not match so the ndll will fail to load anyway, so either way the ndll fails to load. But if you think it's better, I can change it so that Mac64 is added by default.

player-03: No, you have a point there. I guess I didn't think it all the way through.

tobil4sk:
This is needed to avoid the neko loading error when running lime tools with a native build of neko on arm64 Mac. It shouldn't change anything when running via rosetta. Anything left to do before this can be merged? @player-03

player-03:
I tend to avoid merging Mac-related PRs, since I (mostly) don't use Mac. (With the exception of very small PRs that I can verify just by reading the code.)

PR #1759, opened 2024-02-08
Increased accuracy when getting current audio time using OpenAL SOFT on native

ninjamuffin99:
First time writing some bindings stuff! so feel free to tell me where I might need to clean up things. (i will clean up random whitespace / imports changes in the morning, bwah!)

wrote some bindings for `al_get_sourcedv_soft` and broadly followed along side the `get_sourcefv` stuff.

Using openal soft extension, can get a more precise time of the current playing audio.

Using [ninjamuffin99/Lime-Time](https://github.com/ninjamuffin99/Lime-Time) as sample / benchmark.

(the graphs here are sorta "right to left", and loop back to the bottom every 1000ms.)
`stats.update(FlxG.sound.music.time % 1000);`
### Before (60FPS)
image
The timing of the audio 'plateaus' a bit for at least a frame, which may not seem like much, however this is an issue if you were to attempt to get the current playing time from BETWEEN a frame (say, on a keyboard input via an event listener) or on higher framerates (my demo is at 60fps)

### After (60FPS)
image
The values here are ALWAYS being incremented, due to higher precision. No 'plateau' effect.


src/lime/media/OpenALAudioContext.hx
```diff
@@ -91,6 +91,13 @@ class OpenALAudioContext
 	public var DEFAULT_ALL_DEVICES_SPECIFIER:Int = 0x1012;
 	public var ALL_DEVICES_SPECIFIER:Int = 0x1013;
 
+	public var DEVICE_CLOCK_SOFT:Int = 0x1600;
+	public var DEVICE_LATENCY_SOFT:Int = 0x1601;
+	public var DEVICE_CLOCK_LATENCY_SOFT:Int = 0x1602;
+
+	public var SEC_OFFSET_LATENCY_SOFT:Int = 0x1201;
+	public var SEC_OFFSET_CLOCK_SOFT:Int = 0x1203;
+
```
player-03: Ugh, why are we even storing ~90 constants as variables? (Not your fault/responsibility, this PR just made me aware of it.)

Is it so that `openfl.media.SoundChannel` can access it as an instance variable? That implies it might be overridden, like via an alternative AL implementation. But the accesses only happen within `#if lime_openal`, so it wouldn't even attempt to get them if it wasn't OpenAL.

player-03: I took a closer look, and this file's list was always incomplete. Perhaps intentionally? In any case there's no need for this PR to add to it.

ninjamuffin99:
cleaned up the whitespacing, and removed those unneeded public vars 🤝

ghost:
openal soft gives null object referance at iOS
IMG_3256

SomeGuyWhoLovesCoding:
> openal soft gives null object referance at iOS IMG_3256

Ikr? LMAO 🤣

MAJigsaw77:
Maybe a define needs to be added as openal soft is not on ios or lime should add support for openal soft on ios.

ghost:
> Maybe a define needs to be added as openal soft is not on ios or lime should add support for openal soft on ios.

OpenAL SOFT is not supported for `static_link` or for iOS itself so lime needed add support for it

joshtynjala:
> OpenAL SOFT is not supported for static_link or for iOS itself so lime needed add support for it

OpenAL SOFT is LGPL, so as I understand it, it must be dynamically linked.

ghost:
> OpenAL SOFT is LGPL, so as I understand it, it must be dynamically linked.

Makes sense, but still OpenAL SOFT needs iOS fixes/patches

ACrazyTown:
Any updates on this?

MAJigsaw77:
> Any updates on this?

iOS is currently broken

ACrazyTown:
> > Any updates on this?
> 
> iOS is currently broken

Yeah I saw, I asked to see if there was any progress on fixing it and getting this ready to merge

Dimensionscape:
> > > Any updates on this?
> > 
> > 
> > iOS is currently broken
> 
> Yeah I saw, I asked to see if there was any progress on fixing it and getting this ready to merge

We dont use OpenAL when static linking. Instead we use MojoAL. see: https://github.com/openfl/lime/blob/ca52db81efa10b1bd8c76e43c714faa9dbbc4c94/project/Build.xml#L26

Dimensionscape:
We don't use OpenAL Soft when static linking. Instead we use MojoAL. 

See: https://github.com/openfl/lime/blob/ca52db81efa10b1bd8c76e43c714faa9dbbc4c94/project/Build.xml#L26

From what I understand, apple devices have the proprietary licensed OpenAL embedded in the system api's(as opposed to OpenAL Soft) and we use that for those targets [here](https://github.com/openfl/lime/blob/ca52db81efa10b1bd8c76e43c714faa9dbbc4c94/project/src/media/openal/OpenALBindings.cpp#L1C1-L3C24).

Any changes need to take account for our different OpenAL backends.

dimensionscape:
We don't use OpenAL Soft when static linking. Instead we use MojoAL. 

See: https://github.com/openfl/lime/blob/ca52db81efa10b1bd8c76e43c714faa9dbbc4c94/project/Build.xml#L26

From what I understand, apple devices have the proprietary licensed OpenAL embedded in the system api's(as opposed to OpenAL Soft) and we use that for those targets here:
https://github.com/openfl/lime/blob/ca52db81efa10b1bd8c76e43c714faa9dbbc4c94/project/src/media/openal/OpenALBindings.cpp#L2

Any changes need to take into account for our different OpenAL backends.

ninjamuffin99:
hmmm I'll have to do lil tests to see how accurate mojoAL is then. It could be that the previous implemtation might work fine on iOS if it's accurate enough

dimensionscape:
It feels a little dirty having a function that only works for a specific AL backend, but it sounds like it should be something we can manage uniformly. ✌

ACrazyTown:
Just wondering, would [setCurrentTime](https://github.com/FunkinCrew/lime/blob/openal-soft-pr/src/lime/_internal/backend/native/NativeAudioSource.hx#L422) also have to account for these changes? I see it references `AL.BYTE_OFFSET` like the old audio time method did so I'm not quite sure

ACrazyTown:
What this PR seems to do is apply latency compensation to the audio time directly. Wouldn't it be better to maybe have a seperate field/function to retrieve the audio latency and let the user apply it to the time manually? Right now this would create inconsistent behavior as targets using OpenAL Soft would give latency compensated time, while others would not

LeonGamerPS1:
IKR

PR #1764, opened 2024-02-25
Remove conditional compilation from `AssetsMacro` class.

### LINKED ISSUE #1763: Nightly Haxe issues
Just mentioning this here in case it doesn't solve itself.

Flixel CI has started failing on nightly haxe versions due to lime's [shadowed implementation of haxe.io.Bytes](https://github.com/openfl/lime/blob/develop/src/haxe/io/Bytes.hx) with the following errors:

Cpp/Neko:
> /.../_internal/macros/AssetsMacro.hx:41: characters 26-33 : Too many arguments

HL:
> /.../AssetsMacro.hx:78: characters 12-19 : hl.Bytes should be Int for optional function argument 'length'

HTML5:
> /.../AssetsMacro.hx:37: characters 12-26 : js.lib.ArrayBuffer should be Int for optional function argument 'length'

the OG [haxe.io.Bytes](https://github.com/HaxeFoundation/haxe/blob/development/std/haxe/io/Bytes.hx) was changed for the first time in 5 years, all [they did](https://github.com/HaxeFoundation/haxe/pull/11551/files#diff-16a72a23cd8bdf5dbf63ae4216d55c6bed8fd076c79cb8f2f308ab7608a24ad5) was remove `#if cs` sections, so I'm not sure why that would cause these specific errors

My main concern is that we don't really know why lime has it's own Bytes class, so if things do change with Bytes, lime will need to manually make those same changes. I wonder if it's simply better to create a lime.io.Bytes and use that? I also don't see why lime's Bytes has a different constructor for js

player-03:
A [recent change](https://github.com/HaxeFoundation/haxe/commit/342b1387e287241fac6b8300a29e7955d7bf4429) to the macro context seems to prevent us from using some conditional compilation flags. I have only ever relied on `#if macro` working during a macro context, and suspect it was a fluke that `#if html5` ever worked.

In fact, it seems that `#if hl` never worked. ddd19afe5273aea61a2a6487618dcad438c2dfbd adds the check but prefers to call `embedBytesHL()` instead, so it doesn't get checked. Similarly, `#if js` and Haxe-provided flags never work in macros; only Lime's flags ever stood a chance.

This PR switches to a more future-proof approach, and also tries to make the code more readable. I believe this closes #1763 (unless more errors get revealed once these are fixed).

### DIFF
--- a/src/haxe/io/Bytes.hx
+++ b/src/haxe/io/Bytes.hx
@@ -945,7 +945,7 @@ class Bytes
 }
 #elseif hl
 #if !macro
-@:autoBuild(lime._internal.macros.AssetsMacro.embedBytesHL()) // Enable @:bytes embed metadata
+@:autoBuild(lime._internal.macros.AssetsMacro.embedBytes()) // Enable @:bytes embed metadata
 #end
 @:coreApi
 class Bytes
--- a/src/lime/_internal/macros/AssetsMacro.hx
+++ b/src/lime/_internal/macros/AssetsMacro.hx
@@ -46,25 +46,6 @@ class AssetsMacro
 		return fields;
 	}
 
-	macro public static function embedBytesHL():Array<Field>
-	{
-		var fields = embedData(":file");
-		if (fields == null) return null;
-
-		var definition = macro class Temp
-		{
-			public function new(?length:Int, ?bytesData:haxe.io.BytesData)
-			{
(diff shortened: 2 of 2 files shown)


player-03:
The tests aren't going to check this, since none of the sample projects embed their assets. But I tested it on my machine at least.

PR #1774, opened 2024-04-02
Limit total time spent on green threads.

player-03:
Every single-threaded `ThreadPool` takes up a certain fraction of the app's time per frame. Without any coordination, they could take up more than 100% of available time, causing the app to slow down. By using static variables, we can make them work together to limit the total time spent per frame.

For instance, at 60 fps and a `workLoad` of 0.5, `ThreadPool`s will work together to take up approximately 1/120 of a second each frame. This leaves another 1/120 for everything else on the main thread.

This also fixes a null pointer error by waiting to access `Window.frameRate` until a window is actually available.


player-03:
@jgranick This PR should in theory fix the crash(es) you were seeing. Can you confirm?

PR #1775, opened 2024-04-06
sign only when provisioning file exists on Project.xml

ghost:
that'll allow us to compile on github actions or people has no apple account/provisioning thing

### DIFF
--- a/src/lime/tools/IOSHelper.hx
+++ b/src/lime/tools/IOSHelper.hx
@@ -30,6 +30,13 @@ class IOSHelper
 		else
 		{
 			commands.push("build");
+			if (!project.config.exists("ios.provisioning-profile"))
+			{
+				commands.push("CODE_SIGN_IDENTITY=\"\"");
+				commands.push("CODE_SIGNING_REQUIRED=\"NO\"");
+				commands.push("CODE_SIGN_ENTITLEMENTS=\"\"");
+				commands.push("CODE_SIGNING_ALLOWED=\"NO\"");
+			}
 		}
 
 		if (additionalArguments != null)
--- a/src/lime/tools/IOSHelper.hx
+++ b/src/lime/tools/IOSHelper.hx
@@ -26,11 +26,6 @@ class IOSHelper
 			commands.push(project.app.file);
 			commands.push("-archivePath");
 			commands.push(Path.combine("build", Path.combine(configuration + "-" + platformName, project.app.file)));
-			if (!project.config.exists("ios.provisioning-profile"))
-			{
-				commands.push("CODE_SIGNING_REQUIRED=\"NO\"");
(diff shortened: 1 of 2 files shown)


joshtynjala:
When I've built iOS applications with Lime, I don't think that I have ever specified a provisioning profile. Instead, I specify a team ID.

Either this way (in project.xml):

```xml
<config:ios team-id="XXXXXXXXXX"/>
```

Or this way (in a terminal):

```sh
lime build ios --certificate-team-id=XXXXXXXXXX
```

ghost:
[not on actions sadly, on actions it's complicated, provisioning profile can expire](https://docs.github.com/en/actions/deployment/deploying-xcode-applications/installing-an-apple-certificate-on-macos-runners-for-xcode-development) and if you not have an apple account... you can't really build which I don't have any

joshtynjala:
My point was that this change seems like it might break my ability to use team-id only because I never specify provisioning-profile:

> if (!project.targetFlags.exists("simulator") && project.config.exists("ios.provisioning-profile"))

ghost:
> My point was that this change seems like it might break my ability to use team-id only because I never specify provisioning-profile:
> 
> > if (!project.targetFlags.exists("simulator") && project.config.exists("ios.provisioning-profile"))

isn't `project.config` meant to be for Project.xml?

joshtynjala:
Yes, why?

ghost:
I meant to say if it's `project.config.exists("ios.provisioning-profile")` only should work when being set by Project.xml not by global, so I don't think it's going to make issues when you use `lime build ios --certificate-team-id=XXXXXXXXXX`

ghost:
you can try it by yourself too (I don't own a mac sorry)

joshtynjala:
_Sorry, I edited this is a few times to clarify things_

I'm pretty sure that `--certificate-team-id` is treated as an override of `<config:ios team-id=""/>`, so it's basically the same as if you had specified it in project.xml.

Regardless, you seem to have forgotten the first example in my message, which goes in project.xml: `<config:ios team-id="XXXXXXXXXX"/>`. This can be used to sign without also including `<config:ios provisioning-profile="whatever"/>`, so if I add team-id only, it won't sign because it is now strictly requiring provisioning-profile, even though I don't need it.

ghost:
oh yeah sorry I can fix

joshtynjala:
I feel like a `-nosign` command line option would be a better solution. It would allow people to opt into skipping the signing step, while also continuing to provide an error message if you forget team-id/provisioning-profile (or anything else that we might support in the future) when you don't opt in.

ghost:
> I feel like a `-nosign` command line option would be a better solution. It would allow people to opt into skipping the signing step, while also continuing to provide an error message if you forget team-id/provisioning-profile (or anything else that we might support in the future) when you don't opt in.

sure let me try

PR #1776, opened 2024-04-08
`-nosign` option for iOS

ghost:
better version of #1775 but since idk about messaging in hxp, I can't really do that sorry


joshtynjala:
Merged into the 8.2.0-Dev branch.

PR #1780, opened 2024-04-14
Increase audio length limit majorly on desktop

SomeGuyWhoLovesCoding:
So, there was an int64 conversion in the way of making ``time``, ``length``, ``dataLength``, and ``samples`` floats, causing an overflow once you get past the int64 limit. To fix that, I simply converted it to a float, and now you can load audio files over 6.6 hours with 44.1k sample rate!

I did make ``project/src/media/containers/OGG.cpp`` readable too.

I'm also planning to fix the memory leak problem and another problem with audio buffer when streaming audio. But I don't even know how I would fix that because I simply have never heard of some way to fix it entirely when doing ``new AudioSource(AudioBuffer.fromVorbisFile(VorbisFile.fromFile(sourceFile)))`` for the audio source (Which streams the ogg audio).

### DIFF
--- a/src/lime/_internal/backend/native/NativeAudioSource.hx
+++ b/src/lime/_internal/backend/native/NativeAudioSource.hx
@@ -20,8 +20,8 @@ import lime.utils.UInt8Array;
 class NativeAudioSource
 {
 	private static var STREAM_BUFFER_SIZE:Int = 8192;
-	private static var STREAM_NUM_BUFFERS:Int = #if (native_audio_buffers && !macro) Std.parseInt(haxe.macro.Compiler.getDefine("native_audio_buffers")) #else 16 #end ;
-	private static var STREAM_TIMER_FREQUENCY:Int = 80;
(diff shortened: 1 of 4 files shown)


src/lime/_internal/backend/native/NativeAudioSource.hx
```diff
@@ -199,47 +149,35 @@ class NativeAudioSource
 		AL.sourcePause(handle);
 
 		if (streamTimer != null)
-		{
```
joshtynjala, line 202: Please don't remove braces like this. It makes it more difficult to filter through the PR to find the truly relevant parts. Additionally, though, while you may personally prefer to omit braces, others prefer to always include braces, even when optional, so best not to change what may have been an intentional choice.

SomeGuyWhoLovesCoding: Ok

SomeGuyWhoLovesCoding:
The reason why I closed this because it fucks up the audio playback on some native targets, especially on hl

PR #1783, opened 2024-05-10
Build.xml: Add `/opt/homebrew/lib` as rpath on arm64 macos

### LINKED ISSUE #1750: `lime.ndll` fails to load `libneko.dylib` with certain homebrew installs of haxe/neko
When haxe/neko are installed via homebrew, they are installed to `/opt/homebrew` on an Apple Silicon Mac system. This means that libneko exists in `/opt/homebrew/lib` rather than the traditional paths. As a result, when `lime.ndll` tries to load `libneko.dylib` (via [`dlopen`](https://github.com/HaxeFoundation/hxcpp/blob/45ee673796438ea5e8e57cb9ee4b7f47ea47d213/include/hx/CFFINekoLoader.h#L28)), it fails because dlopen does not check the homebrew path.

Running a `lime` command from the terminal gives the error:
```
Could not link to neko.
Called from lime/_internal/backend/native/NativeCFFI.hx line 632
Called from lime/system/CFFI.hx line 177
Called from lime/system/CFFI.hx line 300
Uncaught exception - Segmentation fault
```

Homebrew does not have a solution for this problem: https://github.com/Homebrew/brew/issues/13481.

To work around this, it is possible to set `DYLD_LIBRARY_PATH` to `/opt/homebrew/lib`, but this only works if [System Integrity Protection is disabled](https://developer.apple.com/documentation/security/disabling_and_enabling_system_integrity_protection), which is not ideal.

One possible option is for `lime.ndll` to have a hardcoded full path to `libneko` (by setting install_name). However, system specific paths wouldn't really work for the precompiled lime.ndlls distributed via haxelib. Also, this would only work if lime.ndll is compiled with `-lneko`, which is currently not the case.

Also, it looks like `/usr/local/lib` may not be a default path anymore either: https://developer.apple.com/forums/thread/737920.

Not sure if this is the best place to put this, but lime won't work natively on Apple Silicon without resolving this.

tobil4sk:
This is where arm64 homebrew installs `libneko.dylib`. However, it may still be in `/usr/local/lib` if installed via the haxe .pkg installer, so we add both. See the discussion in: https://github.com/openfl/lime/commit/c70ec9fbe02b8d98f6562ee7a403f812549ea387.

Closes #1750.

We should use `install_name_tool -delete-rpath [path]` to remove rpaths when copying `lime.ndll` into a cpp build, because we only need the rpath for neko builds (where `libneko.dylib` is going to be loaded), and having an unnecessary rpath is a security concern. I'm not too sure where to do this, I guess maybe here after copying it? @joshtynjala 
https://github.com/openfl/lime/blob/c70ec9fbe02b8d98f6562ee7a403f812549ea387/tools/platforms/MacPlatform.hx#L175-L178.

### DIFF
--- a/project/Build.xml
+++ b/project/Build.xml
@@ -447,6 +447,7 @@
 					/usr/local/lib, but we need it for neko
 				-->
 				<vflag name="-rpath" value="/usr/local/lib" />
+				<vflag name="-rpath" value="/opt/homebrew/lib" if="HXCPP_ARM64"/>
 
 				<vflag name="-l" value="iconv" />
 				<vflag name="-framework" value="IOKit" />


joshtynjala:
> I'm not too sure where to do this, I guess maybe here after copying it?

Yeah, in the `if (targetType == "neko")` section below seems good to me:

https://github.com/openfl/lime/blob/c70ec9fbe02b8d98f6562ee7a403f812549ea387/tools/platforms/MacPlatform.hx#L183-L191

I put similar `install_name_tool` commands for HashLink in the `else if (targetType == "hl")` section after that in the 8.2.0-Dev branch.

PR #1787, opened 2024-05-16
Fix error when locating Lime's bundled NDLLs.

### LINKED ISSUE #1786: Using HXP complains about a missing NDLL
# Reproduction

- Use any project with a basic `project.hxp` file overriding the `project.xml`. Here is my Hello World example: https://github.com/EliteMasterEric/Flixel-HelloWorld-HXP
- Run `lime test windows -debug`

# Expected Result
The program should build and run as it does with the HXP disabled (easily test this by renaming the HXP file).

# Actual Result

The following error is displayed:

```
Uncaught exception - Could not find lime.ndll. This file is provided with Lime's Haxelib releases, but not via Git. Please copy it from Lime's latest Haxelib release into either E:\Programming\Haxe\Flixel-HelloWorld\.haxelib\lime/8,1,1/src//ndll/Windows or E:\Programming\Haxe\Flixel-HelloWorld\.haxelib\lime/8,1,1/src//ndll/Windows64, as appropriate for your system. Advanced users may run `lime rebuild cpp` instead.
```

Note the incorrect file path (`./src`); the NDLL file does exist if you look in the correct file path.

# Version info

Haxe 4.3.4
HXP 1.3.0
Lime 8.1.1

player-03:
We _could_ fall back to `haxelib path` for very old versions of Haxelib, but most likely no one uses a version that old. Even if you're stuck on Haxe 3, you can still update Haxelib. Also, `haxelib path` gave incorrect results in the first place, so falling back wouldn't do any good.

I believe this will resolve #1786.

### DIFF
--- a/src/lime/system/CFFI.hx
+++ b/src/lime/system/CFFI.hx
@@ -207,39 +207,29 @@ class CFFI
 	private static function __findHaxelib(library:String):String
 	{
 		#if (sys && !macro && !html5)
+		var proc:Process;
 		try
 		{
-			var proc = new Process("haxelib", ["path", library]);
+			proc = new Process("haxelib", ["libpath", library]);
+			if (proc == null) return "";
+		} catch (e:Dynamic) {}
 
-			if (proc != null)
-			{
-				var stream = proc.stdout;
+		var stream = proc.stdout;
+		var path:String = "";
 
-				try
-				{
-					while (true)
-					{
-						var s = stream.readLine();
+		try
+		{
+			path = stream.readLine();
+			__loaderTrace("Found haxelib " + path);
+		} catch (e:Dynamic) {}
 
-						if (s.substr(0, 1) != "-")
-						{
-							stream.close();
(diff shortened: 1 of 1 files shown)


tobil4sk:
> `haxelib path` gave incorrect results in the first place

If you run `haxelib path lime` the output looks something like this:

```hxml
-L /path/to/lime/git/ndll/
--macro lime._internal.macros.DefineMacro.run()
/path/to/lime/git/src/
-D lime=8.1.1
```

The `-L` flags have the ndll path, that's the correct thing to use if it's coming from `haxelib path`.

player-03:
We could do that too, as long as we rename the function to indicate that's what it's for.

tobil4sk:
Looks like this happens here:
https://github.com/openfl/lime/blob/d2ea41570bc195bd3c1ded9e67a2881b3cf31df2/tools/CommandLineTools.hx#L456-L462

PR #1788, opened 2024-05-23
Support pannerAttr for Howler sounds + improve performance

DigiEggz:
Exposes the `pannerAttr` field, which allows panner values to be set. By setting the `panningModel` to `"equalpower"`, which results in crisp sound within Howler, resolving the problem mentioned in https://github.com/goldfire/howler.js/issues/112.

The processing performance of `"equalpower"` is notable compared to HRTF: https://padenot.github.io/web-audio-perf/#pannernode-when-panningmodel--hrtf

In the future, panner attributes can be passed within the HTML5AudioSource `play()` function, or anywhere after the sound is active. It could also make sense to set a global pannerAttr that is referenced within the `init()` function.

```
public function play():Void
{
    …
    id = parent.buffer.__srcHowl.play();

    untyped parent.buffer.__srcHowl._volume = cacheVolume;
    parent.buffer.__srcHowl.pannerAttr({coneInnerAngle: 50, panningModel: "HRTF”});) // An example of passing two new panner attributes
    // setGain (parent.gain);
    …
}
```

### DIFF
--- a/src/lime/_internal/backend/html5/HTML5AudioSource.hx
+++ b/src/lime/_internal/backend/html5/HTML5AudioSource.hx
@@ -39,7 +39,7 @@ class HTML5AudioSource
 				maxDistance: 10000,
 				refDistance: 1,
 				rolloffFactor: 1,
-				panningModel: "equalpower" // Default to equalpower for crisp sound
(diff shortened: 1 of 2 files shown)


joshtynjala:
Merged into 8.3.0-Dev

PR #1789, opened 2024-05-24
Set title to value instead of return of backend.setTitle

superpowers04:


joshtynjala:
@superpowers04  Is `__backend.setTitle(value)` returning the wrong value?

superpowers04:
Oh god, I am dumb. I uhh thought __backend was a reference to a native library, completely forgetting that the compiler would throw a fit if the function returned something other than a String and completely ignoring that WindowBackend is a typedef at the bottom of the file to a normal Haxe class. I apologise

PR #1790, opened 2024-05-28
Automatically delete stale assets and dependency files.

### LINKED ISSUE #1546: Stale output from previous builds can cause errors

player-03:
This is yet another attempt to solve #1546, and has several advantages over the previous attempts.

- The code is much more concise.
- No change to HXP is required, meaning there's no need for a synchronized update.
- Unlike #1547, it does not need to iterate over all files in the output directory.
- Unlike #1550, it handles all types of assets, not just templates.

### DIFF
--- a/src/lime/tools/AssetHelper.hx
+++ b/src/lime/tools/AssetHelper.hx
@@ -191,25 +191,6 @@ class AssetHelper
 		return manifests;
 	}
 
-	public static function deleteStaleAssets(project:HXProject, targetDirectory:String):Void
-	{
-		var newAssets = [for (asset in project.assets) asset.targetPath];
-
-		var recordFile:String = targetDirectory + "/.assets";
-		if (FileSystem.exists(recordFile))
-		{
-			for (oldAsset in File.getContent(recordFile).split("\n"))
-			{
-				if (oldAsset.length > 0 && newAssets.indexOf(oldAsset) < 0)
-				{
-					System.deleteFile(targetDirectory + "/bin/" + oldAsset);
-				}
-			}
-		}
-
-		File.saveContent(recordFile, newAssets.join("\n"));
-	}
-
 	private static function getAssetData(project:HXProject, pathGroups:Map<String, Array<String>>, libraries:Map<String, Library>, library:String,
 			asset:Asset):Dynamic
 	{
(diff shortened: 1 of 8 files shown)


player-03:
@justin-espedal, how does this look?

player-03:
It looks like this isn't enough to solve the original issue after all. The example given there is an Android dependency, and since that isn't an asset this PR won't do anything about it. Instead, `AndroidPlatform` calls `System.recursiveCopy()` to copy the whole folder, making no attempt to clear out files that weren't touched.

Meanwhile, not all assets and templates use `System` to copy them, so we can't just go based on `System` either.

justin-espedal:
Thanks for considering this issue again after so long.

And thanks for looking into the original problem, too. I suspected that may be the case, but wasn't able to make time to look into it yesterday. This approach does look a lot cleaner though!

player-03:
I have a plan to handle dependencies, but I'm still working on it. The final result won't be quite as concise as this, but it should still work without needing to modify HXP.

player-03:
And that should do it. Give it a try when you have time.

PR #1791, opened 2024-05-30
Create a placeholder object to send `Matrix3`s to C++ code.

player-03:
The C++ code was written with an object in mind, rather than an array buffer. The easiest workaround is to give it what it expects.


joshtynjala (approved):
Looks good to me!

PR #1792, opened 2024-05-30
Make `Matrix3` use column-major order.

player-03:
[What "column-major" means](https://en.wikipedia.org/wiki/Row-_and_column-major_order)

Now that `Matrix3` is stored as a one-dimensional array (see #1778), it's important to keep careful track of what order data is stored in. I chose column-major order because [that's what GLSL uses](https://www.khronos.org/opengl/wiki/Data_Type_\(GLSL\)#Matrices), and it's also how `Matrix4` works.

This PR swaps `b` and `c` (which had been backwards), adds documentation, and tidies up.

### DIFF
--- a/src/lime/math/Matrix3.hx
+++ b/src/lime/math/Matrix3.hx
@@ -22,7 +22,7 @@ import lime.utils.Float32Array;
 @:fileXml('tags="haxe,release"')
 @:noDebug
 #end
-abstract Matrix3(Float32Array) from Float32Array to Float32Array
+abstract Matrix3(Float32Array) to Float32Array
 {
 	/**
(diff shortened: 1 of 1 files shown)


player-03:
I don't think this fixes the error @joshtynjala reported, though it's worth checking. Yes, `b` and `c` were swapped, but they were swapped twice: once in the constructor and once in the getter/setter. Since all existing code accessed the properties rather than the underlying array, it wouldn't matter how out of order the array was.

joshtynjala:
> I don't think this fixes the error @joshtynjala reported, though it's worth checking.

I confirm that it does not, but it's good that you've made things more consistent.

PR #1793, opened 2024-06-02
Android Preserve External Storage for Android 10+

ghost:
See https://developer.android.com/training/data-storage/use-cases#opt-out-scoped-storage and https://developer.android.com/training/data-storage/use-cases#migrate-legacy-storage for more info.

### DIFF
--- a/tools/platforms/AndroidPlatform.hx
+++ b/tools/platforms/AndroidPlatform.hx
@@ -483,6 +483,8 @@ class AndroidPlatform extends PlatformTarget
 			"android:allowBackup": "true",
 			"android:theme": "@android:style/Theme.NoTitleBar" + (project.window.fullscreen ? ".Fullscreen" : null),
 			"android:hardwareAccelerated": "true",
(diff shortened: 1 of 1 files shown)


KarimAkra (approved):

player-03:
By default, OpenFL and Lime save data in the folder returned by [`getFilesDir()`](https://developer.android.com/reference/android/content/Context#getFilesDir\(\)). As far as I can tell from the documentation, this folder will remain accessible.

If you've stored data in other folders and you need these settings to make sure you can access that data, you can enable them from project.xml:

```xml
<config:android>
	<application android:requestLegacyExternalStorage="true" android:preserveLegacyExternalStorage="true" />
</config:android>
```

I might be missing something, but I don't see why these would need to be built in to Lime.

ghost:
Oh, I didn't know sorry closing up this rn

player-03:
No problem. This is still an undocumented feature for now, so you'd have to have been paying very close attention to the PRs to be aware of it.

We'll release proper documentation once Lime 8.2.0 comes out.

PR #1795, opened 2024-06-09
Some `android` fixes and added additional `hardware` features.

MAJigsaw77:
- Use `namespace` in the app's Gradle file across all the Gradle projects to fix deprecation warnings.
- Added Drop file event support by including the intent-filter [reference](https://github.com/libsdl-org/SDL/blob/9651ca59187c16079846918483c40d6b5c2f454c/android-project/app/src/main/AndroidManifest.xml#L95-L102) and support for additional hardware features and USB device attachment.

### DIFF
--- a/dependencies/extension-api/build.gradle
+++ b/dependencies/extension-api/build.gradle
@@ -12,6 +12,7 @@ buildscript {
 apply plugin: 'com.android.library'
 
 android {
+	namespace 'org.haxe.extension'
 	compileSdkVersion Integer.parseInt(project.ANDROID_BUILD_SDK_VERSION)
 	buildToolsVersion project.ANDROID_BUILD_TOOLS_VERSION
 
--- a/dependencies/extension-api/src/main/AndroidManifest.xml
+++ b/dependencies/extension-api/src/main/AndroidManifest.xml
@@ -1,4 +1,2 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="org.haxe.extension" >
-    
-</manifest>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" />
--- a/templates/android/template/app/build.gradle
+++ b/templates/android/template/app/build.gradle
@@ -13,6 +13,7 @@ System.setProperty('java.awt.headless','false')
 } */
 
 android {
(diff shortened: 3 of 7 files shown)


KarimAkra (approved):

MAJigsaw77:
https://github.com/openfl/lime/assets/77043862/3c1c76cc-c4f4-42a6-aadd-7c8379a1d3c5

This shows how the file drop works for android.

moha0100110:
> screen-20240609-144625.mp4 
> This shows how the file drop works for android.

pls can u make a haxe package that adds support for microphone capting pls we need it a lot

player-03:
> -Added Drop file event support by including the intent-filter [reference](https://github.com/libsdl-org/SDL/blob/9651ca59187c16079846918483c40d6b5c2f454c/android-project/app/src/main/AndroidManifest.xml#L95-L102) and support for additional hardware features and USB device attachment.

Since this is a new feature, it definitely goes in 8.2.0-Dev.

> - Use `namespace` in the app's Gradle file across all the Gradle projects to fix deprecation warnings.

Since this is fixing an issue, I'd say it could go on the develop branch. Though it isn't urgent, so 8.2.0-Dev is fine too. Either way, maybe make two different pull requests.

> - Updated `AndroidManifest.xml` to include new hardware features and improve compatibility.
> - Added optional support for:
>   - Bluetooth
>   - Gamepad
>   - USB host
>   - PC-type external mouse input
>   - Low latency audio
>   - Pro audio
> - Included a new intent filter for handling USB device attachment events.
> - Ensured all new features are optional by setting `android:required="false"`.
> 
> These changes enhance the application's flexibility and compatibility with various input methods and devices, ensuring a better user experience across different hardware configurations.

I'm not convinced enough Lime apps use these to make them the default. Why not just set them on a per-app basis?

> Add `Bluetooth` permission.

This means apps will be listed as using Bluetooth even when they actually don't. Could turn users off from downloading an app. Again, why not just set it on a per-app basis?

> pls can u make a haxe package that adds support for microphone capting pls we need it a lot

This is off-topic for this thread. Please discuss it in the issue instead.

MAJigsaw77:
So, should i split the pr?

MAJigsaw77:
Well, the third pr mostly adds things that most sdl apps have by default but if it's not suitable for apps, i can revert the commit.

(Most of them are also optional)

PR #1799, opened 2024-06-13
Support `VIEW` intents on Android.

MAJigsaw77:
This pr adds Drop file event support by including the intent-filter [reference](https://github.com/libsdl-org/SDL/blob/9651ca59187c16079846918483c40d6b5c2f454c/android-project/app/src/main/AndroidManifest.xml#L95-L102) on`AndroidManifest.xml`.

### DIFF
--- a/templates/android/template/app/src/main/AndroidManifest.xml
+++ b/templates/android/template/app/src/main/AndroidManifest.xml
@@ -25,7 +25,7 @@
 
 			</intent-filter>
 
-			::foreach ANDROID_MIME_TYPES::
+			::foreach ANDROID_ACCEPT_FILE_INTENT::
 			<intent-filter>
 
 				<action android:name="android.intent.action.VIEW" />
--- a/templates/android/template/app/src/main/AndroidManifest.xml
+++ b/templates/android/template/app/src/main/AndroidManifest.xml
@@ -25,7 +25,7 @@
 
 			</intent-filter>
 
-			::foreach ANDROID_SUPPORTED_MIME_TYPES::
+			::foreach ANDROID_MIME_TYPES::
 			<intent-filter>
 
 				<action android:name="android.intent.action.VIEW" />
--- a/templates/android/template/app/src/main/AndroidManifest.xml
+++ b/templates/android/template/app/src/main/AndroidManifest.xml
@@ -25,6 +25,14 @@
 
 			</intent-filter>
 
+			<intent-filter>
+
+				<action android:name="android.intent.action.VIEW" />
+				<category android:name="android.intent.category.DEFAULT" />
+				<data android:mimeType="*/*" />
+
+			</intent-filter>
+
 		</activity>
 
 	</application>
--- a/templates/android/template/app/src/main/AndroidManifest.xml
+++ b/templates/android/template/app/src/main/AndroidManifest.xml
@@ -25,13 +25,15 @@
 
 			</intent-filter>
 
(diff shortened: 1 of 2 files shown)


MAJigsaw77:
This pr is splitted from #1795.

player-03:
Now that I've reviewed the documentation, I'm pretty sure this means whenever you choose to view any file, the system will display a list of _every single Lime application_ on the device. Including apps that weren't designed to receive files, whose developers didn't realize this was in the manifest.

So again, I really think this should be handled on a per-app basis by using templates. If Lime was going to include it, it would need to be disabled by default and configurable from project.xml.

MAJigsaw77:
> Now that I've reviewed the documentation, I'm pretty sure this means whenever you choose to view any file, the system will display a list of _every single Lime application_ on the device. Including apps that weren't designed to receive files, whose developers didn't realize this was in the manifest.
> 
> So again, I really think this should be handled on a per-app basis by using templates. If Lime was going to include it, it would need to be disabled by default and configurable from project.xml.

In the code `<data android:mimeType="*/*" />
` this line can indeed be modified to use tamplates.

MAJigsaw77:
@player-03 is the code fine now?

player-03:
I'm a little worried `<config:android mimeType="application/json" />` could give the wrong impression that you're somehow setting the type of the app itself. Here are some other ideas.

```xml
<config:android accepted-mime-type="application/json" />
<config:android drop-file-type="application/json" />
<config:android view-intent-filter="application/json" />
<config:android allow-drop-file="application/json" />
<config:android view-file="application/json" />
<config:android accept-intent="application/json" />
<config:android accept-file-intent="application/json" />
```

I think I like `accept-file-intent` as a compromise between length and descriptiveness.

- "Accept" makes it clear the app is receiving events from elsewhere, rather than creating them (another common way to open files). Plus, terms like "accept" and "allow" imply that there can be more than one of this config entry. Something like `view-intent-filter` could imply there's only one filter, and if you set it a second time you'll overwrite the first or narrow down the filter or something.
- "File" helps distinguish this from any other intents we might decide to support later.
- "Intent" is important in case someone wants to look up more information. Searching for "accept file android" primarily returns results about transferring files to your phone, whereas "accept file intent android" finds `<intent-filter />`.

MAJigsaw77:
> I'm a little worried `<config:android mimeType="application/json" />` could give the wrong impression that you're somehow setting the type of the app itself. Here are some other ideas.
> 
> ```
> <config:android accepted-mime-type="application/json" />
> <config:android drop-file-type="application/json" />
> <config:android view-intent-filter="application/json" />
> <config:android allow-drop-file="application/json" />
> <config:android view-file="application/json" />
> <config:android accept-intent="application/json" />
> <config:android accept-file-intent="application/json" />
> ```
> 
> I think I like `accept-file-intent` as a compromise between length and descriptiveness.
> 
> * "Accept" makes it clear the app is receiving events from elsewhere, rather than creating them (another common way to open files). Plus, terms like "accept" and "allow" imply that there can be more than one of this config entry. Something like `view-intent-filter` could imply there's only one filter, and if you set it a second time you'll overwrite the first or narrow down the filter or something.
> * "File" helps distinguish this from any other intents we might decide to support later.
> * "Intent" is important in case someone wants to look up more information. Searching for "accept file android" primarily returns results about transferring files to your phone, whereas "accept file intent android" finds `<intent-filter />`.

I'll change it to `accept-file-intent` then.

MAJigsaw77:
@player-03 Is it fine now?

PR #1800, opened 2024-06-13
Use `namespace` for gradle projects to fix deprecation warnings.

MAJigsaw77:
This pr updates all `Gradle` project files to use `namespaces`, fixing deprecation warnings.

### DIFF
--- a/dependencies/extension-api/build.gradle
+++ b/dependencies/extension-api/build.gradle
@@ -12,6 +12,7 @@ buildscript {
 apply plugin: 'com.android.library'
 
 android {
+	namespace 'org.haxe.extension'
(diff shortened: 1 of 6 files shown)


MAJigsaw77:
This pr is splitted from #1795.

player-03:
Hmm. Based on the documentation I was worried that this would break compatibility with AGP < 7.3, but after testing, it looks like something else already broke compatibility with those versions. So I guess it's fine?

MAJigsaw77:
@player-03 Will this be pushed?

player-03:
I looked at it again and despite a warning, the correct namespace string ended up in the correct place even on AGP 7.0.2. So it seems there was some level of support in earlier versions.

PR #1804, opened 2024-06-16
Android vibration fixes.

MAJigsaw77:
* Fixed some deprecation warnings.
* Cancel the vibration when the app is being paused or destroyed.

### DIFF
--- a/templates/android/template/app/src/main/java/org/haxe/lime/GameActivity.java
+++ b/templates/android/template/app/src/main/java/org/haxe/lime/GameActivity.java
@@ -139,14 +139,6 @@ public class GameActivity extends SDLActivity {
 
 	@Override protected void onDestroy () {
 
-		if (vibrator != null) {
-			
-			Log.d ("GameActivity", "Cancelling vibration");
-
-			vibrator.cancel ();
-
-		}
-
 		for (Extension extension : extensions) {
 
 			extension.onDestroy ();
--- a/templates/android/template/app/src/main/java/org/haxe/lime/GameActivity.java
+++ b/templates/android/template/app/src/main/java/org/haxe/lime/GameActivity.java
@@ -386,45 +386,35 @@ public class GameActivity extends SDLActivity {
 
 		}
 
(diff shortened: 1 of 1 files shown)


MAJigsaw77:
@player-03 Can i get a review on this pr?

templates/android/template/app/src/main/java/org/haxe/lime/GameActivity.java
```diff
@@ -137,6 +139,14 @@ protected void onCreate (Bundle state) {
...
 	@Override protected void onDestroy () {
 
+		if (vibrator != null) {
+			
+			Log.d ("GameActivity", "Cancelling vibration");
+
+			vibrator.cancel ();
+
+		}
+
```
player-03: Is this necessary? [The documentation](https://developer.android.com/reference/android/os/Vibrator) claims ongoing vibrations will be stopped automatically.

MAJigsaw77: > Is this necessary? [The documentation](https://developer.android.com/reference/android/os/Vibrator) claims ongoing vibrations will be stopped automatically.

I'll remove it then

player-03:
Interesting. According to the documentation, [the first entry in the array is the delay before vibration starts](https://developer.android.com/reference/android/os/VibrationEffect#createWaveform\(long[],%20int\)), so they recommend leaving that 0.

The question is, is this delay a bug or a feature? If it's a bug, we could fix it now. If it's a feature, we probably want to add an argument to toggle it on or off.

MAJigsaw77:
> Interesting. According to the documentation, [the first entry in the array is the delay before vibration starts](https://developer.android.com/reference/android/os/VibrationEffect#createWaveform(long%5B%5D,%20int)), so they recommend leaving that 0.
> 
> The question is, is this delay a bug or a feature? If it's a bug, we could fix it now. If it's a feature, we probably want to add an argument to toggle it on or off.

I don't really know what to say tbh

PR #1811, opened 2024-07-06
[ci] Add mac arm64 ndll to lime package

tobil4sk:


player-03:
I wonder, what's the rule for including an NDLL in the Haxelib package? Because we don't include all possible NDLLs. I think the answer is "any NDLL that's sufficiently popular," and I do get the impression that this is sufficiently popular (or will be, since it's the future of Mac). But maybe we should spell this out somewhere.

joshtynjala:
> any NDLL that's sufficiently popular

That has been my understanding as well.

> this is sufficiently popular (or will be, since it's the future of Mac)

Agreed.

PR #1813, opened 2024-07-08
[tools] Cross compile to x86 from arm64 on mac

tobil4sk:
I'm not sure if `-64` is the right flag to use here (since arm64 is a 64 bit architecture too), maybe it would be better to use something more specific like `-x86-64`?

### DIFF
--- a/tools/platforms/MacPlatform.hx
+++ b/tools/platforms/MacPlatform.hx
@@ -450,7 +450,14 @@ class MacPlatform extends PlatformTarget
 			case X86:
 				commands.push(["-Dmac", "-DHXCPP_CLANG", "-DHXCPP_M32"]);
 			case ARM64:
-				commands.push(["-Dmac", "-DHXCPP_CLANG", "-DHXCPP_ARM64"]);
+				if (targetFlags.exists("64"))
+				{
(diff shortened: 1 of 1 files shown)


joshtynjala:
> I'm not sure if -64 is the right flag to use here (since arm64 is a 64 bit architecture too), maybe it would be better to use something more specific like -x86-64?

We're already using `-64` and `-32` in a lot of places to mean 32-bit and 64-bit x86. With that in mind, I think that we should continue to use `-64` here for consistency.

However, I am not against adding an alternate flag that means the same thing, but has a clearer name. Perhaps even making that alternate the new default and deprecating `-32` and `-64` to eventually get removed in a few years or whatever.

tobil4sk:
> With that in mind, I think that we should continue to use -64 here for consistency.

Makes sense, thanks.

> However, I am not against adding an alternate flag that means the same thing, but has a clearer name.

Sounds good. Maybe we should deal with adding these in a separate PR/issue, because it looks like ~~android~~ ios is already using `"i386"` so we'll have to decide on consistent names for these.

tobil4sk:
Merging, and we can decide on the new flags in #1819.

PR #1814, opened 2024-07-09
Allow adding tags to AndroidManifest.xml from project.xml.

### LINKED ISSUE #1812: Add support for `maxSdkVersion` attribute in `permission` element.
There's currently no support for having a permission with a `maxSdkVersion` attribute, which could be useful to limit permissions to specific Android versions, ensuring compatibility.

player-03:
For instance, if you need to use an obscure tag like [`<uses-permission-sdk-23 />`](https://developer.android.com/guide/topics/manifest/uses-permission-sdk-23-element), this change will let you enter it into `<config:android />` rather than having to make a template.

```xml
<config:android>
    <manifest>
        <uses-permission-sdk-23 android:name="android.permission.ACCESS_WIFI_STATE" android:maxSdkVersion="30" />
    </manifest>
</config:android>
```

The nesting shown here is all required, and it tells Lime where to put this custom tag. In this case, it will go inside the `<manifest />` tag. `<application />` and `<activity />` are also available, and work similarly.

Resolves #1812.


player-03:
Oh yeah, and I streamlined `ConfigData` a bunch. Guess I could split that out into a separate PR.

MAJigsaw77:
I believe the `versionName` and `versionCode` are no longer required to be used in the AndroidManifest.xml as they are already used in the gradle file itself.

MAJigsaw77:
> I believe the `versionName` and `versionCode` are no longer required to be used in the AndroidManifest.xml as they are already used in the gradle file itself.

Even sdl doesn't use them anymore.

player-03:
The documentation doesn't mention them being deprecated, and some of their guides tell you to use build.gradle while others point you to the manifest. I figure we might as well keep them in both places for now, until Google gets their documentation sorted out.

MAJigsaw77:
Why isn't this pushed yet?

player-03:
In retrospect, maybe I could have merged it into 8.3.0.

I switched because I figured d0bbdf6ae4d1cf4dbec6dcf92190375fd8d448f1 might be a breaking change. After all, if the user specifies a config entry starting with "config:", it's now going to be ignored. ...But in the past it would also have been ignored, just for different reasons.

Oh well, 9.0.0 should be fine too.

tobil4sk:
> In retrospect, maybe I could have merged it into 8.3.0.

I'm quite interested in this feature, any chance it could be backported to 8.4.0-Dev as well?

player-03:
Done.

tobil4sk:
Thanks! Turns out there was a better solution for what I wanted to do (moving things to a lime extension's AndroidManfiest.xml and letting gradle merge things), but still nice to have this for extra flexibility.

PR #1818, opened 2024-07-12
Link against hxcpp's mbedtls in static builds

tobil4sk:
This avoids issues with lime's mbedtls taking priority over hxcpp's, which has happened in the past and has broken haxe standard library calls. See #1767 and https://github.com/HaxeFoundation/hxcpp/issues/1126.

The static lime library no longer includes mbedtls definitions. Mbedtls is now added at the end when everything is linked together. This means that if lime was linked against an incompatible version of mbedtls to what the current hxcpp version has, there will be a linker error instead of both versions being merged together silently and resulting in a buggy build. We are compiling hxcpp_ssl, which includes hxcpp's SSL.cpp but this is a small inconvenience and I think most linkers should remove the unneeded definitions. If hxcpp's xml files are reorganised, we can avoid this with future hxcpp versions (see https://github.com/HaxeFoundation/hxcpp/pull/1133).

`MBEDTLS_NET_C` is currently undefined by the mbedtls config for hxcpp (see https://github.com/HaxeFoundation/hxcpp/pull/1128). This causes this linker errors with curl, but we can avoid it by updating to curl 7.87.0.
```
/usr/bin/ld: /.../lime/ndll/Linux64/liblime.a(05cc3e30_mbedtls.o): in function `mbed_connect_common':
mbedtls.c:(.text+0xce1): undefined reference to `mbedtls_net_recv'
/usr/bin/ld: mbedtls.c:(.text+0xcf0): undefined reference to `mbedtls_net_send'
```

Further work seems to still be required to get this working on iOS and android, but right now I'm not too familiar with how the builds work there.

### DIFF
--- a/project/Build.xml
+++ b/project/Build.xml
@@ -389,7 +389,7 @@
 		<files id="native-toolkit-harfbuzz" if="LIME_HARFBUZZ" />
 		<files id="native-toolkit-jpeg" if="LIME_JPEG" />
 		<files id="native-toolkit-lzma" if="LIME_LZMA" />
-		<files id="native-toolkit-mbedtls" if="LIME_MBEDTLS" />
+		<files id="native-toolkit-mbedtls" if="LIME_MBEDTLS" unless="static_link" />
 		<files id="native-toolkit-mojoal" if="LIME_MOJOAL" />
 		<files id="native-toolkit-neko" if="LIME_NEKO" />
 		<files id="native-toolkit-ogg" if="LIME_OGG" />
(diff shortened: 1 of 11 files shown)


tobil4sk:
> Further work seems to still be required to get this working on iOS and android

I've made the required changes so it should be linking properly now on ios. It seems like android static builds are not possible at the moment, but if they are done using the static cpp template then it should include mbedtls correctly.

tobil4sk:
With hxcpp 4.3.2 (using mbedtls 2.9.0) and running `lime test windows -static`, this app crashes with current develop:

```haxe
import lime.app.Application;
import haxe.Http;

class Main extends Application {

	public function new () {

		super ();
		var connection:Http = new Http("https://google.com");
		connection.request();

	}

}
```

So [the ios crash](https://github.com/openfl/lime/pull/1767) would probably still occur with hxcpp 4.3.2 unless this patch is applied.

PR #1820, opened 2024-07-21
Add deleteEffect, deleteFilter to AL Class

Raltyro:
Adds missing function from OpenALBindings.cpp to AL, NativeCFFI Class
I'm aware that it does have a finalizer for gc, though just incase i'm adding this in


NeeEoo:
You should also add `lime_al_delete_auxiliary_effect_slot`.

PR #1801, opened 2024-06-13
Update HashLink to 1.14

EliteMasterEric:
Update HashLink from version 1.12 to version 1.14. This requires updating references to PCRE in the build file to PCRE2.

PR #1827, opened 2024-08-07
Bump HL version to 1.14.

Apprentice-Alchemist:
Closes #1801

### DIFF
--- a/project/lib/custom/hl-ssl/threading_alt.h
+++ b/project/lib/custom/hl-ssl/threading_alt.h
@@ -1,0 +1,6 @@
+#include <windows.h>
+
+typedef struct {
+  CRITICAL_SECTION cs;
+  char is_valid;
+} mbedtls_threading_mutex_t;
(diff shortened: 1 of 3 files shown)


project/lib/custom/hl-ssl/threading_alt.h
```diff
@@ -0,0 +1,6 @@
+#include <windows.h>
```
tobil4sk, line 1: If `WIN32_LEAN_AND_MEAN` is defined before this include then the compiler error with `net_sockets.c` can be avoided.

Otherwise, the way hashlink currently avoids this issue is by undefining the `MBEDTLS_NET_C` define: https://github.com/HaxeFoundation/hashlink/blob/9cf7efa118409bef3a7c263bffc36da107751e60/include/mbedtls/include/mbedtls/config.h#L2153C1-L2153C24

joshtynjala:
Merged into 8.3.0-Dev branch with commit 8d81a0fc8ed3a9f558a44507960b2992816475cc.

PR #1828, opened 2024-08-12
Expose more `ThreadPool` functionality through `Future.withEventualValue()`.

player-03:
Previously, `Future.withEventualValue()` tried to match the `Future` constructor, with one subtle difference in behavior. It interpreted null to mean "run the function again," which causes an infinite loop if the programmer returns null without realizing. Joshua found a regression of this sort in Lime itself, one I should have caught but overlooked.

This PR changes the function signature to avoid any possibility of this error. There is no return value anymore, so you can't accidentally return the wrong value. Instead, as with `ThreadPool`, you are given a `WorkOutput` object and must call `sendComplete()` at some point. The work function has the same signature as in `ThreadPool`, so if users understand the one, they'll understand the other.

I have a couple other branches that are closer to how `withEventualValue()` used to work, and I can bring them back if requested. But `withEventualValue()` is new in 8.2.0, so it shouldn't hurt to change its signature.

### DIFF
--- a/src/lime/app/Future.hx
+++ b/src/lime/app/Future.hx
@@ -84,7 +84,8 @@ import lime.utils.Log;
 			}
 			#end
 
-			FutureWork.run(dispatchWorkFunction, doWork, promise, useThreads ? MULTI_THREADED : SINGLE_THREADED);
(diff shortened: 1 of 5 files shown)


player-03:
Perhaps we should do this through `Promise` instead.

PR #1834, opened 2024-08-18
set correct paths for ndll for linxuarm and linuxarm64 (raspberrypi)

gepatto:
_set correct paths for ndll for linxuarm and linuxarm64 (raspberrypi)
and remove conflicting legacy includepaths in build.xml_

These are the changes I made which allow compilation on a raspberry pi 4 or 5 running 32 or 64bit Bullseye or Bookworm (raspberrypios).  A lot of the `isRaspberryPi` statements have been replaced by checking for Arm64 or Armv7 so that in the future other linux arm boards could be used as well.  Ndlls are built in ndll/LinuxArm and  ndll/LinuxArm64 instead of the previous ndll/RPi  path

Old -I flags have been removed from Build.xml because they point to libraries from older OS versions and do  no longer exist on bullseye and bookworm.  They prevented sdl from compiling on these os versions.
Performance on older versions of the raspberrypi was poor. So I think support  should be for pi4 and up only.
 
SDL still has pi specific code. 
That's why a -Drpi flag can still be found when calling hxcpp. Otherwise a build would fail.

_One important note is that for 64bit the latest (git)build of hxcpp is needed ! There is a bug in haxelibversion that caused threads to throw a bus error. This is solved in master._

### DIFF
--- a/project/Build.xml
+++ b/project/Build.xml
@@ -235,9 +235,6 @@
 
 			<section if="rpi">
 
-				<compilerflag value="-I/opt/vc/include" />
-				<compilerflag value="-I/opt/vc/include/interface/vcos/pthreads" />
-				<compilerflag value="-I/opt/vc/include/interface/vmcs_host/linux" />
 				<compilerflag value="-I/usr/include/libdrm" />
 				<compilerflag value="-I/usr/include/dbus-1.0" />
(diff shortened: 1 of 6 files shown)


joshtynjala:
8.2.0-Dev has already been merged into develop, for the upcoming release, so I changed the base branch of this PR to match.

PR #1837, opened 2024-09-02
Allow `ThreadPool` to run jobs in both modes, and improve scheduling

### LINKED ISSUE #1864: Future freezing application
```haxe
var completedJobs = 0;
lime.app.Future.FutureWork.maxThreads = 16;

var bitmapData:BitmapData = new BitmapData(1920, 1080);
var imageByte:Bytes = bitmapData.encode(bitmapData.rect, new PNGEncoderOptions());

var images:Array<Image> = [];
function createFuture():Future<Dynamic>
{
    trace("createFuture");
    return Image.loadFromBytes(imageByte).onComplete(image -> {
        completedJobs++;
        images.push(image);
        if(images.length > 4)
        {
            //images.shift();
            images = [];
            openfl.system.System.gc();
        }

        createFuture();
    });
}
for(i in 0...16)
{
    createFuture();
}

openfl.Lib.setInterval(() -> {
    trace(lime.app.Future.FutureWork.activeJobs);
    trace(completedJobs);

    // Allocate 1-10MB of memory
    Bytes.alloc(1024 * 1024 * (1 + Std.int(Math.random() * 10)));
}, 0);
```

image
image

player-03:
There are enough use cases for both single-threaded and multi-threaded jobs that it's likely some users will want to use both. Before this PR, this required [creating two `ThreadPool`s](https://github.com/openfl/lime/blob/bf4711a01d7a3cd9bb8e2bdf86fa653bdea356d5/src/lime/app/Future.hx#L352-L355), one for each mode. Now, you only need the one.

If you call `run()` without passing the new `mode` argument, it falls back to the pool's `mode`, producing the same behavior.

I've also updated how jobs are scheduled. Previously, it would call `__startJobs()` only once per frame. Now, it spins up new threads immediately if it hasn't hit the cap, and when a thread sends a `COMPLETE` or `ERROR` event, the main thread will immediately reply with the next job. This is a little bit faster than waiting 1/60th of a second for the next `__startJobs()` call.

On top of that, it can now run multiple single-threaded jobs per frame, including when one job's completion triggers the next.

### DIFF
--- a/src/lime/system/ThreadPool.hx
+++ b/src/lime/system/ThreadPool.hx
@@ -89,9 +89,9 @@ class ThreadPool extends WorkOutput
 	/**
 		__Access this only from the main thread.__
 
-		The sum of all active single-threaded pools' `workPriority` values.
+		The sum of `workPriority` values from all pools with an ongoing
+		single-threaded job.
 	**/
-	@:allow(lime.system.JobList)
 	private static var __totalWorkPriority:Float = 0;
 
 	/**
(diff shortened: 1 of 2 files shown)


rainyt:
I tested this PR and the thread speed returned to normal levels. Nice!

player-03:
Merged in f6393fc10727e97421d1bef278cb774bf52871d6.

PR #1838, opened 2024-09-02
Implement `Promise.completeAsync()`.

player-03:
Like #1828, but designed for the `Promise` class because I think that's more appropriate. Compared to `new Future()`, `completeAsync()` offers:

- The option to run on the main thread.
- The option to send progress events.
- The ability to interrupt the job by calling `promise.error()`.

### DIFF
--- a/src/lime/app/Future.hx
+++ b/src/lime/app/Future.hx
@@ -81,7 +81,7 @@ import lime.utils.Log;
 				var promise = new Promise<T>();
 				promise.future = this;
 
-				FutureWork.run(work, promise);
+				FutureWork.runSimpleJob(work, promise);
 			}
 			else
 			#end
@@ -308,7 +308,6 @@ import lime.utils.Log;
 	}
 }
 
-#if (lime_threads && !html5)
 /**
 	The class that handles asynchronous `work` functions passed to `new Future()`.
 **/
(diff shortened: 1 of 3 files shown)


player-03:
For convenience, here's the new function's documentation and signature:

```haxe
/**
	Runs the given function asynchronously, and resolves this `Promise` with
	the complete, error, and/or progress events sent by that function.
	Sample usage:
	```haxe
	function examplePromise():Future<String>
	{
		var promise = new Promise<String>();
		promise.completeAsync(function(state:State, output:WorkOutput):Void
			{
				output.sendProgress({progress:state.progress, total:10});
				state.progress++;
				if (state.progress == 10)
				{
					output.sendComplete("Done!");
				}
			},
			{progress: 0}, MULTI_THREADED);
		return promise.future;
	}
	var future = examplePromise();
	future.onComplete(function(message) { trace(message); });
	future.onProgress(function(loaded, total) { trace("Progress: " + loaded + ", " + total); });
	```
	@param doWork A function to perform work asynchronously. For best results,
	see the guidelines in the `ThreadPool` class overview.
	@param state The value to pass to `doWork`.
	@param mode Which mode to run the job in: `SINGLE_THREADED` or `MULTI_THREADED`.
**/
public function completeAsync(doWork:WorkFunction<State->WorkOutput->Void>, ?state:State, ?mode:ThreadMode = MULTI_THREADED):Void
```

Sword352:
I gave this a shot and would love this to be a new feature soon.

PR #1839, opened 2024-09-06
Simplify Timer.stop()

NeeEoo:
Simplified the `stop()` method in `haxe/Timer.hx` by removing redundant iterations and updated `NativeApplication.hx` to only run on active timers.


NeeEoo:
Ah thank you!

Issue #1515, opened 2022-01-09
Uncomfortable experiences while using AL.


Cheemsandfriends:
Hello, I wanted to comment that the AL lib has a small but kinda annoying problem.
When you unplug the headphones, for changing it to the PC's speakers, all sounds can't be reproduced.
It gives this error
![Error-Image](https://user-images.githubusercontent.com/84867412/148692204-3cafa7a9-6baa-4f2d-ab5d-4fa06f6dc50d.png)
And then you can't use the sounds anymore unless you restart the game via closing it and opening it again.
It would be great to fix this error for ppl who unplug the headphones.

ACrazyTown:
There was a discussion about this on the OpenAL Soft repository: https://github.com/kcat/openal-soft/issues/533, seems like it can be fixed with an OpenAL Soft extension?

PR #1841, opened 2024-09-23
Allow changing audio output device by updating OpenAL Soft

ACrazyTown:
This PR adds bindings and implements OpenAL Soft's [`ALC_SOFT_reopen_device`](https://[openal-soft.org/openal-extensions/SOFT_reopen_device.txt](https://openal-soft.org/openal-extensions/SOFT_reopen_device.txt)) and [`ALC_SOFT_system_events`](https://[openal-soft.org/openal-extensions/SOFT_system_events.txt](https://openal-soft.org/openal-extensions/SOFT_system_events.txt)) extensions. In order to take advantage of this functionality, I had to update OpenAL Soft, which requires using latest hxcpp (git, until there's a new stable release). New versions of OpenAL Soft require the C++17 standard and the current release of hxcpp doesn't seem to have an official way to use it. (Maybe it could be done via compilerflags in the Build.xml?)

With these changes, Lime will register a callback that gets called every time there's a new event related to audio devices. If the default audio device is changed, Lime will seamlessly switch all audio sources to play on the new audio device. As a side effect, this also resolves the issue where if the current audio device is disabled (eg. unplugged) all audio sources stop playing.

Closes https://github.com/openfl/lime/issues/1515, https://github.com/openfl/openfl/issues/2642


player-03:
> I've also had it working on Hashlink but ever since I've merged this change into 9.0.0-dev I keep getting a strange and seemingly unrelated error: `FATAL ERROR : Invalid signature for function lime@hl_cairo_get_matrix : PDOiOiB_iiii__OiOiB_iiii_ required but PDOdddddd__Odddddd_ found in hdll`

Try `lime rebuild hl` to bring your hdll up to date.

ACrazyTown:
> Try `lime rebuild hl` to bring your hdll up to date.

I've done that, tried doing a clean rebuild of both lime and my testing project. I keep getting the same error. Could this commit https://github.com/openfl/lime/commit/71442d09f4719e7fc48d98f7b5cd885865c95225 be related? It doesn't seem to be in 9.0 branch

player-03:
Oh yeah, if you don't have that merged, you'll need it. Edit: ok, merged into 9.0.0-Dev. You can pull from there.

ACrazyTown:
Okay, I think I did the merge right. Hashlink seems to be fixed and working fine now!

ACrazyTown:
I've gotten Mac to build and work!

I only have Linux, Android & Neko left to figure out, though CI for them will always fail because it uses an older hxcpp version

ACrazyTown:
Tested out Linux & it seems to work. I assume Android would also work with just an updated config but I don't know how I'd even test that... Neko seems to crash whenever it tries to dispatch the callback, I'm assuming it's threading related. I'm not sure what the alternative to hxcpp's `gc_set_top_of_stack` and hl's `hl_register_thread` would be on Neko.

ACrazyTown:
I'll undraft this PR now as I think it should be pretty much ready to go except for a few concerns:

1. Neko support - Neko is crashing when the callback is supposed to be dispatched, probably because of threading? I assume to fix it I'd use [neko_thread_register()](https://github.com/HaxeFoundation/neko/blob/master/vm/neko_vm.h#L42) but I can't figure out how to call functions from the Neko API in the C++ code as I can't seem to include the `neko_vm.h` header.

2. Callback userParam - Not sure how to tackle this. When setting a callback OpenAL Soft [keeps a reference to a user specified value](https://github.com/kcat/openal-soft/blob/cea3015d9d5d962a4bf780cfc80d332035aa453c/alc/events.cpp#L96). Seems like it's not possible to directly set a Haxe value here because it may get collected by the GC before the callback gets dispatched causing issues when sending it back as it would point to invalid memory.

3. ~~Failing Android CI - Android CI was failing on my fork because the compiler version it uses seems to have a bug. This was resolved in a newer commit of OpenAL Soft but it isn't in a stable release yet. Lime seems to only use stable releases so I don't know if its cool to use the commit with the fix~~ EDIT: This was fixed with https://github.com/openfl/lime/pull/1841/commits/4270cfde9deeea01be462010702bb2da316ea36d

PR #1843, opened 2024-09-28
Enhance frame rate calculations with double and SDL performance counter

MAJigsaw77:
* Updated frame rate calculations to improve timing accuracy.

* Replaced Uint32 with double for precise timing values.

* Switched from SDL_GetTicks() to SDL_GetPerformanceCounter() and SDL_GetPerformanceFrequency() for high-resolution timing.

### DIFF
--- a/project/src/backend/sdl/SDLApplication.cpp
+++ b/project/src/backend/sdl/SDLApplication.cpp
@@ -129,7 +129,7 @@ namespace lime {
 
 				if (!inBackground) {
 
-					currentUpdate = SDL_GetTicks ();
+					currentUpdate = ((double) SDL_GetPerformanceCounter () / (double) SDL_GetPerformanceFrequency ()) * 1000.0;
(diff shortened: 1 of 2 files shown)


NeeEoo:
Just Fyi: SDL_AddTimer takes a uint32 (ms) and not a double, so this still causes the framerate to be slightly off. A possible fix would be using std::chrono

dimensionscape:
I appreciate the initiative, but this doesn't resolve any problems, and increases cpu usage. We are going to handle this in a very different way in the next(probably) version of lime.

NeeEoo:
When sdl 3 is implemented we can use SDL_GetTicksNS to get nanoseconds instead and its in Uint64

dimensionscape:
The plan is to stop using SDL to construct the main loop.

> When sdl 3 is implemented we can use SDL_GetTicksNS to get nanoseconds instead and its in Uint64

PR #1852, opened 2024-10-22
Work around ThreadPool error when targeting HashLink in Haxe 4.0-4.1.

### LINKED ISSUE #1851: ThreadPool problems reported by Haxe 4.0 and 4.1 when targeting HashLink
Problems reported by Haxe 4.0.5 for HashLink only:

```
lime/8,2,0/src/lime/system/ThreadPool.hx:428: characters 7-36 : This kind of operation is not supported
lime/8,2,0/src/lime/system/ThreadPool.hx:524: characters 6-28 : This kind of operation is not supported
```

Problems reported by Haxe 4.1.5 for HashLink only:

```
lime/8,2,0/src/lime/system/ThreadPool.hx:428: characters 7-36 : Cannot add sys.thread.Tls.T and Int
lime/8,2,0/src/lime/system/ThreadPool.hx:524: characters 6-28 : Cannot add sys.thread.Tls.T and Int
```

Haxe 4.2.5 and 4.3.6 report no problems for HashLink. I haven't tested all combinations, but it appears that all Haxe 4 versions report no errors when targeting Neko or cpp.

player-03:
Fixes #1851, at least for Haxe 4.0.5. I couldn't reproduce the issue in Haxe 4.1.5, possibly because I'm using Linux rather than Mac. But since it pointed to the same two lines, I think it's safe to assume it's the same bug.

### DIFF
--- a/src/lime/system/ThreadPool.hx
+++ b/src/lime/system/ThreadPool.hx
@@ -425,7 +425,7 @@ class ThreadPool extends WorkOutput
 				{
 					while (!output.__jobComplete.value && (interruption = Thread.readMessage(false)) == null)
 					{
-						output.workIterations.value++;
+						output.workIterations.value = output.workIterations.value + 1;
 						event.job.doWork.dispatch(event.job.state, output);
 					}
 				}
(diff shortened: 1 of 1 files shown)


joshtynjala:
I confirm that this fixes the issue for both Haxe 4.0.5 and 4.1.5 on my machine.

PR #1880, opened 2024-12-27
Fix iOS status bar dimming.

MAJigsaw77:
When using `SDL_WINDOW_FULLSCREEN_DESKTOP` on iOS, the status bar dimming was not functioning correctly.

**Before:**

IMG_0010

**After:**

IMG_0009

(Sorry for the color difference in the images, but the status bar in the second image is clearly dimmed compared to the first.)

This change resolves the issue and restores the expected dimming behavior.

### DIFF
--- a/project/src/backend/sdl/SDLWindow.cpp
+++ b/project/src/backend/sdl/SDLWindow.cpp
@@ -28,8 +28,9 @@ namespace lime {
 	SDL_Cursor* SDLCursor::waitCursor = 0;
 	SDL_Cursor* SDLCursor::waitArrowCursor = 0;
 
+	#if !defined (IPHONE) && !defined (APPLETV)
 	static bool displayModeSet = false;
-
+	#endif
 
 	SDLWindow::SDLWindow (Application* application, int width, int height, int flags, const char* title) {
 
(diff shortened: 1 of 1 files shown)


player-03:
That's a lot of conditionals for such a small thing. Are we sure there isn't a simpler way to handle it?

MAJigsaw77:
> That's a lot of conditionals for such a small thing. Are we sure there isn't a simpler way to handle it?

This worked so i don't think there is an easier way.

MAJigsaw77:
> That's a lot of conditionals for such a small thing. Are we sure there isn't a simpler way to handle it?

This is the easiest way i could make it work, if you can make it be better, feel free to do so!

player-03:
I've taken a stab at simplifying. Got down to only one `#if`, which will make future changes easier if they prove necessary.

MAJigsaw77:
Tbh, I'm actually wondering if android needs the `SDL_WINDOW_FULLSCREEN_DESKTOP` , it isn't doing anything in android as well.

player-03:
Fair question.

MAJigsaw77:
> Fair question.

After checking SDL2's source, there's nothing pointing that android uses `SDL_WINDOW_FULLSCREEN_DESKTOP` in any way, so ig it is fine to just remove it from android as well.

dimensionscape:
is this fix irrelevant now?

MAJigsaw77:
> is this fix irrelevant now?

This pr didn't fix the actual issue as i seen so it was worthless keeping it

PR #1885, opened 2025-01-08
Bumped Android SDK to 34, Google Play won't accept lower. Fixed some typos. Set ios.non-exempt-encryption to false by default.

pozirk:


pozirk:
`ios.non-exempt-encryption` is set to `true` by default, so I had to change `ITSAppUsesNonExemptEncryption` (at {{app.file}}-Info.plist) manually to `false`, because otherwise Apple would not approve my app.

> Set the value for this key to NO in your app’s Information Property List file to indicate that your app—including any third-party libraries you link against—either uses no encryption, or only uses encryption that’s exempt from export compliance requirements, as described in Determine your export compliance requirements. Set the value to YES to indicate that your app uses non-exempt encryption.

From here: https://developer.apple.com/documentation/bundleresources/information-property-list/itsappusesnonexemptencryption

I guess, `false` by default makes mo sense.

ghost (changes requested):
gradle and gradle plugin version needs to be changed to work with targetSdk 34

pozirk:
Hm, I've been manually setting SDK to 34 for some time already, and haven't had any problems with gradle.

ghost:
gradle does a warning that complains about target sdk u have to look

joshtynjala:
Can you split Android and iOS changes into two separate PRs, please? Thanks!

PR #1888, opened 2025-01-08
Bumped Android SDK to 34, Google Play won't accept lower.

pozirk:
There is the following warning when bulding the project, but I had no problems so far.
> WARNING:We recommend using a newer Android Gradle plugin to use compileSdk = 34
> 
> This Android Gradle plugin (7.3.1) was tested up to compileSdk = 33
> 
> This warning can be suppressed by adding
>     android.suppressUnsupportedCompileSdk=34
> to this project's gradle.properties
> 
> The build will continue, but you are strongly encouraged to update your project to
> use a newer Android Gradle Plugin that has been tested with compileSdk = 34

If gradle can be upgraded, would be great.

### DIFF
--- a/tools/platforms/AndroidPlatform.hx
+++ b/tools/platforms/AndroidPlatform.hx
@@ -465,7 +465,7 @@ class AndroidPlatform extends PlatformTarget
 		context.OUTPUT_DIR = targetDirectory;
 		context.ANDROID_INSTALL_LOCATION = project.config.getString("android.install-location", "auto");
 		context.ANDROID_MINIMUM_SDK_VERSION = project.config.getInt("android.minimum-sdk-version", 21);
-		context.ANDROID_TARGET_SDK_VERSION = project.config.getInt("android.target-sdk-version", 30);
+		context.ANDROID_TARGET_SDK_VERSION = project.config.getInt("android.target-sdk-version", 34);
 		context.ANDROID_EXTENSIONS = project.config.getArrayString("android.extension");
 		context.ANDROID_PERMISSIONS = project.config.getArrayString("android.permission", [
 			"android.permission.WAKE_LOCK",


ghost (changes requested):
You may need change the gradle ver with `8.0` and plugin ver with `8.1.1`.

joshtynjala:
It looks like the minimum `android.gradle-plugin` version that supports `android.target-sdk-version` 34 without a warning is 8.1.1.

This requires `android.gradle-version` to be bumped to 8.0 or higher (I found that `android.gradle-version` 8.2 is the minimum that works for me without crashing on macOS).

`android.gradle-plugin` 8.1 says that it requires Java JDK 17, and the previous version of `android.gradle-plugin` that we were using required JDK 11.

So that's a lot of major version bumps. I'm feeling somewhat inclined to wait until Lime 8.3 to make these our new defaults, instead of a bugfix release for Lime 8.2.x.

Waiting feels unfortunate when Google Play strictly requires `android.target-sdk-version` 34 minimum. However, since folks can opt into it now in their project.xml, it's also not an absolute blocker either.

```xml
<config:android target-sdk-version="34" gradle-version="8.2" gradle-plugin="8.1.1"/>
```

ghost:
what about we make gradle version `8.9` and gradle plugin version `8.7.1`?, they work with 35 and it'd do a future patch aswell

pozirk:
Supporting SDK 35 is a great idea.
Google will eventually start pushing everyone to use it.

At the moment I'm building my projects using JDK 17 on Windows and have no problems.

Btw, forgot to put the link to actual SDK 34 requirement by Google:
> Starting August 31 2024:
> New apps and app updates must target Android 14 (API level 34) or higher to be submitted to Google Play; except for Wear OS and Android TV apps, which must target Android 13 (API level 33) or higher.

https://developer.android.com/google/play/requirements/target-sdk

joshtynjala:
Yeah, I guess it makes sense to support 35 with further Gradle upgrades.

Doing some testing now. It looks like I'll need to fix another new warning:

> [CXX1100] android.ndkVersion is [27.0.12079973] but android.ndkPath /path/to/25.2.9519653 refers to a different version [25.2.9519653]

It appears that our build.gradle files will need to specify both ndkPath and ndkVersion. It appears that the value for ndkVersion can be extracted from source.properties in the NDK.

joshtynjala:
Committed with additional changes in the 8.3.0-Dev branch:

2e578d3e569dd369bba8d86933486485f62948c6

PR #1890, opened 2025-01-12
create variables for window parameters

zebbd:
this will simplify the implementation of some settings like vsync switching

### DIFF
--- a/templates/haxe/ApplicationMain.hx
+++ b/templates/haxe/ApplicationMain.hx
@@ -7,6 +7,29 @@ import ::APP_MAIN::;
 
 @:dox(hide) class ApplicationMain
 {
+	private var _allowHighDPI = ::allowHighDPI::;
+	private var _alwaysOnTop = ::alwaysOnTop::;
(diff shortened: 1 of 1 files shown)


player-03:
Where would these variables be used? It looks like they're only available inside `ApplicationMain`, which seems like the wrong place to implement new features.

zebbd:
> Where would these variables be used? It looks like they're only available inside `ApplicationMain`, which seems like the wrong place to implement new features.

you can change them with macros

player-03:
Technically you can already change them with macros, just a bit less easily.

More importantly, macros run at compile time, and you already have a perfectly good way to set vsync at compile time: `<window vsync="false" />`. What's the benefit of using macros instead of project.xml?

joshtynjala:
> you already have a perfectly good way to set vsync at compile time: `<window vsync="false" />`

And if you don't want to modify project.xml for a temporary change, you can override vsync on the command line too.

```sh
lime build windows --window-vsync=false
```

PR #1896, opened 2025-01-22
Add AudioSource.latency

ACrazyTown:
This PR adds a `latency` field to `AudioSource` that returns the estimated playback latency of the sound.

On native targets (that use OpenAL Soft) the same method from https://github.com/openfl/lime/pull/1759 is used, although exposing it as a seperate field instead of applying it directly to the sound's time.

HTML5 uses [`AudioContext.baseLatency`](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/baseLatency) and [`AudioContext.outputLatency`](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/outputLatency)

In the case where it's not possible to query the playback latency (eg. Flash) 0 is returned.


Raltyro:
Wasnt this in ninjamuffin pr?
also i approve

Raltyro (approved):

ghost (changes requested):
iOS will give null since it uses `mojoal` instead if `OpenAL SOFT`.

MAJigsaw77:
> iOS will give null since it uses `mojoal` instead if `OpenAL SOFT`.

@ACrazyTown Since some platforms might use mojoAl or something else for example, I'd suggest checking if the extension is supported first `AL.isExtensionPresent("AL_SOFT_source_latency")`.

ACrazyTown:
@mcagabe19 @MAJigsaw77 The native C++ functions will return null if `LIME_OPENALSOFT` is not defined. This is later checked on the Haxe side: https://github.com/openfl/lime/blob/3f9041ab7d63c7c1cbc397dc482dd8cd2a0e7401/src/lime/_internal/backend/native/NativeAudioSource.hx#L592-L599
Is that not enough?

MAJigsaw77:
> @mcagabe19 @MAJigsaw77 The native C++ functions will return null if `LIME_OPENALSOFT` is not defined. This is later checked on the Haxe side:
> 
> https://github.com/openfl/lime/blob/3f9041ab7d63c7c1cbc397dc482dd8cd2a0e7401/src/lime/_internal/backend/native/NativeAudioSource.hx#L592-L599
> 
> 
> Is that not enough?

It's not because openal soft can be there, but the platform might not support latency, in any case that check should make it fine if added.

ghost (approved):

ACrazyTown:
If it is problematic to add this to the AudioSource API (as Lime uses 3 different OpenAL backends depending on the configuration and only OpenAL Soft can query latency) would it be better to just reduce this to binding the necessary functions and exposing them in Haxe thru OpenALAudioContext or whatever? If so, would be nice to also review https://github.com/openfl/lime/pull/1912 so the backend specific functions can be hidden behind a proper conditional

project/src/media/openal/OpenALBindings.cpp
```diff
@@ -2104,6 +2104,44 @@ namespace lime {
 	}
 
 
+	value lime_al_get_sourcedv_soft (value source, int param, int count) {
+
+		#ifdef LIME_OPENALSOFT
```
player-03, line 2109: We were discussing this on Discord, and came up with the idea of checking `#ifdef AL_SOFT_source_latency` instead of `#ifdef LIME_OPENALSOFT`. If it works, it would be a more reliable check. Can you test it out and report back?

player-03: We could also try `if (alGetSourcedvSOFT == NULL) return alloc_null();`, as a backup.

src/lime/media/openal/AL.hx
```diff
@@ -969,6 +972,24 @@ class AL
 		#end
 	}
 
+	public static function getSourcedvSOFT(source:ALSource, param:Int, count:Int = 1):Array<Float>
+	{
+		#if (lime_cffi && lime_openalsoft && !macro)
+		var result = NativeCFFI.lime_al_get_sourcedv_soft(source, param, count);
+		#if hl
+		if (result == null) return [];
```
player-03, line 980: I think for consistency, it should always return the same value on failure. Since all the other fail states return null, this should too.

src/lime/_internal/backend/native/NativeAudioSource.hx
```diff
@@ -588,4 +599,20 @@ class NativeAudioSource
 
 		return position;
 	}
+
+	public function getLatency():Float
+	{
+		#if lime_openalsoft
```
player-03, line 605: If all the checks work right on the C++ side, we ought to be able to remove all the `#if lime_openalsoft` checks on the Haxe side. That way, you can't accidentally break things by setting different flags than lime.ndll was built with.

PR #1898, opened 2025-01-26
Fixed assets embedding on multiple targets.

MAJigsaw77:
The following changes fix the embedding of assets on `Android`, `iOS`, `tvOS`, and `WebAssembly`.


joshtynjala:
Can you attach a simple project where things doen't work correctly without this change? Thanks!

MAJigsaw77:
> Can you attach a simple project where things doen't work correctly without this change? Thanks!

I've used SimpleImage sample and modified `project.xml` to embed the `Assets` folder.
```xml
<project>
	
	<meta title="Simple Image" package="org.openfl.lime.samples.simpleimage" version="1.0.0" company="OpenFL" />
	<app main="Main" path="Export" file="SimpleImage" />
	<architecture name="arm64" exclude="armv7" if="android" />
	<source path="Source" />
	
	<haxelib name="lime" />
	
	<assets path="Assets" rename="assets" embed="true" />

</project>
```

If i look within the apk file it clearly shows that the file hasnt been embeded at all.

{9B908D19-1F70-472C-B806-389EFA7ABCCB}

But after the fix is applied, it shows that the file has been successfully been embedded without it being left there.

{0C5BD76A-886B-4E50-8787-5034A8215B5D}

This is valid for the other targets aswell.

MAJigsaw77:
> > Can you attach a simple project where things doen't work correctly without this change? Thanks!
> 
> I've used SimpleImage sample and modified `project.xml` to embed the `Assets` folder.
> 
> ```
> <?xml version="1.0" encoding="utf-8"?>
> <project>
> 	
> 	<meta title="Simple Image" package="org.openfl.lime.samples.simpleimage" version="1.0.0" company="OpenFL" />
> 	<app main="Main" path="Export" file="SimpleImage" />
> 	<architecture name="arm64" exclude="armv7" if="android" />
> 	<source path="Source" />
> 	
> 	<haxelib name="lime" />
> 	
> 	<assets path="Assets" rename="assets" embed="true" />
> 
> </project>
> ```
> 
> If i look within the apk file it clearly shows that the file hasnt been embeded at all.
> 
> {9B908D19-1F70-472C-B806-389EFA7ABCCB}
> 
> But after the fix is applied, it shows that the file has been successfully been embedded without it being left there.
> 
> {0C5BD76A-886B-4E50-8787-5034A8215B5D}
> 
> This is valid for the other targets aswell.

@joshtynjala Can i get some feedback on this as of now?

joshtynjala:
It seems likely the correct change to make.

 owever, I can't merge or provide any more detailed feedback until I have time to test and compare how `embed="false"` and `embed="true"` behave on various targets (including windows/mac/linux and air) to be sure that things are generally consistent. I don't usually have any reason to embed assets versus preloading them, so it'll require a bit to time to ensure that I'm up to speed on how it should work (beyond my assumptions based on my experience with Flash/AIR "embedded" assets).

player-03:
Reviewing the other platforms:

- It appears Windows, Mac, and Linux all already did this.
- HTML5 has it commented out, maybe because embedding isn't an option there? I'm not going to touch that.
- Flash and Air check `asset.embed == false` instead of `asset.embed != true`. I think this means that assets are embedded by default on those targets. Not sure if that's correct.
- Tizen also seems to have the incorrect behavior.

There's a _lot_ of duplicate code here. I wonder if it's time to make a function for it...

MAJigsaw77:
> Reviewing the other platforms:
> 
> * It appears Windows, Mac, and Linux all already did this.
> * HTML5 has it commented out, maybe because embedding isn't an option there? I'm not going to touch that.
> * Flash and Air check `asset.embed == false` instead of `asset.embed != true`. I think this means that assets are embedded by default on those targets. Not sure if that's correct.
> * Tizen also seems to have the incorrect behavior.
> 
> There's a _lot_ of duplicate code here. I wonder if it's time to make a function for it...

Perhaps yeah

PR #1907, opened 2025-02-14
Fix #1906 GetDirectory UTF16 encoding issue

### LINKED ISSUE #1906: SDLSystem GetDirectory does not handle UTF16 correctly on Windows
**Repro Steps:**

- Create a user with the following username `øÜêþЯЂў`
- Use GetDirectory with SystemDirectory.DESKTOP, SystemDirectory.DOCUMENTS, SystemDirectory.FONTS or SystemDirectory.USER within a game while running in that user account
- Notice crash

PR incoming with fix.

tomcashman:
Note: I've intentionally used `SHGetFolderPathW` for compatibility with older systems. Vista + later alias this to `SHGetKnownFolderPath`.

### DIFF
--- a/project/src/backend/sdl/SDLSystem.cpp
+++ b/project/src/backend/sdl/SDLSystem.cpp
@@ -142,11 +142,9 @@ namespace lime {
 
 				#elif defined (HX_WINDOWS)
 
-				char folderPath[MAX_PATH] = "";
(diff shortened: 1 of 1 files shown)


tobil4sk (approved):
This looks good to me. According to the microsoft docs:

> New applications should always call the Unicode versions. Many world languages require Unicode. If you use ANSI strings, it will be impossible to localize your application.

https://learn.microsoft.com/en-us/windows/win32/learnwin32/working-with-strings

So this seems like the correct approach

joshtynjala:
@tobil4sk Thanks for taking a look. Merged!

PR #1910, opened 2025-02-20
Implement Window.setVSyncMode()

### LINKED ISSUE #1908: Enhancement: Add an option to toggle VSync at runtime, and add an option for adaptive VSync
Currently, VSync is enabled/disable via a window flag, and can currently only be set via a flag in the project settings.

However, from what I've read, there should be no issue with adding a function to allow the application to change the VSync mode.

This would require making this call into its own function, then making that accessible on the Window object (either via a function or a property variable).

https://github.com/openfl/lime/blob/99ca58d06f2c27149850365e12cf67c425356dc0/project/src/backend/sdl/SDLWindow.cpp#L222-L230

Adding this feature would enable applications to provide a configuration option for users to enable or disable VSync.

Additionally, while we're at it, we should also add the option to use Adaptive VSync. This is done by passing `-1` to the swap interval function, and returns an error if it is not supported (see [the SDL documentation](https://wiki.libsdl.org/SDL2/SDL_GL_SetSwapInterval)). In this case, the option should probably be an enumeration rather than a boolean value.

EliteMasterEric:
Resolves #1908 by adding a function to set the VSync mode after the window has been initialized.

Includes the ability to set the mode to Adaptive VSync, which disables VSync if the frame rate drops below the monitor's refresh rate.

The function returns `false` if the VSync mode failed to be applied properly (i.e. attempting to use Adaptive VSync on a device which does not support it).

### DIFF
--- a/project/include/ui/Window.h
+++ b/project/include/ui/Window.h
@@ -16,6 +16,13 @@
 
 namespace lime {
 
+	enum WindowVSyncMode {
+
+		WINDOW_VSYNC_ADAPTIVE = -1,
+		WINDOW_VSYNC_OFF = 0,
+		WINDOW_VSYNC_ON = 1,
+
+	};
 
 	class Window {
 
@@ -25,6 +32,7 @@ namespace lime {
 			virtual ~Window () {};
 
 			virtual void Alert (const char* message, const char* title) = 0;
+			virtual bool SetVSyncMode (WindowVSyncMode mode) = 0;
 			virtual void Close () = 0;
 			virtual void ContextFlip () = 0;
 			virtual void* ContextLock (bool useCFFIValue) = 0;
@@ -97,7 +105,6 @@ namespace lime {
 		WINDOW_FLAG_COLOR_DEPTH_32_BIT = 0x00010000
 
 	};
-
 }
 
 
--- a/project/src/ExternalInterface.cpp
+++ b/project/src/ExternalInterface.cpp
@@ -3158,6 +3158,22 @@ namespace lime {
 	}
 
 
+	bool lime_window_set_vsync_mode (value window, int mode) {
+
+		Window* targetWindow = (Window*)val_data (window);
+		return targetWindow->SetVSyncMode (mode);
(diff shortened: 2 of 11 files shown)


dimensionscape:
I am going to handle this a bit differently, but I appreciate the effort!

Raltyro:
Still waiting for this to happen...

player-03:
@dimensionscape Could you explain how you'd implement this?

dimensionscape:
Sorry, I should have been more clear, but this overlaps with some pending  changes that involve vsync and the main loop making this merge conflicting and unnecessary. Sorry for being so slow on this.

Raltyro:
So there has been a changes already made for VSync? But why it hasn't been merged to main?

dimensionscape:
@player-03 

Changing VSync at runtime in Lime introduces a number of footguns because several assumptions inside the current main loop and the backend integration don’t safely support arbitrary toggling. Even with a boolean return value, the behavior can still be misleading.

Another issue is that bolting this onto `Window` is conceptually misleading. In Lime, VSync is effectively application-global, not per-window, so exposing it on the window API creates confusion about what can actually happen under the hood.

I’m not opposed to exposing a runtime VSync API eventually, but the internal plumbing needs to change first. VSync should be handled as an explicit and safe request, not a property or a direct command, and not until the frame-sync and main-loop internals are reorganized to support it cleanly.

Right now, we avoid most problems by restricting VSync configuration to compile-time or window-creation. That’s not ideal, but it prevents inconsistent states that would otherwise be possible with ad-hoc toggling.

So I closed this PR for two reasons:

It conflicts with ongoing architectural changes involving the main loop and VSync behavior.

Merging it now risks introducing an API that will quickly become obsolete or constrain decisions we need to make at a lower level.

I should have been more verbose instead of just “fly-swatting” the PR. AGAIN! Sorry about that.

@Raltyro 

Not exactly, but it is part is part of a larger, ongoing update being prepared for Lime 9 that is WIP.

PR #1911, opened 2025-02-25
Fix set_title in Window.hx for the legacy branch

Digivorix:
This PR corrects the parameter being passed to backend.setTitle() inside of the set_title function in ui/Window.hx, which allows the window title to be set at runtime. Currently, the existing window title is passed to backend.setTitle() when the set_title function is run in Window.hx, resulting in no change.


dimensionscape:
I apologize, but we are not providing support or maintenance for the legacy branch.

PR #1912, opened 2025-02-28
Add seperate defines for OpenAL Soft and MojoAL

ACrazyTown:
Lime uses different OpenAL backends based on the target platform and build configuration. Currently there's only one define (`lime_openal`) that only lets us determine if any OpenAL backend is being used, but not what which one specifically.

### DIFF
--- a/include.xml
+++ b/include.xml
@@ -19,7 +19,7 @@
 	<define name="lime-howlerjs" if="html5" />
 	<define name="lime-html5" if="html5" />
 	<define name="lime-native" if="native" />
-	<define name="lime-openalsoft" if="windows || linux || mac || android" unless="static_link || lime-mojoal" />
+	<define name="lime-openalsoft" if="windows || linux || mac || android" unless="static_link || lime-mojoal || winrt" />
 	<define name="lime-mojoal" if="lime-switch || static_link || winrt" unless="lime-openalsoft" />
 	<define name="lime-openal" if="ios || tvos || emscripten || lime-openalsoft || lime-mojoal" />
 	<define name="lime-opengl" if="desktop" unless="html5" />
--- a/include.xml
+++ b/include.xml
@@ -19,7 +19,9 @@
 	<define name="lime-howlerjs" if="html5" />
 	<define name="lime-html5" if="html5" />
 	<define name="lime-native" if="native" />
-	<define name="lime-openal" unless="lime-console || flash || html5" />
+	<define name="lime-openalsoft" if="windows || linux || mac || android" unless="static_link || lime-mojoal" />
+	<define name="lime-mojoal" if="lime-switch || static_link || winrt" unless="lime-openalsoft" />
+	<define name="lime-openal" if="ios || tvos || emscripten || lime-openalsoft || lime-mojoal" />
 	<define name="lime-opengl" if="desktop" unless="html5" />
 	<define name="lime-opengles" if="emscripten || mobile" />
 	<define name="lime-vorbis" if="native" />
--- a/project/Build.xml
+++ b/project/Build.xml
@@ -21,12 +21,9 @@
 	<set name="LIME_MBEDTLS" value="1" unless="emscripten || winrt" />
 	<!-- <set name="LIME_NEKO" value="1" if="linux" /> -->
 	<set name="LIME_OGG" value="1" />
-	<set name="LIME_OPENALSOFT" value="1" if="windows || linux || mac || android" unless="static_link" />
-	<set name="LIME_OPENAL" value="1" if="iphone || webassembly || tvos" />
-	<set name="LIME_MOJOAL" value="1" if="switch || static_link || winrt || mojoal" unless="LIME_OPENAL" />
-	<unset name="LIME_OPENALSOFT" if="LIME_MOJOAL" />
-	<set name="LIME_OPENAL" value="1" if="LIME_OPENALSOFT" />
-	<set name="LIME_OPENAL" value="1" if="LIME_MOJOAL" />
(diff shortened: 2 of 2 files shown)


include.xml
```diff
@@ -20,6 +20,8 @@
 	<define name="lime-html5" if="html5" />
 	<define name="lime-native" if="native" />
 	<define name="lime-openal" unless="lime-console || flash || html5" />
+	<define name="lime-mojoal" if="lime-openal ios || tvos || static_link" />
```
dimensionscape: was this meant to be:

<define name="lime-mojoal" if="lime-openal || ios || tvos || static_link" / ?

Either way, I'm not sure that is right

ACrazyTown: It was supposed to be `if lime-openal AND ios OR tvos OR static_link` but I guess that wouldn't work. I tried to simplify it by simply swapping the two in a new commit

EDIT: Wait no this also wouldn't work because it would define either `lime-openalsoft` or `lime-mojoal`, even though neither of them should be defined when targetting ios

player-03:
I think I'd rather use a single define, with a value. That way, there can only be one set at a time, because setting a new value overwrites the first.

Also, the define should actually determine which backend gets used, otherwise there's no guarantee that it's correct.

ACrazyTown:
> I think I'd rather use a single define, with a value. That way, there can only be one set at a time, because setting a new value overwrites the first.

Yeah this could work too. My initial thought was that since both MojoAL and OpenAL Soft implement the core OpenAL API, the `lime_openal` define should be present always (also wouldn't break already existing code). However, MojoAL and OpenAL Soft might have some exclusive features that should be locked behind their respective conditionals.

> Also, the define should actually determine which backend gets used, otherwise there's no guarantee that it's correct.

I'm not sure how to do this. I thought that the backend was decided when the target's NDLL is built. How would the backend be modified with a define if the necessary native code was not included?

player-03:
> My initial thought was that since both MojoAL and OpenAL Soft implement the core OpenAL API, the `lime_openal` define should be present always (also wouldn't break already existing code). However, MojoAL and OpenAL Soft might have some exclusive features that should be locked behind their respective conditionals.

I don't think that would be a problem, you can just do this.

```haxe
#if lime_openal
//Some version of OpenAL is present.
#end

#if (lime_openal == "mojoal")
//The backend is MojoAL.
#end

#if (lime_openal == "soft")
//The backend is OpenAL-Soft.
#end
```

> I'm not sure how to do this. I thought that the backend was decided when the target's NDLL is built.

Oh, right, those are in the NDLL. But that's ok, because [defines from _include.xml_ are passed through to _Build.xml_](https://github.com/openfl/lime/blob/develop/src/lime/tools/CPPHelper.hx#L245), so we should still be able to access them at rebuild time.

> How would the backend be modified with a define if the necessary native code was not included?

To clarify: I mean that the same define should control both the native code and the Haxe code. After modifying the define in _include.xml_, you'd have to rebuild the NDLL, but that's perfectly fine. "The NDLL is out of sync" is a common problem that we know how to troubleshoot.

ACrazyTown:
> To clarify: I mean that the same define should control both the native code and the Haxe code. After modifying the define in _include.xml_, you'd have to rebuild the NDLL, but that's perfectly fine. "The NDLL is out of sync" is a common problem that we know how to troubleshoot.

Sorry, I'm a bit confused. If I understood it right, we'd have decide which target defaults to what OpenAL backend in `include.xml` and then in `Build.xml` decide which native library to build based on the value of the `lime_openal` define, right? If so, how would we compare the value of the define in the `Build.xml`? I looked through the documentation and unless I missed something obvious it seems like you can only check if the define is present and not what it is.

Also, would you happen to know if there's a part of Lime or some other hxcpp library that does something similar? I'm still wrapping my head around all these XMLs so would be helpful to have an example to base things off of

player-03:
> I looked through the documentation and unless I missed something obvious it seems like you can only check if the define is present and not what it is.

You're right, _Build.xml_ can only check whether or not it's defined. I guess we can stick with the multiple defines approach, then.

player-03 (approved):

include.xml
```diff
@@ -19,7 +19,9 @@
 	<define name="lime-howlerjs" if="html5" />
 	<define name="lime-html5" if="html5" />
 	<define name="lime-native" if="native" />
-	<define name="lime-openal" unless="lime-console || flash || html5" />
+	<define name="lime-openalsoft" if="windows || linux || mac || android" unless="static_link || lime-mojoal" />
+	<define name="lime-mojoal" if="lime-switch || static_link || winrt" unless="lime-openalsoft" />
```
player-03, line 23: Is winrt being handled differently now? Before, it would set `LIME_OPENALSOFT` because `windows`, then set `LIME_MOJOAL` because `winrt`, then unset `LIME_OPENALSOFT` because `LIME_MOJOAL`. Now, it sets `lime-openalsoft` because `windows`, then that prevents it setting `lime-mojoal`.

ACrazyTown: Oversight on my part, I think it should be good to go now although I can't try it out since running `lime rebuild winrt` results in C++ compiler errors.

tobil4sk:
Unfortunately this has introduced regressions because in some cases include.xml doesn't get parsed, for example if you run `lime rebuild linux` outside of the lime directory. This results in a lime.ndll without any openal symbols.

We also discussed on discord not long ago, and this causes some problems because we don't necessarily know that the ndll used was built with the default AL implementation, because it's possible to force the use of mojoal.

It would be better to write backend-agnostic haxe code and then rely on extension tests to see if individual extensions are supported at runtime. Haxe should only be aware of the generic functions, and it shouldn't know details about specific openal implementations.

Issue #1569, opened 2022-07-26
SDL_GetTicks Noticeably mprecise


steviegt6:
`SDL_GetTicks` is only accurate to the millisecond, which results in `deltaTime` values being rather imprecise. This is immediately observable in the FPS display, which is inaccurate by a couple frames:
![image](https://user-images.githubusercontent.com/27323911/180980468-9ae3653c-9796-4623-9979-d19f551b2555.png)

A piece of the relevant code (`SDL_GetTicks` is used in several relevant places in this file):
https://github.com/openfl/lime/blob/862fe55b1cf567bdfb4ada7d5bc78be36353f9e6/project/src/backend/sdl/SDLApplication.cpp#L132

EyeDaleHim:
How interesting, I might fix this in a private repository of mine

Dimensionscape:
I don't think the delta accuracy here causes any major issues other than the current FPS being calculated incorrectly when using the FPS utility api, which has no bearing on anything really. This is actually easy to solve and something that has been bugging me so I suppose I'll get it fixed here pretty soon.

Dimensionscape:
I thought this only affected the frame rate counter, but the event loop delta is actually wrong due to this. Perhaps we should implement the timing for the main loop like Starling and use a float. I will look into this more.

joshtynjala:
> I will look into this more.

Sounds good. I hope that you'll wait to commit until after OpenFL 9.2 is released. This sounds like something that should be in dev builds for a while, just to ensure that there aren't any unintended side effects.

Dimensionscape:
No worries there. The event loop is a critical, core component obviously and any changes should be carefully considered and well tested. There's not enough time to even attempt to fit it into the next release even if I wanted to try. 😂

Cheemsandfriends:
Searching up a bit through SDL's shenanigans, I found this website detailing an alternative to `SDL_GetTicks`

[`SDL_GetPerformanceCounter` with `SDL_GetPerformanceFrequency`](http://seanballais.com/blog/computing-elapsed-time-in-sdl2/)

This could help with the issue facing right now? The website's example seems to be putting the stuff in seconds, but maybe we could fix that by multiplying it to 1000?

dimensionscape:
> Searching up a bit through SDL's shenanigans, I found this website detailing an alternative to `SDL_GetTicks`
> 
[...]

PR #1915, opened 2025-03-06
SDL3 Support

### LINKED ISSUE #1848: Migrate Lime on native platforms to use SDL3
SDL just released [version 3.1.3 stable API preview](https://github.com/libsdl-org/SDL/releases/tag/preview-3.1.3), which means that (while the library still has bugs and issues to resolve before the 3.2 stable release), no existing functions will be removed, making it stable enough for migration.

There is a very long list of useful features that migrating would enable (see the list at the link), and also a long list of issues that are resolved by updating (such as #1569), but this first requires migrating the existing places that SDL2 are used.

I originally attempted this myself by just editing the existing implementation last year, but this resulted in a bunch of graphical glitches, so updating would probably require someone more experienced with C++ and graphics APIs.

Fancy2209:
Current work to get Lime working on SDL3

Fixes: #1848


ghost:
the thing is sdl module needs to switched with sdl3 and vulkan implementation needed in lime and SDL android java codes needs to be updated too

Fancy2209:
> the thing is sdl module needs to switched with sdl3 and vulkan implementation needed in lime and SDL android java codes needs to be updated too

I did update the module to SDL3.
SDL still supports OpenGL, no need for Vulkan.
I will update Android Java when I get this working on desktoo

ghost:
> I did update the module to SDL3.

I meant to say SDL2 needs to be removed from the lime and fully switched with SDL3

Fancy2209:
> > I did update the module to SDL3.
> 
> I meant to say SDL2 needs to be removed from the lime and fully switched with SDL3

It does not? I don't see why SDL2 shouldn't be an option.

EliteMasterEric:
Interesting that the proposal is to have both available

Maybe the SDL2 can be available, and used by default, and deprecated later as Lime's SDL3 implementation is proven to be stable.

Fancy2209:
> Interesting that the proposal is to have both available
> 
> Maybe the SDL2 can be available, and used by default, and deprecated later as Lime's SDL3 implementation is proven to be stable.

It was mostly because mojoAL doesn't have a SDL3 version yet, and from what I was reading on the build files it's used for Audio on the Switch

ACrazyTown:
It'a also used instead of OpenAL Soft when static linking (due to licensing conflicts) I believe

Fancy2209:
I accidentally pulled develop into tbe branch, not sure what to do about that.

ACrazyTown:
What's the status of this?

Fancy2209:
> What's the status of this?

Builds on Linux, doesn't build on everything else,  I was having issues with asset manifests due to Haxe's JSON parser locally

project/src/backend/sdl3/SDLSystem.cpp
```diff
@@ -0,0 +1,997 @@
...
+
+
+	size_t fread (void *ptr, size_t size, size_t count, FILE_HANDLE *stream) {
+
+		size_t nmem;
+		System::GCEnterBlocking ();
+
+		#ifndef HX_WINDOWS
+
+		nmem = SDL_ReadIO (stream ? (SDL_IOStream*)stream->handle : NULL, ptr, size);
```
tobil4sk: `size` here is the size of a unit to read, but here SDL_ReadIO is taking the length in bytes of the content to read. I think we should instead pass `size * count`. This is what caused asset manifest json files to be read in incorrectly.

Fancy2209: Have you been able to confirm this fixes the issue?

tobil4sk: Yes, this has fixed the json loading for me. However, after solving this I now get a crash during png loading which seems like it could have a similar cause.

Note: the fwrite method below also uses `size` incorrectly like this, though this also doesn't fix the png problem.

ACrazyTown:
I've gotten this to build on Windows!

Screenshot 2025-03-28 141614

Here's the commit from my fork: https://github.com/ACrazyTown/lime/commit/9993b5cb29216fb81cf0534d9cfad39a3d13b436. I would've done a pull request to your fork but GitHub won't let me for whatever reason

Fancy2209:
> I've gotten this to build on Windows!
> 
> Screenshot 2025-03-28 141614
> 
> Here's the commit from my fork: [ACrazyTown@9993b5c](https://github.com/ACrazyTown/lime/commit/9993b5cb29216fb81cf0534d9cfad39a3d13b436). I would've done a pull request to your fork but GitHub won't let me for whatever reason

I can just commit it and it should retain you as the author, thank you!
Will try fixing IO outside windows

project/src/backend/sdl3/SDLSystem.cpp
Fancy2209: > Yes, this has fixed the json loading for me. However, after solving this I now get a crash during png loading which seems like it could have a similar cause.
> 
> Note: the fwrite method below also uses `size` incorrectly like this, though this also doesn't fix the png problem.

Last commit fixes it (I think), I had missed the output is now meant to be divided by size too

project/src/backend/sdl3/SDLSystem.cpp
```diff
@@ -0,0 +1,1003 @@
...
+
+		}
+
+		return allow;
+
+	}
+
+
+	FILE* FILE_HANDLE::getFile () {
+		return (FILE*)handle;
```
tobil4sk, line 668: The SDL_IOStream is now an opaque struct, so this cast no longer safe and causes segmentation faults on Linux when the FILE* is passed to png functions. Accessing it via the property solves this problem:

```suggestion
		SDL_PropertiesID properties = SDL_GetIOProperties((SDL_IOStream*)handle);
		return (FILE*)SDL_GetPointerProperty(properties, SDL_PROP_IOSTREAM_STDIO_FILE_POINTER, NULL);
```
> The internal FILE * is available through a standard SDL_IOStream property, for streams made through SDL_IOFromFile() that use stdio behind the scenes; apps use this pointer at their own risk and should make sure that SDL and the app are using the same C runtime.

https://wiki.libsdl.org/SDL3/README/migration#sdl_rwopsh

`SDL_IOFromFile` sets this property, see also:
https://wiki.libsdl.org/SDL3/SDL_IOFromFile#remarks

We may need separate logic for android and possibly windows

tobil4sk: > We may need separate logic for android and possibly windows

Compare to the sdl2 implementation:
https://github.com/Fancy2209/lime/blob/d74b0cbcf972b3287c02ef5a357543275e13a7b7/project/src/backend/sdl/SDLSystem.cpp#L602-L641

Fancy2209:
That was a lot of work, wew
Finally all targets pass Ci and now this is ready for proper reviewing and better testing

EliteMasterEric (approved):
We have been reviewing this pull request internally and have found no major issues so far, only a couple missing graphics (which should be trivially resolved on our end). Performance seems similar and it would be great to have the new features introduced once this is merged.

Fancy2209:
> We have been reviewing this pull request internally and have found no major issues so far, only a couple missing graphics (which should be trivially resolved on our end). Performance seems similar and it would be great to have the new features introduced once this is merged.

Good to know!

About missing graphics, I've had a few FileIO issues, but I thought I had (finally) fully fixed them, could you elaborate one what you mean? I just wanna make sure I didn't break something

Kade-github:
Theres only one function in this PR that I needed to get, which was SDL_GetTicksNS(), which is similar to the function being exposed as "System::GetTimer();" I implemented this as "System::GetTimerNS();" but I am wondering of other SDL3 only features have yet to be exposed? Otherwise this PR looks good :D

EliteMasterEric:
> > We have been reviewing this pull request internally and have found no major issues so far, only a couple missing graphics (which should be trivially resolved on our end). Performance seems similar and it would be great to have the new features introduced once this is merged.
> 
> Good to know!
> 
> About missing graphics, I've had a few FileIO issues, but I thought I had (finally) fully fixed them, could you elaborate one what you mean? I just wanna make sure I didn't break something

After more testing I found that the missing graphic was related to how I was handling SDL timestamps, and resolving the issue fixed the graphic. There are no FileIO issues that I am aware of.

Fancy2209:
> Theres only one function in this PR that I needed to get, which was SDL_GetTicksNS(), which is similar to the function being exposed as "System::GetTimer();" I implemented this as "System::GetTimerNS();" but I am wondering of other SDL3 only features have yet to be exposed? Otherwise this PR looks good :D

Oh yeah definitely, I at the very least want to 
- expose SDL_GetTicksNS like y'all did on Funkin
- Allow Windows to optionally support for transparent backgrounds (I already figured this one out just waiting on this to get merged before PRing it)
- replace file dialogs with the SDL Dialog API when using SDL3
- Expose SDL_GetSystemTheme

I also know SDL3 adds a camera API who would allow us to implement Camera support, same thing with the System Tray API.
I also need to check what we need to change for Clipboard to handle the fact it can get more than just text now.

The only thing currently exposed is that enabling high DPI should also work on windows now, the above is only my wishlist/to work on list after this one gets merged.

If you wanna see everything SDL3 adds this is a pretty good list, though I noticed it doesn't mention some things like SDL_WINDOW_TRANSPARENT
https://wiki.libsdl.org/SDL3/NewFeatures

joshtynjala:
> I am wondering of other SDL3 only features have yet to be exposed

We've asked Fancy to wait on exposing any of SDL3's new features. To be clear, we definitely plan to take advantage of many of those new features in Lime and OpenFL because there are a lot of very nice things in SDL3. However, in terms of testing and review, it's much easier if we can focus on the upgrade with the same set of features that already existed in SDL2. After we merge this foundational SDL3 support, then we can start having fun with the new stuff.

Kade-github:
> > I am wondering of other SDL3 only features have yet to be exposed
> >
> To be clear, we definitely plan to take advantage of many of those new features in Lime and OpenFL because there are a lot of very nice things in SDL3. 

Awesome! Can't wait :D

EliteMasterEric:
It will also be great to start seeing changes in OpenFL as well, such as replacing this line with the proper nanosecond timestamp call:

https://github.com/openfl/openfl/blob/782627b89de76ff6ca85c267f8e36e2e09e1dc92/src/openfl/display/FPS.hx#L54

(i.e. finally resolving #1569 )

dimensionscape:
> It will also be great to start seeing changes in OpenFL as well, such as replacing this line with the proper nanosecond timestamp call:
> 
> https://github.com/openfl/openfl/blob/782627b89de76ff6ca85c267f8e36e2e09e1dc92/src/openfl/display/FPS.hx#L54
> 
> (i.e. finally resolving #1569 )

That line only affects the flash target... and does not actually solve anything. We also can't use anything from our native libraries in flash at all.

WIth that said, the main loop has been overhauled for exact preciseness with better support for vsync (and fallback for hardware that doesnt support vsync). Most of that work is about 95% complete and should make it's way into 8.3 over the coming months.

KarimAkra:
I tried this on Android and image files (PNG atleast) decoding seems to have broke? I did some logging and ended up with the assumption that the `lime::fread` method on `SDLSystem` isn't working properly (although it's literally the same thing in SDL3 migration docs... could be an internal issue with SDL..? Tho i found nothing on SDL issues)
I might be wrong but if that is the source of the issue then any C++ target other than Windows should have Image decoding issues and anything that uses `lime::fread` would be broken on these platforms

KarimAkra:
Also forgot to mention that trying to get displayMode is throwing a segfault for me on Android

Fancy2209:
> Tho i found nothing on SDL issues) I might be wrong but if that is the source of the issue then any C++ target other than Windows should have Image decoding issues and anything that uses `lime::fread` would be broken on these platforms

I know it isn't because this whole PR was made on Linux... Very weird Android is broken

gepatto:
Just chiming in.
I've tested on a Raspberry Pi.
And it works with some adjustments to the build.xml
I have been working with a default sdl3-config file that one gets when compiling sdl3 with default settings.
It's a bit different from custom linux sdl3 config in this lime branch so I'm using a separate file for the Raspberry Pi.

dimensionscape:
> Just chiming in. I've tested on a Raspberry Pi. And it works with some adjustments to the build.xml I have been working with a default sdl3-config file that one gets when compiling sdl3 with default settings. It's a bit different from custom linux sdl3 config in this lime branch so I'm using a separate file for the Raspberry Pi.

Yay this is really important. I'm glad you chimed in I was just wondering how well this worked on Pi.

Fancy2209:
I will close this PR and re open a version with a cleaner tree soon, will also take the time to try to wire mojoAL as it now supports SDL3

joshtynjala:
Today, I took some time to check out this branch, merge in the latest from 9.0.0-dev, and do a little local testing.

The Feathers UI components-explorer sample builds and runs with neko, hl, and cpp targets on macOS. However, rendering on my HiDPI monitor is not quite right when I use`<window allow-high-dpi="true"/>`. Everything renders at the correct scale factor, but it's like the stage bounds reach beyond the window on all four sides. I suspect that this is related to the way that coordinates have changed in SDL3 to better support HiDPI.

For iOS, it builds and launches. However, I'm seeing only a white screen when I launch components-explorer. I haven't investigated deeper than that.

For Android, it builds, launches, but renders everything at a _super_ tiny scale factor. Probably a similar scaling issue with SDL3's changes to coordinates for HiDPI.

I'm planning to test on Windows and Linux later, but these are my initial observations on my main machine.

Fancy2209:
> Today, I took some time to check out this branch, merge in the latest from 9.0.0-dev, and do a little local testing.
> 
> The Feathers UI components-explorer sample builds and runs with neko, hl, and cpp targets on macOS. However, rendering on my HiDPI monitor is not quite right when I use`<window allow-high-dpi="true"/>`. Everything renders at the correct scale factor, but it's like the stage bounds reach beyond the window on all four sides. I suspect that this is related to the way that coordinates have changed in SDL3 to better support HiDPI.
> 
> For iOS, it builds and launches. However, I'm seeing only a white screen when I launch components-explorer. I haven't investigated deeper than that.
> 
> For Android, it builds, launches, but renders everything at a _super_ tiny scale factor. Probably a similar scaling issue with SDL3's changes to coordinates for HiDPI.
> 
> I'm planning to test on Windows and Linux later, but these are my initial observations on my main machine.

That sounds like me using SDL_GetWindowSize() instead of SDL_GL_GetDrawableSize() somewhere

joshtynjala:
And I tested Windows and Linux this afternoon. I got sidetracked for a bit because I needed to upgrade from Visual Studio Community 2019 to 2022 to get SDL3 to compile on Windows. I was getting the following error:

> Cannot open include file: 'gameinput.h': No such file or directory

Upgrading to VS 2022 fixed it, for some reason. That's fine. I'm glad I'm not on an ancient version of VS anymore.

On Linux, rebuilding with hxcpp 4.3.2 didn't work, but I upgraded to latest hxcpp from GitHub and rebuild completed successfully. This is fine. By the time we release Lime 9, hxcpp should be updated on Haxelib anyway.

Anyway, on both Windows and Linux, I could compile and run the Feathers UI components-explorer sample. However, both rendered the window at half of the normal width and height. When I tried to set `allow-high-dpi="false"`, Windows and Linux still had a window at half width/height.

project/src/backend/sdl3/SDLSystem.cpp
```diff
@@ -0,0 +1,1041 @@
...
+			vdynamic* _bounds = (vdynamic*)hl_alloc_dynobj ();
+			hl_dyn_seti (_bounds, id_x, &hlt_i32, bounds.x);
+			hl_dyn_seti (_bounds, id_y, &hlt_i32, bounds.y);
+			hl_dyn_seti (_bounds, id_width, &hlt_i32, bounds.w);
+			hl_dyn_seti (_bounds, id_height, &hlt_i32, bounds.h);
+
+			hl_dyn_setp (display, id_bounds, &hlt_dynobj, _bounds);
+
+			Rectangle safeAreaInsets;
+			Display::GetSafeAreaInsets(id, &safeAreaInsets);
```
ACrazyTown: Was getting a crash on macOS when targeting Hashlink, this fixes it:

```suggestion
			Display::GetSafeAreaInsets(id - 1, &safeAreaInsets);
```

ACrazyTown (changes requested):

dimensionscape:
> I will close this PR and re open a version with a cleaner tree soon, will also take the time to try to wire mojoAL as it now supports SDL3

It's been about a year, so I've gone ahead and started the official migration process with a scoped branch largely based on the accomplishments here. 

https://github.com/openfl/lime/tree/migration/sdl3

Fancy2209:
> > I will close this PR and re open a version with a cleaner tree soon, will also take the time to try to wire mojoAL as it now supports SDL3
> 
> It's been about a year, so I've gone ahead and started the official migration process with a scoped branch largely based on the accomplishments here.
> 
> https://github.com/openfl/lime/tree/migration/sdl3

Thank you!

PR #1920, opened 2025-03-10
Remove HXCPP_CPP11 from build xml files

tobil4sk:
For updating some dependencies, we may want to use a different c++ standard, but it is not possible to do this because HXCPP_CPP11 takes priority

### DIFF
--- a/project/Build.xml
+++ b/project/Build.xml
@@ -1,7 +1,6 @@
 <xml>
 
 	<set name="PLATFORM" value="android-21" if="android" />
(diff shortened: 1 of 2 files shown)


tobil4sk:
It turns out it is not sufficient to remove `HXCPP_CPP11`, as hxcpp still sets it by default. We could use `HXCPP_NO_CPP11`, however, that breaks mac (see #1924). We've reverted the openal-soft update to avoid running into this, so it is best to leave this as is until we decide to update openal-soft again.

PR #1922, opened 2025-03-11
Fix Font.getGlyph(s) with emoji characters

tobil4sk:
The original code didn't work for emojis, but the stack overflow discussion since then includes a fix for this issue. See the comment:

https://stackoverflow.com/questions/2948308/how-do-i-read-utf-8-characters-via-a-pointer#comment72311507_2953960

Now readNextChar can extract the unicode value correctly, however, it still does not result in a correct glyph index being found.

### DIFF
--- a/project/src/text/Font.cpp
+++ b/project/src/text/Font.cpp
@@ -28,109 +28,135 @@
 
 unsigned long readNextChar (const char*& p)
 {
-	 // TODO: since UTF-8 is a variable-length
-	 // encoding, you should pass in the input
-	 // buffer's actual byte length so that you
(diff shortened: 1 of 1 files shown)


tobil4sk:
I was able to get it working properly by adding in `FT_Select_Charmap((FT_Face)face, FT_ENCODING_UNICODE);`. We already have some code to set the char map, so that probably needs to be updated or replaced: https://github.com/openfl/lime/blob/a9f72d65d906b6dec70f9237a9f6e85e6695bf83/project/src/text/Font.cpp#L402-L422

PR #1925, opened 2025-03-14
Include PLATFORM define for older NDK support

dimensionscape:
Hxcpp needs the PLATFORM define for older ndks that require the gcc toolchain. We initially removed it since Lime potentially was going to have a minimum requirement of android sdk 28 due to openal soft version 1.20.1, however, this was solved by upgrading to 1.21.1 which presented no further conflicts with sdk 21. 

To simply things, we just provide both defines for now.

### DIFF
--- a/tools/platforms/AndroidPlatform.hx
+++ b/tools/platforms/AndroidPlatform.hx
@@ -162,8 +162,9 @@ class AndroidPlatform extends PlatformTarget
 		for (architecture in architectures)
 		{
 			var minSDKVer = project.config.getInt("android.minimum-sdk-version", 21);
-			var haxeParams = [hxml, "-D", "android", "-D", 'PLATFORM_NUMBER=$minSDKVer'];
-			var cppParams = ["-Dandroid", '-DPLATFORM_NUMBER=$minSDKVer'];
+			//PLATFORM define needed for older ndk and gcc toolchain
+			var haxeParams = [hxml, "-D", "android", "-D", 'PLATFORM_NUMBER=$minSDKVer', "-D", 'PLATFORM=$minSDKVer'];
+			var cppParams = ["-Dandroid", '-DPLATFORM_NUMBER=$minSDKVer', '-DPLATFORM=$minSDKVer'];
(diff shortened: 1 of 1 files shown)


tobil4sk:
We mainly need `PLATFORM_NUMBER` for hxcpp 4.3.2, where it became configurable however min sdk 21 wasn't set by default yet for 32 bit architectures.

This may actually be useful to go in `develop`, as it may solve some issues building on android and prevent people having to resort to git hxcpp. Disabling 32 bit builds by default may also help even further even for older versions of hxcpp.

Also note, that at least for builds with openal-soft, you can't build lime with [an ndk older than 20](https://github.com/openfl/lime/commit/f5a24dc40aeb23981fff35997089787826178939), so `PLATFORM` would never be read in that case. I guess it's possible that builds with mojoal might still work with older ndks but I haven't tested.

tobil4sk:
> We mainly need PLATFORM_NUMBER for hxcpp 4.3.2

It turns out I had confused myself, 4.3.2 does not include the change so `PLATFORM_NUMBER` does nothing unless using git hxcpp, where the min sdk is already set to 21 currently. This means we can't really do anything to fix things with hxcpp 4.3.2.

tobil4sk:
I've tested with some older ndks (in particular, I was successful with r15 and r17). I had to make some modifications to hxcpp, but I was able to get them compiling when using `-Dmojoal`, so if we want to continue supporting that then this change makes sense.

tobil4sk (approved):

player-03:
> To simply things, we just provide both defines for now.

But what actually happened is we forgot about it, and no one seems to have complained. I don't know, still sounds like a bug worth fixing, but with Google becoming more and more strict, there are probably fewer and fewer people who care about old NDKs.

tobil4sk:
> But what actually happened is we forgot about it, and no one seems to have complained.

Note, this was on the 8.3.0 branch until now, which might mean fewer users interacted with it.

I think it is best to merge this unless we explicitly disallow ndks older than 19 from being used for lime projects.

PR #1928, opened 2025-03-14
[Android] Phase out older ARM architectures

MAJigsaw77:
- **Removed `armv7` by default** → Modern devices use `arm64`; keeping `armv7` adds overhead.  
- **Removed `armv5` completely** → Too old, rarely used, and complicates maintenance.

### DIFF
--- a/tools/platforms/AndroidPlatform.hx
+++ b/tools/platforms/AndroidPlatform.hx
@@ -135,20 +135,17 @@ class AndroidPlatform extends PlatformTarget
 		var hxml = targetDirectory + "/haxe/" + buildType + ".hxml";
 		var sourceSet = destination + "/app/src/main";
 
-		var hasARMV5 = (ArrayTools.containsValue(project.architectures, Architecture.ARMV5)
-			|| ArrayTools.containsValue(project.architectures, Architecture.ARMV6));
-		var hasARMV7 = ArrayTools.containsValue(project.architectures, Architecture.ARMV7);
 		var hasARM64 = ArrayTools.containsValue(project.architectures, Architecture.ARM64);
-		var hasX86 = ArrayTools.containsValue(project.architectures, Architecture.X86);
+		var hasARMV7 = ArrayTools.containsValue(project.architectures, Architecture.ARMV7);
 		var hasX64 = ArrayTools.containsValue(project.architectures, Architecture.X64);
+		var hasX86 = ArrayTools.containsValue(project.architectures, Architecture.X86);
 
 		var architectures = [];
 
-		if (hasARMV5) architectures.push(Architecture.ARMV5);
-		if (hasARMV7) architectures.push(Architecture.ARMV7);
 		if (hasARM64) architectures.push(Architecture.ARM64);
-		if (hasX86) architectures.push(Architecture.X86);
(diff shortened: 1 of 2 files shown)


tools/platforms/AndroidPlatform.hx
```diff
@@ -135,20 +135,17 @@ class AndroidPlatform extends PlatformTarget
...
 		var sourceSet = destination + "/app/src/main";
 
-		var hasARMV5 = (ArrayTools.containsValue(project.architectures, Architecture.ARMV5)
-			|| ArrayTools.containsValue(project.architectures, Architecture.ARMV6));
-		var hasARMV7 = ArrayTools.containsValue(project.architectures, Architecture.ARMV7);
 		var hasARM64 = ArrayTools.containsValue(project.architectures, Architecture.ARM64);
-		var hasX86 = ArrayTools.containsValue(project.architectures, Architecture.X86);
+		var hasARMV7 = ArrayTools.containsValue(project.architectures, Architecture.ARMV7);
 		var hasX64 = ArrayTools.containsValue(project.architectures, Architecture.X64);
+		var hasX86 = ArrayTools.containsValue(project.architectures, Architecture.X86);
```
tobil4sk, line 140: Could we keep the order of the lines the same as before here (and below in the file)? It complicates the diff

tools/platforms/AndroidPlatform.hx
```diff
@@ -364,23 +353,14 @@ class AndroidPlatform extends PlatformTarget
...
-		if (armv5) commands.push(["-Dandroid", platformDefine]);
-		if (armv7) commands.push(["-Dandroid", "-DHXCPP_ARMV7", platformDefine]);
-		if (arm64) commands.push(["-Dandroid", "-DHXCPP_ARM64", platformDefine]);
-		if (x86) commands.push(["-Dandroid", "-DHXCPP_X86", platformDefine]);
-		if (x64) commands.push(["-Dandroid", "-DHXCPP_X86_64", platformDefine]);
+
+		commands.push(["-Dandroid", "-DHXCPP_ARM64", platformDefine]);
+		commands.push(["-Dandroid", "-DHXCPP_ARMV7", platformDefine]);
+		commands.push(["-Dandroid", "-DHXCPP_X86_64", platformDefine]);
+		commands.push(["-Dandroid", "-DHXCPP_X86", platformDefine]);
```
tobil4sk: It would be useful to do the same for the rebuild as well:
- don't rebuild armv7 and x86_32 by default
- rebuild if the project contains them, like we used to do for x86_32

tobil4sk (changes requested):

tobil4sk (approved):
Thanks!

ghost:
why don't we remove armv5, armv6, mips and mipsel entirely? they're now not compatible with 2025 standarts

tobil4sk:
If we are intending to support older ndks for now (see #1925), then it may make sense to leave in armv5

MAJigsaw77:
> If we are intending to support older ndks for now (see #1925), then it may make sense to leave in armv5

I doubt anyone would use `armv5` anymore tho.

player-03:
If we're going to drop support for armv5 and older NDKs, it would probably happen in a major release, like 9.0.0.

Issue #1923, opened 2025-03-12
Unable to build lime with android ndk r28


tobil4sk:
Two of our dependencies fail to build with android ndk r28:

openal:
```
./lib/openal/al/event.cpp:189:9: warning: ignoring temporary created by a constructor declared with 'nodiscard' attribute [-Wunused-value]
  189 |         std::lock_guard<std::mutex>{context->mEventCbLock};
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
Error: ./lib/openal/common/almalloc.cpp:24:12: error: use of undeclared identifier 'aligned_alloc'
   24 |     return aligned_alloc(alignment, size);
      |            ^
1 error generated.
```

sdl:
```
Error: ./lib/sdl/src/sensor/android/SDL_androidsensor.c:174:9: error: 'ALooper_pollAll' is unavailable: obsoleted in Android 1 - ALooper_pollAll may ignore wakes. Use ALooper_pollOnce instead. See The API documentation for more information
  174 |     if (ALooper_pollAll(0, NULL, &events, (void**)&source) == LOOPER_ID_USER) {
      |         ^
/home/tobil/Dev/Android/SDK/ndk/r28/toolchains/llvm/prebuilt/linux-x86_64/bin/../sysroot/usr/include/android/looper.h:234:5: note: 'ALooper_pollAll' has been explicitly marked unavailable here
  234 | int ALooper_pollAll(int timeoutMillis, int* outFd, int* outEvents, void** outData)
      |     ^
1 error generated.
```

tobil4sk:
Note, these issues also happen with r27c, which is currently the oldest ndk version that's still in support: https://github.com/android/ndk/wiki#releases

joshtynjala:
We recently updated the 8.3.0-Dev branch to use SDL 2.30.12. I wonder if that version supports NDK r27 and r28? That was the newest SDL2 update at the time, but it looks like there may be a couple of newer SDL2 updates since then that we could try.

dimensionscape:
`aligned_alloc` is introduced in sdk 28, so make sure the minSdkVersion is set to 28. 

[...]

PR #1929, opened 2025-03-15
[ci] Run against stable android ndk

tobil4sk:
Now that the 8.3.0 branch supports stable ndks (see #1923), this will help maintain that support.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -340,6 +340,7 @@ jobs:
             ndll/Android/
             !**/.gitignore
           if-no-files-found: error
+        if: matrix.ndk-version != 'stable'
 
       - name: Install samples
         run: |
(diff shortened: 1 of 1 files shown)


ghost:
suggestion: use [ANDROID_NDK_LATEST_HOME](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#environment-variables-2) and update CI's

tobil4sk:
We should definitely test against the LTS version, because that should be a priority (and is likely easier) over supporting the very latest released ndk. Consider that up until now, lime 8.1 had to be built with ndk 15 and hxcpp 4.3.2 only supports up to 21. I guess we could run with both ndks (LTS and latest if available), it's just a question of whether it would help much or not.

ghost:
well I don't know, when it comes to lime I use latest available ndk possible (incase `r29-beta1`) and I use the `custom` folder to fix lib issues, it's up to OpenFL and to you to decide that

PR #1936, opened 2025-03-27
Update mojoal to fix audio bug

tobil4sk:
Currently, builds that use mojoal (explicitly via `-Dmojoal` or implicitly in static builds) have broken audio. This can be seen in the PiratePig openfl sample, and I've reproduced it on Android, Linux, and Windows. I've seen crashes occur, though most times the app starts fine but audio is heavily corrupted.

Updating mojoal seems to solve the problem.


tobil4sk:
@joshtynjala it might be good to put this in 8.3.0 if that's still possible, from my testing audio was unusable for any `-static` build.

joshtynjala:
@tobil4sk Depending on availability to do a release, we could be less than a week away from releasing 8.3. Not guaranteed, and it's likely to be later in November, but I didn't want to make a last minute change.

We might consider moving it into develop after the release, for 8.3.1.

PR #1938, opened 2025-03-28
Prefer currently booted iOS Simulator

gregdenness:
This PR updates XCodeHelper.getSimulatorID to first check if a simulator is already running and use that device's UDID, if there are no simulators running it will fall back to original behaviour.

This improves usability when switching simulators manually via the Simulator app, as users may wish to switch simulators multiple times during testing. For example if I open an "iPad 11 inch" simulator specifically and then build I'd expect it to be installed and run on that simulator.

This change is non-breaking and falls back to the existing logic if no simulator is currently running.

### DIFF
--- a/src/lime/tools/XCodeHelper.hx
+++ b/src/lime/tools/XCodeHelper.hx
@@ -43,15 +43,35 @@ class XCodeHelper
 
 	private static function extractSimulatorID(line:String):String
 	{
-		var id = line.substring(line.indexOf("(") + 1, line.indexOf(")"));
-
-		if (id.indexOf("inch") > -1 || id.indexOf("generation") > -1)
+		// Simulator ID's are always:
+		// XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
+		var searchPos = 0;
+	
+		while (true)
 		{
-			var startIndex = line.indexOf(")") + 2;
-			id = line.substring(line.indexOf("(", startIndex) + 1, line.indexOf(")", startIndex));
+			var openParen = line.indexOf("(", searchPos);
+			var closeParen = line.indexOf(")", openParen);
+			
+			if (openParen == -1 || closeParen == -1)
+			{
+				// Couldn't find a valid ID
+				break;
+			}
+	
+			var potentialID = line.substring(openParen + 1, closeParen);
+	
+			// Simulator IDs are always UUID strings exactly 36 chars long
+			if (potentialID.length == 36 && potentialID.indexOf("-") > -1)
+			{
+				return potentialID;
+			}
+			
+			// Skip past the current set of parentheses and keep looking
+			searchPos = closeParen + 1;
 		}
-
-		return id;
+	
+		// No valid ID found
+		return "";
 	}
 
 	private static function getBootedSimulator():SimulatorInfo
@@ -62,7 +82,7 @@ class XCodeHelper
 		{
 			if (line.indexOf("(Booted)") > -1)
 			{
-				return { id: extractSimulatorID(line), name: extractSimulatorFullName(line) };
+				return { id: extractSimulatorID(StringTools.trim(line)), name: extractSimulatorFullName(StringTools.trim(line)) };
 			}
 		}
 		return null;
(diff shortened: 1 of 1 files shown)


src/lime/tools/XCodeHelper.hx
```diff
@@ -54,6 +54,21 @@ class XCodeHelper
...
+	public static function getBootedSimulatorID():String 
+	{
+		// attempts to find a simulator which is currently booted and running
+		var result = System.runProcess("", "xcrun", ["simctl", "list", "devices", "booted"]);
+		var lines = result.split("\n");
+		for (line in lines) {
+			// Match a line like: "    iPhone 15 Pro (E5B048F6-BE70-498E-B5F6-B84B8594DDBF) (Booted)"
+			var match = ~/.*\(([A-F0-9\-]{36})\)\s+\(Booted\)/;
+			if (match.match(line)) {
+				return match.matched(1);
```
joshtynjala: For consistency with the rest of the code in this class, can you tweak this to use the existing `getSimulatorID()` method that we're using elsewhere? You can probably simplify the initial match as `line.indexOf("(Booted)")` in that case.

src/lime/tools/XCodeHelper.hx
```diff
@@ -173,6 +188,15 @@ class XCodeHelper
 
 	public static function getSimulatorID(project:HXProject):String
 	{
+		// try and get the simulator which is currently open and running
+		// if there are none running then fall back to getting the selected simulator
+		var booted = getBootedSimulatorID();
```
joshtynjala: Can you also modify `getSimulatorName()` to also return the booted simulator name? Both `getSimulatorID()` and `getSimulatorName()` are expected to return the same simulator.

joshtynjala (changes requested):

src/lime/tools/XCodeHelper.hx
gregdenness: Hi, yeah sure no problem. I'll do this next week. Thanks for the quick reply.

joshtynjala: Great, thanks! It's a good enhancement.

gregdenness:
I have updated the code to be consistent with the existing code as requested. I then encountered a strange issue where it wasn't working on certain simulators - those which had the cpu type in their names - such as "iPad air (M2)" - the (M2) part was breaking the extractSimulatorID function because of the parenthesis. I changed extractSimulatorID so that it looks for a 36 digit ID (all simulator IDs are). This is probably affecting the current lime release, though obviously only simulators which have the CPU type in their names.

joshtynjala:
Thanks! I'll test a bit and get it merged when I have a chance.

src/lime/tools/XCodeHelper.hx
```diff
@@ -43,15 +43,35 @@ class XCodeHelper
...
+			
+			if (openParen == -1 || closeParen == -1)
+			{
+				// Couldn't find a valid ID
+				break;
+			}
+	
+			var potentialID = line.substring(openParen + 1, closeParen);
+	
+			// Simulator IDs are always UUID strings exactly 36 chars long
```
player-03, line 63: If it really is that simple, we can save a lot of space here.

```haxe
private static function extractSimulatorID(line:String):String
{
	// Simulator IDs are always UUID strings exactly 36 chars long, in parentheses
	var id:EReg = ~/\(([A-F0-9\-]{36})\)/;
	return id.match(line) ? id.matched(1) : "";
}
```

src/lime/tools/XCodeHelper.hx
```diff
@@ -173,22 +207,32 @@ class XCodeHelper
...
+		}
+
+		simulator = getSelectedSimulator(project);
+		if (simulator != null)
 		{
-			return null;
+			return simulator;
 		}
-		return simulator.name;
+		return null;
```
player-03, line 235: `if (simulator != null) return simulator else return null` can be simplified to `return simulator`.

src/lime/tools/XCodeHelper.hx
gregdenness: Yeah that's much shorter for sure.

PR #1939, opened 2025-03-31
Allow easier integration with Swift libraries.

MAJigsaw77:
Libraries like AdMob use Objective-C but also integrate Swift for optimization purposes. Since Lime lacks Swift support, this can cause issues when integrating AdMob, Unity Ads, and similar frameworks.

This PR addresses the issue by adding the necessary Swift dependencies and libraries, ensuring proper compatibility with Swift-based SDKs.

### DIFF
--- a/templates/ios/template/{{app.file}}.xcodeproj/project.pbxproj
+++ b/templates/ios/template/{{app.file}}.xcodeproj/project.pbxproj
@@ -270,6 +270,15 @@
 		C01FCF4F08A954540054247B /* Debug */ = {/* Build configuration list for PBXProject "::APP_TITLE::" */
 			isa = XCBuildConfiguration;
 			buildSettings = {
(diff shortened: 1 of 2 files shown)


flashultra:
With this PR, is it now possible to write Lime extensions directly in Swift, or do we still need an Objective-C bridge for Haxe/Lime to call Swift code?

MAJigsaw77:
> With this PR, is it now possible to write Lime extensions directly in Swift, or do we still need an Objective-C bridge for Haxe/Lime to call Swift code?

This pr is mainly ment so libs that use swift behind the scenes but have an objc interface can be used, for example `Admob` `UnityAds` etc.

flashultra:
I see, thanks for clarifying. It’s a pity, as most modern SDKs and examples are Swift-based, and writing Objective-C bridges can be hard.

PR #1942, opened 2025-04-09
Replace Map with IMap on AssetCache

EliteMasterEric:
This allows `audio`, `font`, and `image` to be replaced with any object which implements `IMap`, which includes custom map implementations with different underlying behavior. This is required to replace the map with a different class, as I don't believe you can extend the standard `Map` class since it is an abstract and the underlying implementation is provided by the standard library.

In my use case, I am replacing Map with a CallbackMap class I made, which re-implements the standard Map but allows me to register callbacks when values are added or removed.

### DIFF
--- a/src/lime/utils/AssetCache.hx
+++ b/src/lime/utils/AssetCache.hx
@@ -1,6 +1,7 @@
 package lime.utils;
 
 import haxe.macro.Compiler;
+import haxe.Constraints.IMap;
 import lime.media.AudioBuffer;
 import lime.graphics.Image;
 #if !(macro || commonjs)
@@ -13,10 +14,10 @@ import lime._internal.macros.AssetsMacro;
 #end
 class AssetCache
 {
-	public var audio:Map<String, AudioBuffer>;
+	public var audio:IMap<String, AudioBuffer>;
 	public var enabled:Bool = true;
-	public var image:Map<String, Image>;
-	public var font:Map<String, Dynamic /*Font*/>;
+	public var image:IMap<String, Image>;
+	public var font:IMap<String, Dynamic /*Font*/>;
 	public var version:Int;
 
 	public function new()


dimensionscape:
Have you considered using a static extension or abstracting your own behavior over map or are you trying to hook into internal mutations?

dimensionscape:
Hey @EliteMasterEric, thanks again for the PR and the interesting approach.

I just want to clarify something, since I think this comes down to intent.

If the goal here is to hook into internal set() or get() calls made by Lime/OpenFL itself (e.g., tracking when assets are added or retrieved from AssetCache internally), then switching to IMap isn’t a reliable or maintainable way to do that. It introduces risks around platform behavior, weakens the API contract, and relies on assumptions about how Lime uses the map internally (which could change or be replaced at any time).

If instead you're trying to hook into map access or mutations from your own code, then that can be done cleanly today using:

An abstract over Map<K, V> that wraps the returned map from Lime and adds your custom get()/set() logic

Static extension methods for more readable external hooks (e.g., MapHooks.setWithLog(map, key, value))

These options let you customize behavior without needing any changes in Lime itself, and without sacrificing type safety or platform consistency.

So unless Lime/OpenFL explicitly adds support for cache-level hooks (like onAssetSet()), I'd rather avoid changing the API to accept IMap, since it opens the door to subtle bugs without a clear gain for the majority of users.

Let me know if I misunderstood your intent. Happy to discuss further!

player-03:
I agree with Chris. Generally speaking, the only reason you should examine a cache that deeply is to debug a specific issue. Otherwise, better to leave it as a black box. Perhaps we should make those variables `(default, null)`...

But as things stand, you can freely set `Assets.cache`. Make a subclass that overrides `AssetCache.set()`, and you can call as many listeners as you need, without any changes to Lime. Plus this way, `AssetCache.clear()` won't accidentally delete your listeners.

PR #1944, opened 2025-04-14
IOS embed a dynamic framework xcode project changes

lcsatchell:
This pull request updates the generated Xcode project when using embed=true with an iOS framework dependency. 
Specifically, it:
Adds a "Copy Files" build phase
Updates the Runpath Search Paths

These changes ensure that dynamic framework dependencies are correctly embedded and included in the final app bundle.
Note: You’ll need to rebuild Lime tools after applying these changes.

Why this is needed:
Dynamic frameworks aren’t automatically embedded in the same way static frameworks are. This becomes particularly important when working with Swift code, which must be included in a dynamic framework.

Use case:
To include Swift code in an extension, it was added to a custom dynamic framework. That framework was then included in the extension using embed=true. Without these changes, the framework wouldn't be properly compiled into the app.

### DIFF
--- a/templates/ios/template/{{app.file}}.xcodeproj/project.pbxproj
+++ b/templates/ios/template/{{app.file}}.xcodeproj/project.pbxproj
@@ -41,6 +41,10 @@
 		};
 /* End PBXContainerItemProxy section */
 
+/* Begin PBXCopyFilesBuildPhase section */
+::ADDL_PBX_COPY_FILES_BUILD_PHASE::
+/* End PBXCopyFilesBuildPhase section */
+
 /* Begin PBXFileReference section */
 		1D6058910D05DD3D006BFB54 /* ::APP_TITLE::.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = "::APP_TITLE::.app"; sourceTree = BUILT_PRODUCTS_DIR; };
 		1E2E17A5131E8B5D0048F3C7 /* ::APP_FILE::/assets */ = {isa = PBXFileReference; lastKnownFileType = folder; path = "::APP_FILE::/assets"; sourceTree = "<group>"; };
@@ -179,7 +183,7 @@
 			buildPhases = (
 				1D60588D0D05DD3D006BFB54 /* Resources */,
 				1D60588E0D05DD3D006BFB54 /* Sources */,
-				1D60588F0D05DD3D006BFB54 /* Frameworks */,
+				1D60588F0D05DD3D006BFB54 /* Frameworks */,::ADDL_PBX_BUILD_PHASE_LIST::
 			);
 			buildRules = (
 			);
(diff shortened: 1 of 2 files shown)


templates/ios/template/{{app.file}}.xcodeproj/project.pbxproj
```diff
@@ -41,6 +41,10 @@
 		};
 /* End PBXContainerItemProxy section */
 
+/* Begin PBXCopyFilesBuildPhase section */
+::ADDL_PBX_COPY_FILES_BUILD_PHASE::
+/* End PBXCopyFilesBuildPhase section */
```
IriySoft, line 46: I would add a null check here, like:

```
::if (ADDL_PBX_COPY_FILES_BUILD_PHASE != null)::
::ADDL_PBX_COPY_FILES_BUILD_PHASE::
::end::
```

because if you try to use this pbxproj template without having the build tools updated and rebuilt, you get the following:

```
/ Begin PBXCopyFilesBuildPhase section /
null
/ End PBXCopyFilesBuildPhase section /
```

That results in a corrupted pbxproj file

templates/ios/template/{{app.file}}.xcodeproj/project.pbxproj
```diff
@@ -179,7 +183,7 @@
 			buildPhases = (
 				1D60588D0D05DD3D006BFB54 /* Resources */,
 				1D60588E0D05DD3D006BFB54 /* Sources */,
-				1D60588F0D05DD3D006BFB54 /* Frameworks */,
+				1D60588F0D05DD3D006BFB54 /* Frameworks */,::ADDL_PBX_BUILD_PHASE_LIST::
```
IriySoft: Should be like this to be non-destructive:
`				1D60588F0D05DD3D006BFB54 /* Frameworks */,::if (ADDL_PBX_BUILD_PHASE_LIST != null)::::ADDL_PBX_BUILD_PHASE_LIST::::end`

templates/ios/template/{{app.file}}.xcodeproj/project.pbxproj
```diff
@@ -370,6 +374,7 @@
 				GCC_PREFIX_HEADER = "::APP_FILE::/::APP_FILE::-Prefix.pch";
 				INFOPLIST_FILE = "::APP_FILE::/::APP_FILE::-Info.plist";
 				IPHONEOS_DEPLOYMENT_TARGET = ::DEPLOYMENT::;
+				::ADDL_PBX_RUNPATH_SEARCH_PATHS::
```
IriySoft, line 377: `				::if(ADDL_PBX_RUNPATH_SEARCH_PATHS != null)::::ADDL_PBX_RUNPATH_SEARCH_PATHS::::end::`

templates/ios/template/{{app.file}}.xcodeproj/project.pbxproj
```diff
@@ -428,6 +433,7 @@
...
```
IriySoft: `				::if (ADDL_PBX_RUNPATH_SEARCH_PATHS != null)::::ADDL_PBX_RUNPATH_SEARCH_PATHS::::end::`

IriySoft (commented):
I suggest to add checks in case the project template will be used in the environment that adopts frameworks embedding patch and then someone will try to use the same template in the environment that doesn't.

player-03:
Are people really going to use this template without also updating the build tools?

IriySoft:
> Are people really going to use this template without also updating the build tools?

Well…. The short answer is no :)

I had to copy the project template to the extension that I’m making and got these build errors when switching Lime versions. And I thought that the extensions should at least build without errors if someone tries to use it with the current Lome, because I don’t know if the pull request ends up  accepted. But I added these null checks to my template copy so you’re right, these null checks are redundant in the pull request.

PR #1955, opened 2025-06-04
Fix SDLWindow behavior

barisyild:
This pull request fixes SDLWindow behavior on latest SDL.


Fancy2209:
Wouldn't it make sense to use SDLGetWindowSizeInPixels instead of GL Drawable Size?

barisyild:
Yeah you're right my mistake

PR #1961, opened 2025-07-17
Don't use `ReplaceVistaIcon` on windows `cpp` target anymore.

MAJigsaw77:
This pr makes it so it does no longer require the use of `ReplaceVistaIcon.ese` in order to apply an icon to to the `.exe` file by using the `.rc` file aswell as making it able to no longer ship with a `icon.ico` useless file anymore.


dimensionscape:
Makes sense to me.

joshtynjala:
Yeah, makes sense to me too. Does it still work with Haxe 3.4.7?

MAJigsaw77:
> Yeah, makes sense to me too. Does it still work with Haxe 3.4.7?

Idk lol

joshtynjala:
Well, someone will need to test that before merging. Good to test older Haxe 4 versions like 4.0 and 4.1 too.

If we find that it breaks Haxe 3.4.7 (which is still supported in Lime 8), it will either need to go into 9.0.0-dev instead, or we'll need to keep using ReplaceVistaIcon.exe for any version that breaks.

I seem to recall that there was some new rc file in hxcpp, so the Haxe compiler version may not matter, as long as folks can use a new enough hxcpp. Hopefully, that includes the current newest release of hxcpp on Haxelib, or we'll need to wait for an official hxcpp update there before merging.

MAJigsaw77:
> Well, someone will need to test that before merging. Good to test older Haxe 4 versions like 4.0 and 4.1 too.
> 
> If we find that it breaks Haxe 3.4.7 (which is still supported in Lime 8), it will either need to go into 9.0.0-dev instead, or we'll need to keep using ReplaceVistaIcon.exe for any version that breaks.
> 
> I seem to recall that there was some new rc file in hxcpp, so the Haxe compiler version may not matter, as long as folks can use a new enough hxcpp. Hopefully, that includes the current newest release of hxcpp on Haxelib, or we'll need to wait for an official hxcpp update there before merging.

From what I've seen, as long as hxcpp is `4.3.2` atleast it should work just fine, this is the [commit](https://github.com/HaxeFoundation/hxcpp/commit/b045f2522beea47971c9ac92db694dfaf08f3276) where the resource file has been added.

ADA-Funni:
🚀

Issue #1797, opened 2024-06-10
[ENHANCEMENT] Audio recording support


ACrazyTown:
Having support for audio recording would be really useful for something like voice chat in games.

For native targets, I think both SDL and OpenAL have support for audio recording, its just a matter of writing bindings and adding them thru CFFI. I've actually tried to  add bindings for OpenAL audio capturing but I got stuck trying to figure out how to send the data from`alcCaptureSamples()` from C++ to Haxe.

I haven't looked into HTML5 stuff but I'm pretty sure its a bit more straightforward compared to native

moha0100110:
i support this too, i want too :)

dimensionscape:
Closing as completed considering audio capture was implemented.

PR #1965, opened 2025-07-26
Implement OpenAL audio capture extension

ACrazyTown:
A step towards #1797. Implements the OpenAL audio capture functionality which I believe is supported on all 3 OpenAL backends Lime uses! I've tested this on Windows, Neko and HL


Raltyro:
Microphone!!

PR #1968, opened 2025-08-01
replace deprecated `apple-mobile-web-app-capable` with `mobile-web-app-capable`

ninjamuffin99:
see https://github.com/openfl/openfl/pull/2791

note: im unsure in which context the `winjs` target/template is used, if thats an older target it could potentially *not* be deprecated!


joshtynjala:
Already fixed by commit 3802d51e5c6c7865a96329835c3b437b611bb809, but thanks!

PR #1970, opened 2025-08-13
Fix System::GetDirectory unicode paths on windows

### LINKED ISSUE #1906: SDLSystem GetDirectory does not handle UTF16 correctly on Windows

tobil4sk:
The conversion was not valid, because the string returned from SHGetFolderPathA is ANSI, not ASCII.

By using the W variant of the function, we can avoid the need for a conversion altogether. According to the microsoft docs, this is what should be used anyway:

> New applications should always call the Unicode versions. Many world
> languages require Unicode. If you use ANSI strings, it will be
> impossible to localize your application.

https://learn.microsoft.com/en-us/windows/win32/learnwin32/working-with-strings

Fixes #1906

### DIFF
--- a/project/src/backend/sdl/SDLSystem.cpp
+++ b/project/src/backend/sdl/SDLSystem.cpp
@@ -142,11 +142,9 @@ namespace lime {
 
 				#elif defined (HX_WINDOWS)
 
-				char folderPath[MAX_PATH] = "";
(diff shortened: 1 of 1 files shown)


tobil4sk:
Whoops, didn't see there is already a PR for this #1907

PR #1971, opened 2025-08-14
Perform null check for XMLHttpRequest.upload

rainyt:
Some browsers may not support uploading, and accessing upload will return null.

### DIFF
--- a/src/lime/_internal/backend/html5/HTML5HTTPRequest.hx
+++ b/src/lime/_internal/backend/html5/HTML5HTTPRequest.hx
@@ -64,7 +64,8 @@ class HTML5HTTPRequest
 
 		if (parent.method == POST)
 		{
-			request.upload.addEventListener("progress", progress, false);
+			if(request.upload != null)
+				request.upload.addEventListener("progress", progress, false);
 		}
 		else
 		{


joshtynjala:
According to [MDN: XMLHttpRequest: upload property](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/upload):

> This feature is well established and works across many devices and browser versions. It’s been available across browsers since July 2015.

On [CanIUse: XMLHttpRequest API: upload](https://caniuse.com/mdn-api_xmlhttprequest_upload), I see that IE 9 and Android 4.3 are the most noteworthy browsers that didn't support it yet. Extremely old.

I'm not sure if it makes sense for us to add support for browser versions that are over 10 years old at this point. Back in the day, you didn't have any choice with Internet Explorer, but today, all major browsers are considered "evergreen" and get frequent auto-updates.

I also don't see any notes on MDN mentioning that the `upload` property may be null/undefined under any special circumstances (such as when the URL is http instead of https, or for other situations when security might be locked down).

rainyt:
> On [CanIUse: XMLHttpRequest API: upload](https://caniuse.com/mdn-api_xmlhttprequest_upload), I see that IE 9 and Android 4.3 are the most noteworthy browsers that didn't support it yet. Extremely old.

Yes, only very old browsers experience this issue. At present, I am adding it to be compatible with the quick game engine mode (like facebook mini game.). In some implementations of fast game engines, their XMLHttpRequest does not have an upload property.

PR #1974, opened 2025-08-16
Fix unicode system path conversions on linux

### LINKED ISSUE #1937: Invalid UTF-16 on Linux if program path contains non-ascii characters
For some reason, and it might correlate with #1906 but Linux seems to have issues with starting games made with Lime if the system language isn’t English, I’m currently away from home and I don’t have Linux installed on my new computer yet, but I’ve seen two examples of this issue happening (which I’ll link here)

https://github.com/CodenameCrew/CodenameEngine/issues/573 (Tested on Arch Linux)
https://github.com/JordanSantiagoYT/FNF-JS-Engine/issues/975 (Tested on an Debian/Ubuntu based Distribution)

Seems to happen on any Linux distribution as far as I know, this also has similar issues when running from a folder with special characters, but this was already fixed I believe, I haven’t tested on Windows/Mac yet, and I don’t have an Mac

tobil4sk:
For non windows platforms, we previously just passed `char*` into `wstring`, which didn't perform the necessary conversion into utf32 which is expected when hxcpp converts a wchar string to a hxstring on platforms where wchar has 4 bytes.

Ideally, we would skip the wstring conversion altogether, since it is an unnecessary conversion which has caused lots of issues like this in the past, however, for now this fixes #1937. I also tested on Windows and it seems to be working correctly still (`DESKTOP`, `APPLICATION` and `APPLICATION_STORAGE` at least). It would be useful if others could test on other systems like Mac. This is what I used to test on linux for example:
```haxe
// Source/Main.hx
class Main extends lime.app.Application {
	public override function onPreloadComplete() {
		trace(lime.system.System.applicationDirectory);
		trace(lime.system.System.applicationStorageDirectory);

		Sys.putEnv("HOME", "/TÉST");
		trace(lime.system.System.documentsDirectory);
		trace(lime.system.System.userDirectory);
		trace(lime.system.System.desktopDirectory);
	}
}
```

```xml
<!-- project.xml -->
<project>

	<meta title="HelloWorld" package="com.sample.helloworld" version="1.0.0" company="Cømpany Name" />
	<app main="Main" path="Éxport" file="HelloWorld" />

	<source path="Source" />

	<haxelib name="lime" />

	<assets path="Assets" rename="assets" />

</project>
```

### DIFF
--- a/project/src/backend/sdl/SDLSystem.cpp
+++ b/project/src/backend/sdl/SDLSystem.cpp
@@ -111,9 +111,9 @@ namespace lime {
 
 				if (path != nullptr) {
 
-						wstring_convert converter;
-						result = new std::wstring (converter.from_bytes(path));
-						SDL_free (path);
+					wstring_convert converter;
+					result = new std::wstring (converter.from_bytes(path));
+					SDL_free (path);
 
 				}
 
@@ -127,11 +127,11 @@ namespace lime {
 
 				if (path != nullptr) {
 
-        		wstring_convert converter;
-						result = new std::wstring (converter.from_bytes(path));
-						SDL_free (path);
+        			wstring_convert converter;
+					result = new std::wstring (converter.from_bytes(path));
+					SDL_free (path);
 
-        }
+				}
 
 				break;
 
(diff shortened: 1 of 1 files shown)


moxie-coder (approved):

ACoolioDude (approved):

HomuHomu833 (approved):

joshtynjala:
Tested on macOS and working well. Manually merged on command line due to conflict with previously merged PR.

tobil4sk:
> Manually merged on command line due to conflict with previously merged PR.

Thanks for testing and merging in! I just also made a fix to correct a minor merge error: d1369f1aaa4c6bc4131c6ca30c30fef47a22c71c

PR #1979, opened 2025-08-27
Add mappings for PS5 and Xbox series controllers on MacOS

slameron:
Not sure if this is necessary to do on windows or linux as well, my controllers seem to work fine on windows.


joshtynjala:
Fixed in 88769f618053a05b33206149a96601fa7a7944b7, along with many more new mappings.

PR #1980, opened 2025-09-01
Fix duplicate code in `PlatformTarget` subclasses, fix assets being included twice.

player-03:
Having eight separate implementations made it harder to maintain.

When an asset embedding bug got fixed on desktop targets, it got overlooked on others. This lead to assets being included twice on those other targets: once embedded, and once normally. (#1898) Now, that behavior is controlled from one place, and the bug is fixed for eight targets at once.

This also standardizes the case of `asset.embed == true && asset.type == TEMPLATE`. Previously, some targets would prioritze `embed` over `type`, embedding the template into the app without even processing it as a template. Now, it will always prioritize `type`, processing templates as templates and never trying to embed them.

I left out `AIRPlatform`, `FlashPlatform`, and `HTML5Platform` since they have more complex asset embedding behavior. I haven't checked if the bug is present on any of those.

### DIFF
--- a/tools/platforms/TizenPlatform.hx
+++ b/tools/platforms/TizenPlatform.hx
@@ -237,14 +237,6 @@ class TizenPlatform extends PlatformTarget
 		ProjectHelper.recursiveSmartCopyTemplate(project, "haxe", targetDirectory + "/haxe", context);
 		ProjectHelper.recursiveSmartCopyTemplate(project, "tizen/hxml", targetDirectory + "/haxe", context);
 
-		for (asset in project.assets)
-		{
-			if (asset.targetPath == "/appinfo.json" || asset.targetPath == "appinfo.json")
-			{
-				asset.type = AssetType.TEMPLATE;
-			}
-		}
-
 		// going to root directory now, but should it be a forced "assets" folder later?
 		copyProjectAssets(destination + "res/", "");
 		// copyProjectAssets(destination + "res/", "assets");
--- a/tools/platforms/TizenPlatform.hx
+++ b/tools/platforms/TizenPlatform.hx
@@ -261,4 +261,6 @@ class TizenPlatform extends PlatformTarget
 	@ignore public override function install():Void {}
 
 	@ignore public override function uninstall():Void {}
+
+	@ignore public override function watch():Void {}
 }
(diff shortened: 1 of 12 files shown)


player-03:
@MAJigsaw77 Check this out and let me know if it works.

player-03:
Cleaned up some more duplicate code while I was at it. I think that's it, though; the rest of the functions look unique.

tools/platforms/TizenPlatform.hx
```diff
@@ -258,31 +239,20 @@ class TizenPlatform extends PlatformTarget
...
-				{
-					// going to root directory now, but should it be a forced "assets" folder later?
-
-					AssetHelper.copyAssetIfNewer(asset, path);
-				}
-			}
-			else
-			{
-				AssetHelper.copyAsset(asset, path, context);
+				asset.type = AssetType.TEMPLATE;
```
player-03: _appinfo.json_ has been a special case since _TizenPlatform.hx_ was first created, and no one seems to have noticed it not working. Clearly the intent was to process it as a template, which is what I've done here. But... no one seems to have noticed the problem in seven years, so maybe we don't need the special case?

https://github.com/openfl/lime/blob/ecd7876c4b38853022e368415288ec2b93509ed1/tools/platforms/TizenPlatform.hx#L197-L209

tools/platforms/WebAssemblyPlatform.hx
```diff
@@ -481,42 +462,20 @@ class WebAssemblyPlatform extends PlatformTarget
...
+		copyProjectAssets(destination, targetDirectory + "/obj/assets");
 	}
 
 	@ignore public override function install():Void {}
 
 	@ignore public override function trace():Void {}
 
 	@ignore public override function uninstall():Void {}
+
+	@ignore public override function watch():Void {}
```
player-03, line 481: `watch()` relies on `getDisplayHXML()`, which is missing from `AIRPlatform` and `TizenPlatform`, so that explains why we don't have it there.

However, `WebAssemblyPlatform` has `getDisplayHXML()` and doesn't have `watch()`. I don't know why, but I kept that behavior anyway. It seems like it ought to work, if we just enable it.

MAJigsaw77:
{0983020E-7020-487A-8788-454710077708}

Screenshot_20250901_093432

Can confirm that it works on `Android`.

PR #1981, opened 2025-09-01
Implement `Future` aggregation functions.

player-03:
These are based on [JavaScript's promise concurrency functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#promise_concurrency), using Lime's nomenclature.

Supersedes #1844

### DIFF
--- a/src/lime/app/Future.hx
+++ b/src/lime/app/Future.hx
@@ -306,6 +306,135 @@ import lime.utils.Log;
 		future.value = value;
 		return future;
 	}
+
+	// Aggregation functions
+
+	/**
+	 * Creates a `Future` which completes once all the input `Future`s complete
+	 * successfully. Its completion value will be an array of the input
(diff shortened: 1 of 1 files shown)


player-03:
@EliteMasterEric How do these look to you?

player-03:
I'm going back and forth between calling it "`allResolved`" or "`allSettled`".

- Lime's `Promise` class consistently uses the term "resolve".
- We're copying JavaScript's functions, so it'd be more intuitive to use JavaScript's names.
- The function works slightly differently than the JS version, so maybe a slightly different name will make the change more obvious.
- [Wikipedia uses the term "resolve"](https://en.wikipedia.org/wiki/Futures_and_promises).

EliteMasterEric:
Just saw the notification for this.

I'm happy with this implementation. I think `resolve` makes the most sense considering the rest of the codebase uses that terminology.

Issue #1983, opened 2025-09-23
Lime doesn't respond to pixelratio change in runtime


IriySoft:
If an html5 app made with Lime/OpenFL is zoomed up in Chrome (didn't check the other browsers), like 500% or such, it becomes blurry. But if the page is reloaded without zooming back to 100%, the app becomes looking good. Can be tested in PiratePig app from OpenFL samples. 

I found that this happens because [HTML5Window.hx](https://github.com/openfl/lime/blob/develop/src/lime/_internal/backend/html5/HTML5Window.hx) only reads Browser.window.devicePixelRatio once in the constructor and caches it in `scale` variable (line 108). Then `scale` is used to set parent scale and to calculate canvas width and height.

https://github.com/user-attachments/assets/cd7a7007-2db6-4e91-998f-e4ac26f977d4

I could fix it the issue reading Browser.window.devicePixelRatio in `updateSize()` (line 1282) and changing scale, parent scale and canvas dimensions like this:

```
	var newScale = Browser.window.devicePixelRatio;
	if (scale != newScale) {
		scale = newScale;
		parent.__scale = scale;
		if (canvas != null)
		{
			canvas.width = Math.floor(parent.__width * scale);
			canvas.height = Math.floor(parent.__height * scale);
			canvas.style.width = parent.__width + "px";
			canvas.style.height = parent.__height + "px";
		}
	}

```
I had to modify openfl.display.Stage.hx too to make the renderer adapt to the possible pixel ratio change in __resize() like this:

```
		if (__renderer != null)
		{
			__renderer.__pixelRatio = #if openfl_disable_hdpi 1 #else window.scale #end;
		}

```

PR #1984, opened 2025-09-24
[HTML5] Respond to pixel ratio changes

IriySoft:
Fix for the issue:
https://github.com/openfl/lime/issues/1983

Short: Lime doesn't respond to the pixel ratio change that may happen in runtime, for example, when the user zooms the browser window with a running HTML5 app in Chrome with Ctrl +/-.


joshtynjala:
If we make this change, we should possibly also add some kind of event to Window to indicate to developers that the scale has changed. They may need to switch textures or do other re-init in response. With that in mind, this will need to go into a feature release. Probably 8.4, so that it has more time for testing too.

When we switch to SDL3 for native targets in Lime 9, we can dispatch the same event when we respond to pixel ratio changes indicated by its C++ event.

IriySoft:
> If we make this change, we should possibly also add some kind of event to Window to indicate to developers that the scale has changed. They may need to switch textures or do other re-init in response. With that in mind, this will need to go into a feature release. Probably 8.4, so that it has more time for testing too.

I believe this could be fired by OnResize, then the developer can check pixel ratio manually.

I don't think that the zoom issue happens often - most probably no one even noticed :) There is one case when it can be noticed: according to the report from the QA team of Crazy Games portal, the games become blurry when switched to full screen on Chromebook - most probably when it happens, pixel ratio is changed.

IriySoft:
Need to check the fresh changes in main project :))

PR #1988, opened 2025-10-13
Implement ALC.getStringList()

### LINKED ISSUE #1986: OpenAL strings not handled properly
Trying to get a list of all available audio devices by doing `ALC.getString(null, ALC.ALL_DEVICES_SPECIFIER);` will only return the first one. OpenAL returns a char array with multiple null terminated strings, but it seems that this is not accounted for in CFFI.

This occurs on Hashlink, Windows and Neko. Probably on all native targets, but that's all I can test.

ACrazyTown:
Fixes #1986

There are certain OpenAL parameters, meant to be passed into `ALC.getString()`, that return a list of strings. For example, `ALC_ALL_DEVICES_SPECIFIER` to get the names of available playback audio devices or `ALC_CAPTURE_DEVICE_SPECIFIER`to get the names of available capture devices.

`ALC.getString()` assumes that the result will always be a standard C string with a single NULL character at the end, but this is not the case when passing the params mentioned above:

> An alcGetString query of ALC_DEVICE_SPECIFIER or
ALC_CAPTURE_DEVICE_SPECIFIER with a NULL device passed in will return a list
of available devices. Each device name will be separated by a single NULL character and
the list will be terminated with two NULL characters.
(https://www.openal.org/documentation/openal-1.1-specification.pdf)

This PR introduces the `ALC.getStringList(device, param)` method which handles these lists properly and returns them as an `Array<String>`. If list handling is unnecessary for the param provided, the result will simply be `[ALC.getString(device, param)]`

### DIFF
--- a/project/src/media/openal/OpenALBindings.cpp
+++ b/project/src/media/openal/OpenALBindings.cpp
@@ -3456,6 +3456,7 @@ namespace lime {
 			listData++;
 
 			result += length;
+			free(_result);
 
 		}
 
--- a/project/src/media/openal/OpenALBindings.cpp
+++ b/project/src/media/openal/OpenALBindings.cpp
@@ -3398,6 +3398,72 @@ namespace lime {
 	}
 
 
+	value lime_alc_get_string_list (value device, int param) {
+
+		ALCdevice* alcDevice = (ALCdevice*)val_data (device);
+		const char* result = alcGetString (alcDevice, param);
+
+		if (!result || *result == '\0') {
+
+			return alloc_null ();
+
+		}
+
+		value list = alloc_array (0);
+
+		while (*result != '\0') {
+
+			val_array_push (list, alloc_string (result));
+			result += strlen (result) + 1;
+
+		}
+
+		return list;
+
+	}
+
+
+	HL_PRIM varray* HL_NAME(hl_alc_get_string_list) (HL_CFFIPointer* device, int param) {
+
+		ALCdevice* alcDevice = device ? (ALCdevice*)device->ptr : 0;
+		const char* result = alcGetString(alcDevice, param);
+
+		if (!result || *result == '\0') {
+
+			return 0;
+
+		}
+
+		int count = 0;
+		const char* temp = result;
+		while (*temp != '\0') {
+
+			count++;
+			temp += strlen (temp) + 1;
+
+		}
+
+		varray* list = hl_alloc_array (&hlt_bytes, count);
+		vbyte** listData = hl_aptr (list, vbyte*);
+
+		while (*result != '\0') {
+
+			int length = strlen (result) + 1;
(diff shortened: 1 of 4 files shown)


KarimAkra:
This looks good but a little hardcoded..?

I personally went with this:
image

And we just split the string at the comma on haxe
image

ACrazyTown:
> This looks good but a little hardcoded..?

Yeah, it's a bit ugly but I had to do that since `lime_alc_get_string_list` assumes it's dealing with a list and not a standard C-string. I figured it'd be fine since it's a private API meant to be called by Lime internally.

Maybe the check could be done on the C++ side but ngl my C(++) is a bit rusty.

> And we just split the string at the comma on haxe

This could've be done as well but I opted in for a seperate method because I wanted to do as much as possible on the C++ side and other OpenAL bindings seem to do [something](https://javadoc.lwjgl.org/org/lwjgl/openal/ALUtil.html#getStringList(long,int)) [similar](https://opentk.net/api/OpenTK.Audio.OpenAL.AlcGetStringList.html).

project/src/media/openal/OpenALBindings.cpp
```diff
@@ -3398,6 +3398,72 @@ namespace lime {
...
+
+		}
+
+		varray* list = hl_alloc_array (&hlt_bytes, count);
+		vbyte** listData = hl_aptr (list, vbyte*);
+
+		while (*result != '\0') {
+
+			int length = strlen (result) + 1;
+			char* _result = (char*)malloc (length);
```
player-03, line 3452: Don't we need a `free()` call to go with this?

dtwotwo: > Don't we need a `free()` call to go with this?

Maybe `hl_alloc_bytes` would be safer instead of `malloc`, since we can't call `free()` immediately?

ACrazyTown: I've added a `free()` call. I've tested it a bit and it seems to work fine, but I will say again that my C(++) is a bit rusty so please do tell if I've overlooked something!!

I will also note that there's other `malloc()` calls in the OpenAL bindings that do not have a corresponding `free()` call. Perhaps this should also be handled in a seperate issue?

https://github.com/openfl/lime/blob/36d795369dbdb18a442825c8484466e62ed62289/project/src/media/openal/OpenALBindings.cpp#L2209-L2217

https://github.com/openfl/lime/blob/36d795369dbdb18a442825c8484466e62ed62289/project/src/media/openal/OpenALBindings.cpp#L3389-L3398

Raltyro:
oh, i didnt notice youve made a pull request about this too until no
ive made my own implementation for funkincrew's pull request not long ago.
https://github.com/FunkinCrew/lime/pull/39/changes

PR #1993, opened 2025-10-23
Allow executing arbitrary commands from project.xml.

### LINKED ISSUE #1959: Project prebuild command is running too late
I need to pre-process my assets for final builds so I have assets defined in a project.xml as follows. It will use `game/assets` for dev builds and `game/assets-final` for final builds -

```xml
<assets path="game/assets" rename="assets" unless="final" />
<assets path="game/assets-final" rename="assets" if="final" />
```

I have a script which I run to pre-process the assets, which will create the `game/assets-final` directory.

I wanted to run the script as a prebuild command in the project.xml -
```xml
<prebuild command="pre-process.sh" />
```

The problem is that lime's ProjectXMLParser will fail during project.xml parsing when it validates assets elements due to the directory not existing; `Error: Could not find asset path "game/assets-final"`. This is because the assets elements validation occurs before the prebuild commands are executed.

To workaround this I could make sure that `game/assets-final` exists before doing a final build, however this isn't ideal.

So the question is, can we change the tooling to run the prebuild commands before fully parsing the project.xml? Or perhaps add a preparse command which would run before the parsing?

player-03:
The `<execute />` tag immediately runs a single command, and optionally saves its output and/or exit code. Sample usage:

```xml
<section unless="display">
	<execute command="script.bat" if="windows" exitCode="scriptExitCode" />
	<execute command="./script.sh" unless="windows" exitCode="scriptExitCode" />
	<error value="Custom script failed!" if="${scriptExitCode}!=0" />

	<execute command="dir" if="windows" output="directoryContents" />
	<execute command="ls" unless="windows" output="directoryContents" />
	<echo value="Directory contents: ${directoryContents}" />
</section>
```

Checking `unless="display"` is optional but recommended. Without it, the commands will run every time code completion is initialized. Edit: that is, except in the case this is primarily designed to solve (generating assets). Since assets are parsed in display mode, any asset generation must also happen in display mode.

Closes #1959

### DIFF
--- a/src/lime/tools/ProjectXMLParser.hx
+++ b/src/lime/tools/ProjectXMLParser.hx
@@ -11,6 +11,7 @@ import lime.tools.HXProject;
 import lime.utils.AssetManifest;
 #end
 import sys.io.File;
+import sys.io.Process;
 import sys.FileSystem;
 #if (haxe_ver >= 4)
 import haxe.xml.Access;
@@ -1864,6 +1865,22 @@ class ProjectXMLParser extends HXProject
 
 				case "config":
 					config.parse(element, substitute);
+
+				case "execute":
+					var process:Process = new Process(element.att.command);
+					var exitCode:Int = process.exitCode(true);
+
+					if (element.has.output)
+					{
+						defines.set(element.att.output, process.stdout.readAll().toString());
+					}
+
+					if (element.has.exitCode)
+					{
+						defines.set(element.att.exitCode, Std.string(exitCode));
+					}
+
+					process.close();
 
 				case "prebuild":
 					parseCommandElement(element, preBuildCallbacks);


ACrazyTown:
Is this being considered for 8.4.0?

player-03:
We could merge it, I guess I was hoping for more feedback/testimonials.

And I guess we have a second. One thing I noticed is that both your case and the original issue that inspired this could be solved with a `<preupdate />` tag, for processing assets before they're copied. It wouldn't be as open-ended, but it's less likely to cause problems by automatically running during `display`.

Or we could just do this and document the `display` issue, either way.

ACrazyTown:
`<preupdate />` sounds nice as well, though personally I don't mind either as long as there's *some* way to do it lol

joshtynjala:
The `unless="display"` feels a little funky.

I like the idea of `<preupdate />`.

player-03:
> I like the idea of `<preupdate />`.

You said that last time too, now that I look at it. I've submitted a pull request for that.

PR #1997, opened 2025-11-12
[html5] Runtime scale check for HTML5Window

IriySoft:
Fixes the issue https://github.com/openfl/lime/issues/1983 - an HTML5 app doesn't respond to pixel ratio change in runtime.

Re-submit of this one: https://github.com/openfl/lime/pull/1984

Also requires changes in OpenFL: https://github.com/openfl/openfl/pull/2822


joshtynjala:
Commit 838b9254cdbe30718dcf30d1ede5f306499b9945 is a more robust solution that uses a media query to track changes (the resize event is not guaranteed to be dispatched for devicePixelRatio changes), and it updates the canvas size. (And it includes a native implementation too).

PR #1998, opened 2025-11-12
[HTML5] Added "outline:none" property to canvas/div style in HTML5Window.hx

IriySoft:
Had to add this to address the following issue: in some cases (mine was Mac + Chrome) a visible outline appears around the render area when a key on the keyboard is pressed. 

Looks like this:

Screenshot 2025-10-10 at 15 49 09

### DIFF
--- a/src/lime/_internal/backend/html5/HTML5Window.hx
+++ b/src/lime/_internal/backend/html5/HTML5Window.hx
@@ -140,6 +140,7 @@ class HTML5Window
 			var style = canvas.style;
 			style.setProperty("-webkit-transform", "translateZ(0)", null);
 			style.setProperty("transform", "translateZ(0)", null);
+			style.setProperty("outline", "none", null);
 		}
 		else if (div != null)
 		{
@@ -154,6 +155,7 @@ class HTML5Window
 			style.setProperty("-moz-user-select", "none", null);
 			style.setProperty("-ms-user-select", "none", null);
 			style.setProperty("-o-user-select", "none", null);
+			style.setProperty("outline", "none", null);
 		}
 
 		if (parent.__width == 0 && parent.__height == 0)


joshtynjala:
I'm not able to reproduce this issue on macOS 26.1 with Chrome 142 in any of my projects that can target HTML5.

I wonder if there's something in your site's CSS, or in your Haxe code that accesses JS APIs, that might be affecting this behavior.

IriySoft:
> I'm not able to reproduce this issue on macOS 26.1 with Chrome 142 in any of my projects that can target HTML5.
> 
> I wonder if there's something in your site's CSS, or in your Haxe code that accesses JS APIs, that might be affecting this behavior.

In fact, I couldn't reproduce it myself, but the customer (Crazygames) shared screenshots and videos and insisted on fixing this by adding `"outline:none"` property to the canvas' style. Then I realizes that this is only possible with the source code and the fix worked.

Maybe it is possible to access canvas and modify its style properties from outside of OpenFL to avoid the suggested source code changes?

IriySoft:
@joshtynjala I didn't find how to access canvas (yet), but notices that HTML5Window.hx has an empty setBorderless function. Probably this code could go there instead of modifying the default canvas style? Something like this:

```
public function setBorderless(value:Bool):Bool
	{
		if (canvas != null) {
				if (value) canvas.style.setProperty("outline", "none", null);
				else canvas.style.removeProperty("outline");
		}
		return value;
	}
```

joshtynjala:
> HTML5Window.hx has an empty setBorderless function. Probably this code could go there instead of modifying the default canvas style

No, I don't think that an appropriate place for it. The borderless property is for native windows, and it has nothing to do with focus rectangles (which is what the outline is). I believe that borderless removes the title bar and any other operating system chrome around the content.

> Then I realizes that this is only possible with the source code

It would definitely be possible with CSS. Something like this (replace openfl-content with the id, if you're using a different one):

```css
#openfl-content canvas {
    outline: none;
}
```

IriySoft:
OK, I guess I should close the pull request then. In any case, this shouldn't go to the default style properties.

PR #1999, opened 2025-11-14
Allow specifying Gradle plugins from project.xml.

player-03:
This allows including [published plugins](http://plugins.gradle.org/) or custom user-created ones. Since plugins can make arbitrary changes to the build script, this allows multiple extensions to add to _app/build.gradle_ without overwriting each other's changes.

https://docs.gradle.org/current/userguide/plugins.html
https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html

Sample usage: `<config:android gradle-apply-plugin="com.google.gms.google-services" />`

From a quick search, there seem to be two ways to include a plugin. `apply plugin` is the more backwards-compatible option, so I went with that. However, `plugin { }` could be more future-proof. Maybe at some point we'll switch.

My main concern for now is, is "gradle-apply-plugin" distinctive enough from "gradle-plugin"? Or would another name be clearer?

### DIFF
--- a/templates/android/template/app/build.gradle
+++ b/templates/android/template/app/build.gradle
@@ -4,7 +4,8 @@ import java.awt.GridBagConstraints
 import javax.swing.border.EmptyBorder
 
 apply plugin: 'com.android.application'
-
+::foreach ANDROID_GRADLE_APPLY_PLUGIN::apply plugin: '::__current__::'
+::end::
 System.setProperty('java.awt.headless','false')
 
 //Uncomment to debug deprecation warnings.
--- a/tools/platforms/AndroidPlatform.hx
+++ b/tools/platforms/AndroidPlatform.hx
@@ -437,6 +437,7 @@ class AndroidPlatform extends PlatformTarget
 		]);
 		context.ANDROID_GRADLE_VERSION = project.config.getString("android.gradle-version", "8.9");
 		context.ANDROID_GRADLE_PLUGIN = project.config.getString("android.gradle-plugin", "8.7.3");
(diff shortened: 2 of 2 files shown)


IriySoft:
> This allows including [published plugins](http://plugins.gradle.org/) or custom user-created ones. Since plugins can make arbitrary changes to the build script, this allows multiple extensions to add to _app/build.gradle_ without overwriting each other's changes.
> 
> https://docs.gradle.org/current/userguide/plugins.html https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html
> 
> Sample usage: `<config:android gradle-apply-plugin="com.google.gms.google-services" />`

There should be a classpath dependency added in the `build.gradle` too, otherwise Gradle won't be able to apply the plugin with "Plugin with id 'com.google.gms.google-services' not found." exception. That in turn requires adding the repositories paths to the app-level build.gradle (these are currently not present there). Something like this:

```
import groovy.swing.SwingBuilder
import java.awt.GridBagLayout
import java.awt.GridBagConstraints
import javax.swing.border.EmptyBorder

buildscript {
  repositories {
    google()
    mavenCentral()
  }

  dependencies {
    classpath 'com.google.gms:google-services:4.4.4'
  }
}

apply plugin: 'com.android.application'

. . .
```

player-03:
That's a lot of changes, and I'd rather keep the config options simple. Can't you create your own plugin to do all that?

https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html

IriySoft:
> That's a lot of changes, and I'd rather keep the config options simple. Can't you create your own plugin to do all that?
> 
> https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html

I can try this indeed, thanks!

PR #2002, opened 2025-12-01
HTML5Window: update context.attributes.hardware when creating context

ACrazyTown:
Follow up to #1902

If the project requests a hardware accelerated context, but the browser does not give it for whatever reason (WebGL disabled?), `context.attributes.hardware` would not update to reflect this. With this, `context.attributes.hardware` will properly update according to whether we have an WebGL context or not


joshtynjala:
Thanks!

PR #2005, opened 2025-12-08
Fix Emscripten

barisyild:
Fix Emscripten compile & rendering issues, implement HTTP requests, remove assets from virtual file system and remove curl and efsw to reduce file size

# Emscripten Status
- Rendering
    - Rendering [X] -> [✓]
    - Auto Size Canvas [X] -> [✓]
- Network
    - HTTP Request [X] -> [✓]
- Compile
	- Compile [X] -> [✓]
    - Optimizations [X] -> [✓]
- File Size Optimizations
    - Removing curl and efsw from Build [X] -> [✓]

### DIFF
--- a/project/src/ExternalInterface.cpp
+++ b/project/src/ExternalInterface.cpp
@@ -4421,9 +4421,7 @@ extern "C" int lime_vorbis_register_prims () { return 0; }
 extern "C" int lime_register_prims () {
 
 	lime_cairo_register_prims ();
-	#ifndef LIME_CURL
 	lime_curl_register_prims ();
-	#endif
 	lime_harfbuzz_register_prims ();
 	lime_openal_register_prims ();
 	lime_opengl_register_prims ();
--- a/project/src/ExternalInterface.cpp
+++ b/project/src/ExternalInterface.cpp
@@ -4421,7 +4421,9 @@ extern "C" int lime_vorbis_register_prims () { return 0; }
 extern "C" int lime_register_prims () {
 
 	lime_cairo_register_prims ();
+	#ifndef LIME_CURL
 	lime_curl_register_prims ();
+	#endif
 	lime_harfbuzz_register_prims ();
 	lime_openal_register_prims ();
 	lime_opengl_register_prims ();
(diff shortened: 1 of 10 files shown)


project/src/ExternalInterface.cpp
```diff
@@ -4421,7 +4421,9 @@ extern "C" int lime_vorbis_register_prims () { return 0; }
 extern "C" int lime_register_prims () {
 
 	lime_cairo_register_prims ();
+	#ifndef LIME_CURL
```
player-03: Sorry, why is this `ifndef`?

barisyild: debug artifact

barisyild:
Only rendering issues remain:

- Some areas are not being rendered
- Some areas are rendered with incorrect colors

barisyild:
> Only rendering issues remain:
> 
> * Some areas are not being rendered
> * Some areas are rendered with incorrect colors

Even though the project is using WebGL2 (OpenGL 3), this might be happening because the API being used is actually WebGL (OpenGL 2). I’m not very familiar with the behavioral differences between these two versions.

https://github.com/openfl/lime/blob/17fe04721d8bc4c8b18c1e3d4983b9371326e72d/project/src/graphics/opengl/OpenGL.h#L13-L19

barisyild:
Rendering fixed, waiting for merging.

Requires https://github.com/openfl/openfl/pull/2860 on OpenFL

tools/platforms/WebAssemblyPlatform.hx
```diff
@@ -511,6 +490,11 @@ class WebAssemblyPlatform extends PlatformTarget
 				System.mkdir(Path.directory(path));
 				AssetHelper.copyAsset(asset, path, context);
 			}
+            else
+            {
+                System.mkdir(Path.directory(path));
+                AssetHelper.copyAsset(asset, path, context);
+            }
```
player-03, line 497: Does WASM really need two copies of the non-template assets? You've got this here, and above on line 472 it also copies it into _obj/assets_.

barisyild: I'm not sure about that because it's been a while since I wrote the code; need to test it.

PR #2007, opened 2025-12-13
Fix eventloop usage for haxe 5

tobil4sk:
Haxe 5 nightlies have a new api for event loops: https://github.com/HaxeFoundation/haxe/pull/12388

See the equivalent heaps patch: https://github.com/HeapsIO/heaps/commit/d40f2d04cab5fb2b14cdaa7c40943d98ffd5ea97#diff-c0fa4e889efa97012174175edcce85cc359792651eb5e7435d2f0c0ed26dec0a


tobil4sk:
I think this needs to be updated too: https://github.com/openfl/lime/blob/17fe04721d8bc4c8b18c1e3d4983b9371326e72d/src/lime/system/JNI.hx#L470

Also, this will have the same issue as the [openfl dictionary size PR](https://github.com/openfl/openfl/pull/2801) since haxe 5 preview 1 didn't have this updated api yet.

player-03 (approved):

player-03:
Copying Josh's comment here so I can find it later:

> I think we can merge more Haxe 5 stuff once they update their version to include `preview.2`. right now, nightlies are still called `preview.1`, but some big things have changed since the real `preview.1` and we can't differentiate.

MRK2009:
Wait, I have a question, what about the warnings when using nightly Haxe version?

player-03:
If they're related to the event loop, we could add them to this PR, otherwise it would be a separate PR.

MRK2009:
@tobil4sk It got conflicted 💀

dimensionscape:
Is there anything development related blocking this merge?

player-03:
Nothing on our side AFAIK, just Josh's concern that Haxe might still make changes.

dimensionscape:
Ok fair, no rush then.

PR #2009, opened 2025-12-17
[BUGFIX] Run Macro Flags when using a local haxelib repository (Move flags to `include.xml`)

### LINKED ISSUE #2004: Bug: Embedded Sounds causes an exception with Haxelib local repo
Attempting to access embedded sounds while having a haxelib local repo causes it to return **"Exception has occured. Access Violation"** at [lime.media.AudioBuffer#176](https://github.com/openfl/lime/blob/develop/src/lime/media/AudioBuffer.hx#L176) (This error is displayed while debugging on HashLink. It still happens on cpp targets but there is no error as my debug server is being jank so instead it just abruptly closes the application.)

This was discovered on an OpenFL project and then tested on a standalone lime template (`haxelib run lime create SimpleAudio`)

TechnikTil:
## Related Issues
Fixes #2004

## Description
When compiling a `lime` project while using a local `haxelib` repository, the `extraParams.hxml` in `lime` is skipped.
This causes `AssetsMacro.embedBytes()` to never run on `haxe.io.Bytes`, causing #2004.

To fix this, I have moved those flags over to `include.xml`.

This is my first PR to here, so let me know if I should do something different in the future!

### DIFF
--- a/extraParams.hxml
+++ b/extraParams.hxml
@@ -1,2 +1,0 @@
---macro lime._internal.macros.DefineMacro.run()
---macro haxe.macro.Compiler.addMetadata("@:autoBuild(lime._internal.macros.AssetsMacro.embedBytes())", "haxe.io.Bytes")
--- a/include.xml
+++ b/include.xml
@@ -224,4 +224,7 @@
 
 	</section>
 
+	<haxeflag name="--macro" value="lime._internal.macros.DefineMacro.run()" />
+	<haxeflag name="--macro" value="haxe.macro.Compiler.addMetadata('@:autoBuild(lime._internal.macros.AssetsMacro.embedBytes())', 'haxe.io.Bytes')" />
+
 </extension>


player-03:
~~Odd. Looking at the generated hxml files, it generates `-cp /path/to/global/Haxelib/lime/8,3,0/src`. This implies that, for whatever reason, Lime deliberately bypasses the local repo, and includes the globally-installed copy.~~ Edit: No, that was a quirk of how I tested it. But the below stands.

Normally, `haxelib path` would include the contents of _extraParams.hxml_. However, this bypass code evidently does not. I'll see if I can track it down.

player-03:
Ok, found it. If a path override is defined, it won't call `haxelib path`. This happens [here](https://github.com/openfl/lime/blob/2032404ff659daf5de854a91a51971f1d9f81aaf/tools/CommandLineTools.hx#L1475) and [here](https://github.com/openfl/lime/blob/2032404ff659daf5de854a91a51971f1d9f81aaf/tools/CommandLineTools.hx#L2098) in this case, but can happen for any number of reasons.

https://github.com/openfl/lime/blob/2032404ff659daf5de854a91a51971f1d9f81aaf/src/lime/tools/HXProject.hx#L1240-L1265

Since this can happen to more libraries than just Lime, I think the correct solution is to include _extraParams.hxml_, not to move the extra params on a case-by-case basis.

TechnikTil:
> Since this can happen to more libraries than just Lime, I think the correct solution is to include _extraParams.hxml_, not to move the extra params on a case-by-case basis.

Should I edit the PR to do so then?
Or should I just close this

player-03:
Check out #2010 and if that fixes it for you, we can close this one.

TechnikTil:
Yup that works perfectly, thank you so much!

PR #2010, opened 2025-12-17
Include extraParams.hxml when library path is overridden.

### LINKED ISSUE #2004: Bug: Embedded Sounds causes an exception with Haxelib local repo

player-03:
When including a haxelib with an overridden path (such as Lime itself, when using a local Haxelib repo), Lime fails to copy _extraParams.hxml_, potentially breaking compilation. This checks for the file and copies all non-comment lines if found.

Closes #2004, supersedes #2009.

### DIFF
--- a/src/lime/tools/HXProject.hx
+++ b/src/lime/tools/HXProject.hx
@@ -1241,7 +1241,7 @@ class HXProject extends Script
 			{
 				var path = Haxelib.pathOverrides.get(name);
 				var jsonPath = Path.combine(path, "haxelib.json");
-				var added = false;
+				var extraParamsPath = Path.combine(path, "extraParams.hxml");
 
 				try
 				{
@@ -1262,6 +1262,18 @@ class HXProject extends Script
 				var param = "-cp " + path;
 				compilerFlags.remove(param);
 				compilerFlags.push(param);
+
+				try
+				{
+					if (FileSystem.exists(extraParamsPath))
+					{
+						var extraParams = Lambda.filter(File.getContent(extraParamsPath).split("\n"),
+							function(param:String) return param.length > 0 && param.charCodeAt(0) != "#".code);
+
+						compilerFlags = ArrayTools.concatUnique(compilerFlags, extraParams);
+					}
+				}
+				catch (e:Dynamic) {}
 			}
 			else
 			{


tobil4sk:
It would be ideal if we could pass this onto haxelib, then we wouldn't have to worry about missing behaviour in future

player-03:
Agreed, that would be ideal.

But as far as I know, you can't pass a file path to `haxelib path`, only a library name. Nor is there a different command that takes a path. So using Haxelib would mean getting rid of `pathOverrides` entirely, and I assume it's there for a reason.

Tutez64:
Closing #2101 as a duplicate of this. The extras versus this branch (trim, skip `-L`, `-D name` without `=`, matching the `haxelib path` branch) are on https://github.com/player-03/lime/pull/10

player-03:
Having looked closer at Haxelib's implementation, it doesn't look like they have any interest in supporting custom paths. But I came up with a couple other options.

- On *nix systems, we could add `HAXELIB_PATH=custom/path` before running `haxelib path lib_name` to get it to find the directory we wanted. However, AFAIK there's no equivalent on Windows. (Short of setting `HAXELIB_PATH` globally, then reverting it, which is dangerous.)
- Haxelib supports local repos named ".haxelib". However, it will use the local repo exclusively and will fail to find anything installed globally (which is a problem if that's where Lime is). We could work around this in various convoluted ways, but ultimately, local repos aren't meant for this use case.
- Presumably if we asked the Haxelib developers, they would tell us that this is what `haxelib dev` is for. And that would certainly work, but it still involves overwriting an existing library's path, even if only for a moment.

My best idea so far is creating a local repo in _Export/.haxelib_, and using `haxelib dev` inside that, followed by `haxelib path`. Since it's in _Export_, Haxelib will only find it if we `setCwd()` first, otherwise it'll use the global repo.

PR #2012, opened 2025-12-29
Add allowNativeHeapPointerTagging attribute to application

rainyt:
When the compiled version is greater than 29, memory access conflicts may occur, causing the application to crash.
```
E/libEGL: validate_display:547 error 3008 (EGL_BAD_DISPLAY)
```


player-03:
We have that already.

https://github.com/openfl/lime/blob/2032404ff659daf5de854a91a51971f1d9f81aaf/tools/platforms/AndroidPlatform.hx#L499

PR #2016, opened 2026-01-11
Fix perspective camera

Teslotik:
In a perspective projection matrix, the homogeneous coordinate w must depend on z (depth) to enable correct perspective division.  
This ensures objects farther from the camera appear smaller and depth is properly mapped.

I found a useful resource explaining this in detail, which you can check here:
https://www.scratchapixel.com/lessons/3d-basic-rendering/perspective-and-orthographic-projection-matrix/opengl-perspective-projection-matrix.html
The issue with `w` is explained in the beggining of the article.

In the attached GIF, I demonstrate the issue with a 3D scene - although it looks like the camera is rotating around the origin, the real problem lies in the incorrect perspective projection formula.

chrome_G1WVXoEk3k


moxie-coder:
50f1628a-b78a-48b0-bba5-4807399749fd_text

PR #2021, opened 2026-01-13
[hxp] Fix compatbility with haxe 3

tobil4sk:
haxe 3 compatibility with hxp projects was broken in:: 22032ef6b51ed9e3e0f4c865cf65fd9d515abaa3

This is because haxe 3 doesn't support --main, only -main.

This gave [errors](https://github.com/openfl/hxp/actions/runs/18919536447/job/54011696998#step:7:10) like:
```
$ haxelib run lime build neko
Error: : unknown option '--main'.
```

### DIFF
--- a/src/lime/tools/HXProject.hx
+++ b/src/lime/tools/HXProject.hx
@@ -452,7 +452,7 @@ class HXProject extends Script
 		{
 			#if (lime && !eval)
 			var nekoOutput = Path.combine(tempDirectory, name + ".n");
-			System.runCommand("", "haxe", args.concat(["--main", "lime.tools.HXProject", "-neko", nekoOutput]));
+			System.runCommand("", "haxe", args.concat(["-main", "lime.tools.HXProject", "-neko", nekoOutput]));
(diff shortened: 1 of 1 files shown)


player-03:
According to the documentation, `--main` is correct, so I'd argue `-main` is less future-proof.

That said, `-m` is also officially supported. Does Haxe 3 recognize that?

tobil4sk:
> I'd argue -main is less future-proof.

Development versions of Haxe 5 still support `-main`, and there doesn't seem to be a push to get rid of it. For example, the haxe extension for vscode still generates [templates using `-main`](https://github.com/vshaxe/vshaxe/blob/352db8ee58c87de513c820321be038565dbe91ad/scaffold/project/build.hxml#L3). I think if `-main` ever does stop working, haxe 3 support will probably be gone by then too.

> That said, -m is also officially supported. Does Haxe 3 recognize that?

Seems like it doesn't:
```
$ ./haxe -m Main.hx
Error: : unknown option '-m'.
Haxe Compiler 3.4.7 - (C)2005-2017 Haxe Foundation
 Usage : haxe -main <class> [...]
```

Issue #2018, opened 2026-01-12
LIme8.3.0 Sound.loadCompressedDataFromByteArray Wrong


rainyt:
<img width="600" height="158" alt="Image" src="https://github.com/user-attachments/assets/e4878e90-ae3e-4815-87f0-faf773823625" />

After testing, the 8.0.2 version did not have this issue until it occurred in the 8.3.0 version.

rainyt:
[sound-loadCompressedDataFromByteArray.zip](https://github.com/user-attachments/files/24577261/sound-loadCompressedDataFromByteArray.zip)
Here is an example that can reproduce the problem.
```
OpenFL 9.5.0
Lime 8.3.0
```

joshtynjala:
In your example project, you need `urlloader.dataFormat = BINARY;`

Otherwise, the value of `urlloader.data` will be a `String` instead of a `ByteArray`.

rainyt:
Thanks,It seems to be another issue, I will check the other issues.

rainyt:
<img width="489" height="428" alt="Image" src="https://github.com/user-attachments/assets/e69cabb6-1519-4287-85d0-890b9c39686b" />

hi, I noticed that this file was deleted in version 8.3.0, Resulting in some differences in binary data processing. Especially for Sound.loadCompressiedDataFromByteArray, it has an impact as some audio may not load properly using this method. emm, But I'm not sure why this was deleted yet.

rainyt:
oh, Does it seem like it can only be used correctly for haxe5? If the user is using haxe4.3 version, this may cause issues. https://github.com/openfl/lime/pull/1765

rainyt:
[sound-loadCompressedDataFromByteArray-error2.zip](https://github.com/user-attachments/files/24603319/sound-loadCompressedDataFromByteArray-error2.zip)

I reproduced the issue using an audio that could trigger the error

Must use  Haxe version 4.3, reproducible without the haxe.io.bytes file in Lime.

PR #2023, opened 2026-01-14
haxe.io.Bytes source code compatible with Haxe4

rainyt:
Issues: https://github.com/openfl/lime/issues/2018

If compiled using Haxe5, the haxe.io.Bytes file is not required.

### DIFF
--- a/haxe4/haxe/io/Bytes.hx
+++ b/haxe4/haxe/io/Bytes.hx
@@ -1,0 +1,1192 @@
+/*
+ * Copyright (C)2005-2018 Haxe Foundation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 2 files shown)


rainyt:
Although Haxe5 has been improved to handle bytes correctly, it still handles them incorrectly in Haxe4. Therefore, when compiling with Haxe4.3 version, the Bytes.hx compatibility file of Haxe4 should be retained.

player-03:
I'm still not a fan of shadowing a core class. If a core class is messing with your use case, you could shadow it yourself, or request that Haxe release a patch.

tobil4sk:
I agree, shadowing the class doesn't seem like the correct fix here. It might be worth figuring out what exactly is the root cause of the broken processing, perhaps there is a nicer fix for it than shadowing a core type.

rainyt:
> I'm still not a fan of shadowing a core class. If a core class is messing with your use case, you could shadow it yourself, or request that Haxe release a patch.

Okay, I don't think Haxe4 will specifically update a byte patch to 4.3.8 anymore? Nevertheless, it has already affected users before version 4.3.7, although this is a very subtle error that may not be encountered by all users.

My main consideration is whether downward compatibility is needed. If it is no longer needed, then this PR can be closed. Thank you!

joshtynjala:
I found a more targeted fix for issue #2018 that doesn't require us to shadow `haxe.io.Bytes`.

9b2be25b31c1852a4cf297f3aa02395c167ec622

tobil4sk:
> I found a more targeted fix for issue https://github.com/openfl/lime/issues/2018 that doesn't require us to shadow haxe.io.Bytes.

In that case it looks like haxe 4 behaviour was correct, it makes sense for getString to throw if the byte sequence isn't valid utf-8 (unless the old decoder was broken). Maybe it would be good to have a comment explaining what is going on there with the empty catch block.

joshtynjala:
> Maybe it would be good to have a comment explaining what is going on there with the empty catch block.

Okay, I'll add one!

rainyt:
Yeah, That's great, so there's no need to keep this, thank you!

PR #2025, opened 2026-01-25
Bytes: Fix length from fromBytes on JS

Geokureli:
Potentially related to: https://github.com/openfl/openfl/issues/2360

In JS the underlying byte data may have extras, due to the underlying types used in js, explained here: https://github.com/HaxeFoundation/haxe/issues/8974. This pr will ignore those extras.

Example reproducing the inconsistent behavior:
```hx
final output = new haxe.io.BytesOutput();
output.writeString("test");
final byteArray = lime.utils.Bytes.fromBytes(output.getBytes());
trace('byteArray.length: ${byteArray.length}'); // 16 on JS, 4 otherwise
```

Tested on chrome for mac

### DIFF
--- a/src/lime/utils/Bytes.hx
+++ b/src/lime/utils/Bytes.hx
@@ -19,7 +19,7 @@ abstract Bytes(HaxeBytes) from HaxeBytes to HaxeBytes
 	public function new(length:Int, bytesData:BytesData)
 	{
 		#if js
-		// bytesData may have extra bytes
+		// bytesData may have extra bytes. See https://github.com/HaxeFoundation/haxe/issues/8974
 		this = new HaxeBytes(bytesData.slice(0, length));
 		#elseif hl
 		this = new HaxeBytes(bytesData, length);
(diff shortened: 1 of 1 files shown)


player-03:
> Tested on chrome for mac

I also got 16 in Firefox on Linux.

Geokureli:
I also added [a fix](https://github.com/HaxeFoundation/haxe/pull/12533) to haxe, but lime should probably account for this on older versions

Edit: The haxe PR was rejected; I'm still working towards a fix for https://github.com/HaxeFoundation/haxe/issues/8974

src/lime/utils/Bytes.hx
```diff
@@ -19,7 +19,8 @@ abstract Bytes(HaxeBytes) from HaxeBytes to HaxeBytes
 	public function new(length:Int, bytesData:BytesData)
 	{
 		#if js
-		this = new HaxeBytes(bytesData);
+		// bytesData may have extra bytes. See https://github.com/HaxeFoundation/haxe/issues/8974
+		this = new HaxeBytes(bytesData.slice(0, length));
```
tobil4sk: [`slice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice) copies the buffer, so perhaps another option might be to just manually override: `this.length = length`? (might require @:privateAccess)

Geokureli: doesn't require @:privateAccess as the abstract has `@:access(haxe.io.Bytes)`

PR #2032, opened 2026-02-16
[ci] Set HXCPP_COMPILE_CACHE without "~"

tobil4sk:
The quotes used previously prevented ~ from being expanded by the shell, which broke the path on some platforms.

This allows the cache to function properly and should speed up build times.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,6 +3,7 @@ on: [push, pull_request]
 
 env:
   HAXE_VERSION: 4.2.5
+  HXCPP_COMPILE_CACHE: ${{ github.workspace }}/.hxcpp_cache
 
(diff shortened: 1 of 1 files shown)


tobil4sk:
Some comparisons (with last [develop run](https://github.com/openfl/lime/actions/runs/21785992963/job/62857432265)):

| | before | after |
| - | - | - |
| macos | 44 mins | 37 mins |
| android | 27 mins | 15 mins |
| ios | 47 mins | 34 mins |
| linux | 13 mins | 10 mins | 
| linux-arm64 | 9 mins | 8 mins |
| windows | 26 mins | 25 mins |

No change on windows because that was already using a working cache path. Some of it could be down to random fluctuations, but it does seem to help.

PR #2039, opened 2026-03-08
Fix GL objects not getting GC'd on HL

ACrazyTown:
On Hashlink, prior to this change, no OpenGL objects would be cleaned up by the GC. 

This was caused because the object was keyed by its `void*` pointer, but the finalizer would receive a `HL_CFFIPointer*` and thus the map check would fail. To fix it I added a HL specific GC method that takes in a `HL_CFFIPointer*` and then retrieves the actual object pointer, to use for the map checks.


joshtynjala:
Good catch!

PR #2040, opened 2026-03-10
make logging functions dynamic

Vortex2Oblivion:
makes functions dynamic so you can override them with your own behavior

### DIFF
--- a/src/lime/utils/Log.hx
+++ b/src/lime/utils/Log.hx
@@ -11,7 +11,7 @@ class Log
 	public static var level:LogLevel;
 	public static var throwErrors:Bool = true;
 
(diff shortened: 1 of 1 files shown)


Vortex2Oblivion:
openfl/openfl#2846 will also need to be merged

player-03:
Other approaches to consider:

- If you just need to override this class for your own project, you could shadow it. That is, add a file named _lime/utils/Log.hx_ to your classpath, and put whatever you need in there. It will then be used instead of Lime's version, even by Lime classes.
- If you want to suppress log messages, you can do that with `level`.
- If you want to print to somewhere else, we could make a variable like `public static var println:String -> Void` that you could set, and all functions would call.

Issue #1335, opened 2019-06-21
Can't compile iOS for Distribution


aW4KeNiNG:
Recently I updated Lime from 6.2.0 to 7.5.0. With the new version I can't compile for distribution. When I execute `openfl build ios -final`, the error is:

`error: Provisioning profile "XXXX distribution" doesn't match the entitlements file's value for the get-task-allow entitlement.`

If I open the xcode project, I can see the entitlements file with `get-task-allow = true`. I think that value should be only included in development builds and it is the origin problem in distribution builds.

I'm using XCode 10.2.1

PR #1854, opened 2024-10-23
Generic Linux ARM and Windows ARM64 Support

ghost:
note that this is NOT tested, [this pr on hxcpp](https://github.com/HaxeFoundation/hxcpp/pull/1159) needs to be merged and Raspberry Pi support needs to be reworked with this codes.

PR #2050, opened 2026-04-14
feat: Add Windows ARM64 (MSVC) support

FixFlare:
# Add Windows ARM64 and WinRT build support

This started as just ARM64 Support but I kept going until I had my game running on my Xbox One. See the second half of this post for the winRT portion.

## Description

This PR adds native compilation support for Windows ARM64 targets, allowing Lime projects to be cross-compiled for devices with ARM64 processors (Snapdragon X Elite laptops, Surface Pro X, Windows Dev Kit 2023, etc).

Usage follows the same convention as other platforms:

```
lime rebuild windows -arm64          # Rebuild Lime native libraries for ARM64
lime build windows -arm64            # Build a project targeting ARM64
lime rebuild windows -arm64 -static  # Static linking (single .exe output)
```

This should match the `-arm64` flag behavior on macOS, iOS, and Android.

## Prerequisites

Developers cross-compiling from x64 must install the **MSVC ARM64 build tools** via the Visual Studio Installer:

- Open Visual Studio Installer → Modify → Individual Components
- Check **"MSVC Build Tools for ARM64/ARM64EC (Latest)"**

#### `tools/platforms/WindowsPlatform.hx`

- Added `isArm` flag to detect `Architecture.ARM64` targets
- Updated output directory logic to use `WindowsArm64` paths (consistent with `MacArm64`, `LinuxArm64`)
- Adjusted the `rebuild()` method which was unconditionally passing `-DHXCPP_M64` for all 64-bit targets, ignoring the `-arm64` flag entirely. Now it passes `-DHXCPP_ARM64` when ARM64 is targeted
- Injects `-DHXCPP_ARM64` into the hxcpp compiler flags for both HL and cpp target types

#### `tools/CommandLineTools.hx`

- Added `WindowsArm64/` to the Neko module loader path resolution for Windows ARM64 hosts (matching the existing `MacArm64/` and `LinuxArm64/` pattern)
- Updated the `-arm64` help text to include `windows` alongside `mac|ios|android`

#### `project/Build.xml`
- **Decoupled `LIME_OPENALSOFT` from the `static_link` restriction:** Removed `unless="static_link"` from the desktop platform OpenAL Soft selection logic. This allows static binaries to use the more stable OpenAL Soft backend instead of being forced into MojoAL. Prior to this change, running a Windows ARM build on Linux via wine would freeze.
- **Implemented Priority-Based Audio Logic:** 
    - Updated `LIME_MOJOAL` to only trigger for static builds `unless="LIME_OPENALSOFT || LIME_OPENAL"`. 
    - This ensures that if a stable/native OpenAL backend is available (like OpenAL Soft for Windows ARM64), it is selected over the MojoAL shim.
    - Added an `unset` safety for `LIME_OPENALSOFT` when `LIME_MOJOAL` is explicitly selected (e.g., for Switch) to prevent duplicate symbol linker errors.
- Consolidated the selection of the generic `LIME_OPENAL` flag to trigger based on either backend, this should simplify downstream feature detection in the build script.

#### `project/lib/custom/openal/include/config-windows-arm64.h` [NEW]

- New OpenAL Soft configuration header for Windows ARM64 targets
- Defines appropriate API attributes, available backends (WinMM, DirectSound, WaveFile), and processor feature flags

#### `project/lib/custom/openal/include/config.h`

- Added conditional include of `config-windows-arm64.h` when `HXCPP_ARM64` is defined on Windows

#### `project/lib/openal-files.xml`

- Excluded `mixer_neon.cpp` on Windows (`unless="windows"`). This file uses GCC/Clang-specific NEON struct initialization syntax (`{ a, b, c, d }` on `float32x4_t`) that MSVC rejects with `error C2078: too many initializers`. OpenAL cleanly falls back to `mixer_c.cpp`, which MSVC auto-vectorizes for ARM64

#### `project/lib/custom/pixman/config.h`

- Disabled `USE_SSE2`, `USE_SSE3`, and `X86_MMX` when `HXCPP_ARM64` is defined on Windows. These are x86-specific SIMD intrinsics that are unavailable on ARM64

#### `project/lib/custom/mojoal/mojoal.c` [NEW]

- Added a patched version of `mojoal.c` to resolve MSVC ARM64 compatibility issues:
  - Added `_alloca` shim for MSVC.
  - Fixed ARMv8 NEON detection to exclude MSVC (forcing the scalar path which MSVC auto-vectorizes, avoiding MSVC-incompatible GCC vector operators).
  - This allows the `project/lib/mojoal` submodule to remain clean and untouched.

#### `project/lib/mojoal-files.xml`

- Updated to use the custom patched `mojoal.c` specifically when building for ARM64 (`if="HXCPP_ARM64"`), ensuring the build succeeds without requiring modifications to subprojects.

## Testing

- [x] `lime rebuild tools` — CLI rebuilds successfully
- [x] `lime rebuild windows -arm64 -verbose` — Native libraries (ndll) compile and link for ARM64 with zero errors
- [x] `lime rebuild windows -arm64 -static -verbose` — Static library variant compiles successfully
- [x] Verified output goes to `ndll/WindowsArm64/` directory (not `Windows64/`)
- [x] Verified linker uses `-machine:ARM64` (not `-machine:x64`)
- [x] Existing `lime rebuild windows -verbose` (x64) is unaffected
- [x] Tested my WIP game and was able to run on Windows ARM w/ sound working.

## Notes

- The NEON-related changes (`mixer_neon.cpp` exclusion, mojoal.c fix) only affect Windows ARM64 builds. All other platforms (Linux ARM64, macOS ARM64, iOS, Android) continue to use their existing NEON code paths via GCC/Clang, which fully support the vector subscript syntax
- No changes that should affect Raspberry Pi, Android, macOS, iOS, or standard Windows x64 builds

# WinRT Support

This is my attempt to complete the WinRT (Universal Windows Platform) [milestone](https://github.com/openfl/lime/milestone/11) in Lime. This allows installation to **Xbox One / Series** consoles and **Windows ARM64** devices. Was pretty cool to see my game running on my Xbox One.

**Important** This relies on an upstream PR being merged to HXCPP, [PR #1335](https://github.com/HaxeFoundation/hxcpp/pull/1335)

#### OpenGL Function Pointer Resolution (Fixed Startup Crash)

`OpenGLBindings::Init` had an early return for WinRT that bypassed the extension loader. On Xbox/UWP, this resulted in all OpenGL function pointers being `NULL`, leading to a `0xC0000005` (Access Violation) as soon as the engine tried to render. By removing this return, ANGLE correctly resolves pointers to D3D11, allowing rendering to work perfectly.

#### UWP Sandbox & Xbox Hardening

- **Path Resolution**: Redirected `HAXE_ROOT` to `ApplicationData::Current->LocalFolder` in the `Main.cpp` template. I believe this is probably required for Xbox, where the install directory is read-only.
- **SDL Sync**: Patched `SDL_syscond_cv.c` to force native Condition Variables, avoiding restricted Win32 semaphore APIs.
- **Stubs**: Added shims for Cairo and SDL to handle missing Win32 APIs like `CreateFileW`.
- **ANGLE Integration**: Included pre-compiled ANGLE binaries (`libEGL.dll`, `libGLESv2.dll`) to provide out-of-the-box OpenGL ES translation to D3D11, which is required for WinRT/Xbox hardware.

#### Licensing & Redistribution

The included **ANGLE** binaries are licensed under the **BSD 3-Clause License**. This is fully compatible with Lime's **MIT License**, as both allow redistribution of binaries. A copy of the ANGLE license is included at `dependencies/angle/LICENSE` to ensure compliance.

#### Asset Pipeline & UX Stabilization

- **Persistence**: Re-ordered the `WindowsPlatform` lifecycle to protect custom icons from being wiped during `-rebuild` operations.
- **Aspect-Aware Icons**: Implemented a "Best Fit" selection algorithm for icons and splash screens. It now automatically selects the source image from `project.xml` with the closest aspect ratio to prevent distorted images and icons.
- **Windowing & FPS**: Added support for custom window dimensions (e.g., Portrait mode) and unlocked FPS settings beyond the hardcoded 60fps cap.
- **Automated Packaging & Signing**:
  - Added support for the `-msix` target flag.
  - Implemented automated discovery of Windows SDK tools (`MakeAppx`, `SignTool`).
  - Integrated native signing that respects `<certificate path="..." password="..." />` in `project.xml`, replacing the legacy manual PowerShell workflow.

### Usage

```powershell
# 1. Rebuild Lime for WinRT (x64)
lime rebuild lime winrt -static

# 2. Build and Package for Xbox
lime build project.xml winrt -static -msix -rebuild

# 3. Build and Package for ARM64 (Surface Pro X, etc.)
# Rebuild the engine first
lime rebuild lime winrt -static -arm64
# Then build the game
lime build project.xml winrt -static -msix -rebuild -arm64
```

### Testing & Verification

- **Verified on Xbox One Hardware**: Successfully deployed and executed via Xbox Device Portal. Confirmed stable rendering, WASAPI audio, and writable storage.
- **Portrait & High-DPI**: Verified that the application correctly launches in portrait mode on Windows 11 with working icons.
- **ARM64 Compilation**: Verified the full pipeline from rebuild to MSIX packaging for ARM64 targets on an x64 host.


joshtynjala:
> Decoupled LIME_OPENALSOFT from the static_link restriction: Removed unless="static_link" from the desktop platform OpenAL Soft selection logic. This allows static binaries to use the more stable OpenAL Soft backend instead of being forced into MojoAL. Prior to this change, running a Windows ARM build on Linux via wine would freeze.

Just adding some additional context here. As I understand it, this restriction exists for **legal** reasons, rather than technical reasons. OpenAL-Soft is LGPL licensed, which can be dynamically linked without affecting the license of an application. However, if you want to statically link it, the application must be LGPL too, or the application must be distributed with symbols for relinking with a modified version of the library. That's not a requirement that we want to force on all users of Lime.

tobil4sk:
The `mojoal.c` change should happen upstream rather than by overriding it here

FixFlare:
Hey guys - thanks for the feedback on this. 

I hadn't considered any of the legal aspects of this. Trying to get mojoal working though really sent me down a rabbit hole with debugging the sdl sound playback.

In my original post I mentioned needing to force OpenAL Soft because static ARM64 builds would freeze on launch via Wine

After adding way too many log statements, I found the true cause of that freeze - it was actually a memory alignment mismatch in the OGG decoder (stb_vorbis) on ARM64-MSVC. The structs were being padded incorrectly which caused the sample_rate to be read back as 0. That 0 was what was actually hanging the audio initialization and making it look like a MojoAL stall. (divide by 0 problem)

@tobil4sk: I reverted all manual source-code patches to mojoal.c since I found the true cause of the freeze and was able to resolve the alignment mismatch by just using a /Zp8 compiler flag in the sdl_sound build script. This ensures the structs are packed correctly on ARM64 without tainting the submodule code.

@joshtynjala: Now that the decoder bug is resolved and MojoAL is stable, I updated the logic to prioritize MojoAL for static ARM64 builds. I believe this should resolve the legal issue but let me know if you believe otherwise.

project/lib/custom/pixman/config.h
```diff
@@ -293,14 +293,14 @@
 /* #undef USE_OPENMP */
 
 /* use SSE2 compiler intrinsics */
-#if defined(HX_WINDOWS) || (defined(HX_MACOS) && !defined(HXCPP_ARM64)) || (defined(HX_LINUX) && !defined(RASPBERRYPI) && !defined(HXCPP_ARM64))
+#if (defined(HX_WINDOWS) && !defined(HXCPP_ARM64)) || (defined(HX_MACOS) && !defined(HXCPP_ARM64)) || (defined(HX_LINUX) && !defined(RASPBERRYPI) && !defined(HXCPP_ARM64))
```
tobil4sk: Here `!defined(HXCPP_ARM64)` is common to all of these now, so let's factor that out (same below)

FixFlare: Good catch, that's much cleaner - factored out the !defined(HXCPP_ARM64) check here and in the blocks below as you suggested.

MRK2009 (approved):
Looks good to me anyway.

FixFlare:
@tobil4sk Please take a look at my latest changes if you would please. I added winrt support.

I thought about doing another PR but it relied on this being merged first.

I also wasn't too sure about including the angle binaries in the commit. I've included their license text which should make it all above board as far as I know.

tobil4sk:
Winrt should go in a separate PR since it is not related to windows arm64 support.

Aside from license issues, we really want to avoid checking binaries into the repository because it increases the size of the `.git` folder and cannot be reversed.

FixFlare:
@tobil4sk   - it kinda does though.

The modern way of packaging windows apps for the windows store is to build for all architectures and package them into a .msixbundle file.

That way when someone downloads it from the store it can download just the architecture they need. Kinda similar to the way android works these days.

Fair enough on the binaries but then how would you suggest we handle that? Just something in the readme telling them to download it from the angle repo? some sort of setup script that downloads them?

tools/platforms/WindowsPlatform.hx
```diff
@@ -188,7 +189,12 @@ class WindowsPlatform extends PlatformTarget
 
 		for (architecture in project.architectures)
 		{
-			if (architecture == Architecture.X64)
+			if (architecture == Architecture.ARM64)
+			{
+				isArm = true;
```
MRK2009, line 194: Hmm Why not just do `isArm64` instead of `isArm`, when you're targeting for 64-bit architecture only?
I'm feeling confused rn.

FixFlare: Existing conventions really. I used isArm instead of isArm64 to maintain consistency with Lime's existing naming conventions for Windows target directories. Lime uses the pattern WindowsArm64 (and WinRTArm64), meaning the architecture suffix is constructed as (isArm ? "Arm" : "") + (is64 ? "64" : ""). Introducing an isArm64 variable would break that concatenation logic without a larger refactor.

Although, it looks like Haxe 5.0 is dropping 32-bit support? It might be worth considering a larger refactor for that. Didn't really feel like it was my place to make that decision though.

MRK2009: Ah okay, now I understand.

tobil4sk: > Although, it looks like Haxe 5.0 is dropping 32-bit support?

Haxe 5.0 will no longer provide Windows 32-bit binaries of the compiler, but 32 bit support in hxcpp is not being removed

FixFlare: @tobil4sk - ahh ok, thanks for the clarification, makes sense.

I removed the angle binaries and added a `lime setup winrt` command that downloads the angle binaries from nuget and extracts them to the proper platform directories.

I also realized, since it is so old, that the arm folder inside of that nuget package is for arm32, not arm64, but Xbox requires the x64 ones anyways. Microsoft hasn't released a new package since 2016, but I added logic to download the latest version should they ever do so.

I'm not currently aware of any place hosting up-to-date standalone angle binaries besides this. If anyone has suggestions, let me know. Chromium builds have it in the zip but it's like 300MB and downloading that to get a 2MB .dll did not feel practical. Even though they're from 2016 though, runs just fine on my Xbox one.

MRK2009:
Lime 8.3.2 is out btw.

So you could try fix the conflicts as possible before 8.4.0.

FixFlare:
@MRK2009  Thanks for the heads up. I merged the latest `8.4.0-dev` branch and resolved the conflict.

The conflict in `Build.xml` happened because the base branch had some tweaks to the static link logic that overlapped with the audio priority changes from this PR. 

**Incoming Logic `8.4.0-dev`:**

```xml
<set name="LIME_OPENALSOFT" value="1" if="windows || linux || mac || android" unless="static_link" />
<set name="LIME_OPENAL" value="1" if="iphone || webassembly || tvos" />
<set name="LIME_MOJOAL" value="1" if="switch || static_link || winrt || mojoal" unless="LIME_OPENAL" />
<unset name="LIME_OPENALSOFT" if="LIME_MOJOAL" />
<set name="LIME_OPENAL" value="1" if="LIME_OPENALSOFT || LIME_MOJOAL" />
```

LIME_OPENAL is set twice under different conditions. Which makes it harder to track which backend is triggering the generic OpenAL flag.

**New Logic:**

```xml
<set name="LIME_MOJOAL" value="1" if="switch || winrt || mojoal || static_link" />
<set name="LIME_OPENALSOFT" value="1" if="windows || linux || mac || android" unless="LIME_MOJOAL" />
<unset name="LIME_OPENALSOFT" if="LIME_MOJOAL" />
<set name="LIME_OPENAL" value="1" if="LIME_OPENALSOFT || LIME_MOJOAL || iphone || webassembly || tvos" />
```

My thinking here is:

1. Establish `LIME_MOJOAL` as the definitive priority for static builds upfront.
2. Cleanly fall back to `LIME_OPENALSOFT` only if MojoAL wasn't triggered.
3. Consolidate the `LIME_OPENAL` setter into a single, easy-to-read line that covers the mobile platforms and desktop backends at once.

but just let me know if I'm missing something here.

I tested `lime rebuild windows -arm64 -static` after these changes and it is building fine still.

tobil4sk:
> @tobil4sk - it kinda does though.
\
The modern way of packaging windows apps for the windows store is to build for all architectures and package them into a .msixbundle file.

I see that it would be useful together, but to make it easier to review and quicker to merge it would be better to split the changes up into two PRs:
1. Windows arm64 support
2. WinRT support

Ideally we could add a windows arm64 build to CI, but it may be a challenge since haxe doesn't work on windows arm yet.

MRK2009:
I have noticed that someone tried to add both Linux and Windows ARM from pr https://github.com/openfl/lime/pull/1854 from the past, but failed anyway.
Well at least you're the one who's doing Windows ARM for Lime.

FixFlare:
@tobil4sk 

Noted for next time but there isn't really anything in the PR guidelines about that. I followed them to the best of my ability, and honestly, it has been so long already I'd have trouble splitting them up again. There would be a much higher chance of me making a mistake if I try that now.

This has been on the milestone list for 8 years. I think it would be worth trying to push it through and not reject it over that.

PR #2055, opened 2026-05-04
Feature: Add native MP3 decoding support via minimp3

MasterAlexS:
### Description
This PR adds native support for decoding and playing MP3 files in the C++ backend, matching the existing architecture used for WAV and OGG formats.

To keep the implementation lightweight and avoid heavy external dependencies, this uses the `minimp3` header-only library (which is already included in the ecosystem for the HashLink target).

### Changes made
* **Submodule:** Added `minimp3` as a Git submodule in `lib/minimp3`.
* **Build System:** Updated `Build.xml` to include the `LIME_MP3` flag and created `lib/mp3-files.xml` for the build toolkit configuration.
* **C++ Decoder:** Implemented `include/media/containers/MP3.h` and `src/media/containers/MP3.cpp` with the `Decode` method. It handles both memory buffers and file paths. Memory is safely mapped to Haxe's `ArrayBufferView` using `alloc_null()` with explicit byte assignment.
* **CFFI Integration:** Updated all four loading functions in `src/ExternalInterface.cpp` (`lime_audio_load_bytes`, `hl_audio_load_bytes`, `lime_audio_load_file`, `hl_audio_load_file`) to properly route MP3 checks to the new decoder.

### Testing
* Successfully compiled on Linux (CachyOS).
* Tested playback using a sample application calling `AudioBuffer.loadFromFile()` and successfully played the decoded MP3 data through `AudioSource`.
* [LimeMP3Testing.zip](https://github.com/user-attachments/files/27369944/LimeMP3Testing.zip)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -83,5 +83,9 @@
 	url = https://github.com/HaxeFoundation/hashlink
 	shallow = true
 [submodule "project/lib/sdl_sound"]
-	path = project/lib/sdl_sound
-	url = https://github.com/icculus/SDL_sound.git
+    path = project/lib/sdl_sound
+    url = https://github.com/icculus/SDL_sound.git
+[submodule "project/lib/minimp3"]
+    path = project/lib/minimp3
+    url = https://github.com/lieff/minimp3
+    shallow = true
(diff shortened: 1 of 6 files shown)


dimensionscape:
Lime 8.4.0 already introduces mp3 support and streaming along with support for formats like flac via sdl sound

PR #2057, opened 2026-05-08
Fix Linux window icon handling

Tutez64:
## Summary

This PR makes Linux builds honor `<icon path="...">` more like the other desktop targets.

Previously, Linux parsed project icons but did not generate an icon file in the application output, and the generated `ApplicationMain` did not apply an icon to the SDL window. As a result, Linux window managers could show the generic X11/default icon even when a project icon was configured.

This change:

- generates `bin/linux/bin/icon.png` from `project.icons`
- falls back to Lime’s default icon when no project icon is configured
- updates the generated Lime `ApplicationMain` to load `icon.png` from the application directory and call `window.setIcon(...)` on Linux

Windows and macOS already handle icons through their native packaging mechanisms, so this adds the missing Linux runtime path.

## Repro

[LimeLinuxIconRepro.zip](https://github.com/user-attachments/files/27539127/LimeLinuxIconRepro.zip)

### DIFF
--- a/templates/haxe/ApplicationMain.hx
+++ b/templates/haxe/ApplicationMain.hx
@@ -89,6 +89,17 @@ import ::APP_MAIN::;
 		}
 
 		app.createWindow(attributes);
+		#if (linux && sys)
+		var iconPath = lime.system.System.applicationDirectory + "/icon.png";
+		if (sys.FileSystem.exists(iconPath))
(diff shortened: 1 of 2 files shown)


MRK2009:
It got conflicted.

Tutez64:
I updated it.

PR #2059, opened 2026-05-11
SDLWindow: MSAA fallback retry + null-safe CreateWindow

### LINKED ISSUE #1782: Android emulator crashes with antialiasing >= 2
If I set antialiasing to 2 or higher in a project (e.g. the BunnyMark sample) the app crashes in the emulator I'm running with `lime test android -64 -debug`.

```xml
<window antialiasing="2" />
```

Using the `develop` branch of lime (c70ec9fbe02b8d98f6562ee7a403f812549ea387). Lime is built with ndk r15c (`lime rebuild android -64 -debug`) and with this patch: https://github.com/HaxeFoundation/hxcpp/pull/855.
The project is built with r21e.

Here is the popup:
image

Here is the crash dump:
```
********** Crash dump: **********
Build fingerprint: 'google/sdk_gpc_x86_64/emu64xa:13/TE2A.221127.024/11217414:userdebug/dev-keys'
#00 0x000000000087f1cd /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/liblime.so
SDL_GetWindowSize_REAL
/.../BunnyMark/lime/project/./lib/sdl/src/video/SDL_video.c:2057:0
#01 0x00000000008bae2a /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/liblime.so (SDL_GetWindowSize+46)
                                                                                                                                            SDL_GetWindowSize
                                                                                                                                            /.../BunnyMark/lime/project/./lib/sdl/src/dynapi/SDL_dynapi_procs.h:555:0
#02 0x0000000000116fb7 /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/liblime.so
lime::SDLWindow::GetWidth()
/.../BunnyMark/lime/project/./src/backend/sdl/SDLWindow.cpp:673:0
#03 0x000000000010540d /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/liblime.so
lime::lime_window_get_width(_value*)
/.../BunnyMark/lime/project/./src/ExternalInterface.cpp:3440:0
#04 0x00000000006f84a4 /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/libApplicationMain.so
lime::_internal::backend::native::NativeWindow_obj::__construct(hx::ObjectPtr<lime::ui::Window_obj>)
/.../BunnyMark/Export/android/obj/./src/lime/_internal/backend/native/NativeWindow.cpp:311:38
#05 0x0000000000701159 /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/libApplicationMain.so
lime::_internal::backend::native::NativeWindow_obj::__alloc(hx::ImmixAllocator*, hx::ObjectPtr<lime::ui::Window_obj>)
/.../BunnyMark/Export/android/obj/./src/lime/_internal/backend/native/NativeWindow.cpp:1095:10
#06 0x00000000004d5cd9 /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/libApplicationMain.so
lime::ui::Window_obj::__construct(hx::ObjectPtr<lime::app::Application_obj>, Dynamic)
/.../BunnyMark/Export/android/obj/./src/lime/ui/Window.cpp:195:38
#07 0x00000000004da591 /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/libApplicationMain.so
lime::ui::Window_obj::__alloc(hx::ImmixAllocator*, hx::ObjectPtr<lime::app::Application_obj>, Dynamic)
/.../BunnyMark/Export/android/obj/./src/lime/ui/Window.cpp:740:10
#08 0x000000000062ca40 /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/libApplicationMain.so
lime::app::Application_obj::_hx___createWindow(Dynamic)
/.../BunnyMark/Export/android/obj/./src/lime/app/Application.cpp:656:45
#09 0x000000000062c943 /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/libApplicationMain.so
lime::app::Application_obj::createWindow(Dynamic)
/.../BunnyMark/Export/android/obj/./src/lime/app/Application.cpp:245:50
#10 0x00000000006ebd0d /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/libApplicationMain.so
ApplicationMain_obj::create(Dynamic)
/.../BunnyMark/Export/android/obj/./src/ApplicationMain.cpp:137:20
#11 0x00000000006ea8e8 /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/libApplicationMain.so
ApplicationMain_obj::main()
/.../BunnyMark/Export/android/obj/./src/ApplicationMain.cpp:75:15
#12 0x0000000000758b8c /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/libApplicationMain.so
__hxcpp_main
/.../BunnyMark/Export/android/obj/./src/__main__.cpp:25:15
#13 0x0000000000758a12 /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/libApplicationMain.so (hxcpp_main+50)
                                                                                                                                                       hxcpp_main
                                                                                                                                                       /.../hxcpp/include/hx/HxcppMain.h:23:10
#14 0x00000000008950a6 /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/lib/x86_64/liblime.so (Java_org_libsdl_app_SDLActivity_nativeRunMain+882)
                                                                                                                                            Java_org_libsdl_app_SDLActivity_nativeRunMain
                                                                                                                                           /.../BunnyMark/lime/project/./lib/sdl/src/core/android/SDL_android.c:761:0
#15 0x000000000037d70b /apex/com.android.art/lib64/libart.so (art_quick_generic_jni_trampoline+219) (BuildId: 44bc52550248fc712a381bef164b75c7)
#16 0x0000000000368c95 /apex/com.android.art/lib64/libart.so (nterp_helper+165) (BuildId: 44bc52550248fc712a381bef164b75c7)
#17 0x000000000000b788 [anon:dalvik-classes4.dex extracted in memory from /data/app/~~eDbH_7qOlZkjF8j4l0-4Pw==/org.openfl.lime.demos.bunnymark-y3ugXL7TXCKIgp4CAJEQOg==/base.apk!classes4.dex] (org.libsdl.app.SDLMain.run+168)
#18 0x000000000036a7e5 /apex/com.android.art/lib64/libart.so (nterp_helper+7157) (BuildId: 44bc52550248fc712a381bef164b75c7)
#19 0x00000000000f6740 /apex/com.android.art/javalib/core-oj.jar (java.lang.Thread.run+8)
#20 0x00000000003725c4 /apex/com.android.art/lib64/libart.so (art_quick_invoke_stub+756) (BuildId: 44bc52550248fc712a381bef164b75c7)
#21 0x00000000003f21a6 /apex/com.android.art/lib64/libart.so (art::ArtMethod::Invoke(art::Thread*, unsigned int*, unsigned int, art::JValue*, char const*)+214) (BuildId: 44bc52550248fc712a381bef164b75c7)
#22 0x00000000007e617e /apex/com.android.art/lib64/libart.so (art::JValue art::InvokeVirtualOrInterfaceWithJValues<art::ArtMethod*>(art::ScopedObjectAccessAlreadyRunnable const&, _jobject*, art::ArtMethod*, jvalue const*)+478) (BuildId: 44bc52550248fc712a381bef164b75c7)
#23 0x000000000083a91c /apex/com.android.art/lib64/libart.so (art::Thread::CreateCallback(void*)+1404) (BuildId: 44bc52550248fc712a381bef164b75c7)
#24 0x00000000000ccd2a /apex/com.android.runtime/lib64/bionic/libc.so (__pthread_start(void*)+58) (BuildId: 007cb2313464df63debf8020e631c990)
#25 0x0000000000060d37 /apex/com.android.runtime/lib64/bionic/libc.so (__start_thread+55) (BuildId: 007cb2313464df63debf8020e631c990)
Crash dump is completed
```

AxGord:
## Summary

Three small defensive fixes in `SDLWindow.cpp` that we've been carrying
in our production fork (https://github.com/soccertutor/lime) for ~1 year+
on native OpenFL apps shipping to macOS, Windows, Android, and iPad:

1. **MSAA fallback on window creation failure.** When `flags` include
   `WINDOW_FLAG_HW_AA` / `WINDOW_FLAG_HW_AA_HIRES` and the initial
   `SDL_CreateWindow` fails, retry once with multisample buffers
   disabled. The Android Studio emulator (AVD) rejects window requests
   with multisampling, even though it supports the OpenGL context
   itself. Today the only signal a user gets is the generic
   "Could not create SDL window" message and a hard crash — the same
   app launches fine on physical Android hardware with MSAA support.
   The fallback turns "no window at all" into "a window without
   antialiasing."

2. **`int width, height;` → `int width = 0, height = 0;`** in
   `GetWidth()` / `GetHeight()`. If `SDL_GetWindowSize` returns
   without writing one of the values (early-out on a torn-down
   window), the previous version returned a stack-uninitialized int.
   Defensive only — observable as occasional one-frame size jumps
   during shutdown.

3. **Null-safe `CreateWindow`.** If the `SDLWindow` constructor
   finishes but `sdlWindow == NULL` (creation failed and we already
   logged), `delete` the half-constructed wrapper and return NULL
   instead of handing back an object that holds a null SDL handle.
   Without this, every subsequent caller has to defensively null-check
   `sdlWindow` on its own — and many don't, which manifests as a
   NULL deref deep inside event dispatch on the next frame.

## Diff

```
project/src/backend/sdl/SDLWindow.cpp | 28 +++++++++++++++++++++++-----
1 file changed, 23 insertions(+), 5 deletions(-)
```

## Test

The MSAA fallback path is the one with the most observable behaviour
change. Tested in production on an Android AVD that previously refused
to launch any Lime app with `<window antialiasing=\"4\"/>` in
`project.xml` — with this patch the app launches successfully with
antialiasing dropped to 0 on the same emulator image. No regression
on physical Android hardware that supports MSAA (the second
`SDL_CreateWindow` is only attempted when the first failed and MSAA
was requested).

The width/height defensive init has no observable behaviour change in
the happy path, only narrows the window during teardown races.

The null-safe `CreateWindow` is dead code on platforms where SDL
window creation never half-succeeds, but on the AVD failure path
described above the original code returned a window object pointing
to NULL — every `SDL_*` call against it then segfaulted. With this
patch the call site sees a clean NULL and can fall back to
`Application::onException` / display an error.

## Risk

- Hardware/driver combos that previously failed window creation with
  MSAA flags now succeed with MSAA off. Apps that depend on receiving
  an MSAA-enabled context (e.g. expecting `glIsEnabled(GL_MULTISAMPLE)`
  to be true) and treat its absence as a fatal error would see
  different behaviour. We've never seen that in practice — apps that
  want MSAA enable it via `<window antialiasing=\"N\"/>` and treat
  "didn't get it" the same as "got it without".
- No platform code or public API touched outside `SDLWindow.cpp`.

### DIFF
--- a/project/src/backend/sdl/SDLWindow.cpp
+++ b/project/src/backend/sdl/SDLWindow.cpp
@@ -160,6 +160,15 @@ namespace lime {
 		}
 		#endif
 
+		if (!sdlWindow && (flags & (WINDOW_FLAG_HW_AA | WINDOW_FLAG_HW_AA_HIRES))) {
+
+			// Retry without antialiasing — emulators and some devices don't support MSAA
+			SDL_GL_SetAttribute (SDL_GL_MULTISAMPLEBUFFERS, 0);
+			SDL_GL_SetAttribute (SDL_GL_MULTISAMPLESAMPLES, 0);
+			sdlWindow = SDL_CreateWindow (title, SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, sdlWindowFlags);
+
+		}
+
 		if (!sdlWindow) {
 
 			printf ("Could not create SDL window: %s.\n", SDL_GetError ());
@@ -584,8 +593,8 @@ namespace lime {
 
 	int SDLWindow::GetHeight () {
 
-		int width;
-		int height;
+		int width = 0;
+		int height = 0;
 
 		SDL_GetWindowSize (sdlWindow, &width, &height);
 
@@ -667,8 +676,8 @@ namespace lime {
 
 	int SDLWindow::GetWidth () {
 
-		int width;
-		int height;
+		int width = 0;
+		int height = 0;
 
 		SDL_GetWindowSize (sdlWindow, &width, &height);
 
@@ -1133,7 +1142,16 @@ namespace lime {
 
 	Window* CreateWindow (Application* application, int width, int height, int flags, const char* title) {
 
-		return new SDLWindow (application, width, height, flags, title);
+		SDLWindow* window = new SDLWindow (application, width, height, flags, title);
+
+		if (!window->sdlWindow) {
+
+			delete window;
+			return NULL;
+
+		}
+
+		return window;
 
 	}
 
--- a/project/src/backend/sdl/SDLWindow.cpp
+++ b/project/src/backend/sdl/SDLWindow.cpp
@@ -806,10 +806,9 @@ namespace lime {
 
 	int SDLWindow::GetHeight () {
 
-		int width = 0;
 		int height = 0;
 
-		SDL_GetWindowSize (sdlWindow, &width, &height);
+		SDL_GetWindowSize (sdlWindow, NULL, &height);
 
 		return height;
 
@@ -890,9 +889,8 @@ namespace lime {
 	int SDLWindow::GetWidth () {
 
 		int width = 0;
-		int height = 0;
 
-		SDL_GetWindowSize (sdlWindow, &width, &height);
+		SDL_GetWindowSize (sdlWindow, &width, NULL);
 
 		return width;
 
--- a/project/src/backend/sdl/SDLWindow.cpp
+++ b/project/src/backend/sdl/SDLWindow.cpp
@@ -162,7 +162,7 @@ namespace lime {
 
 		if (!sdlWindow && (flags & (WINDOW_FLAG_HW_AA | WINDOW_FLAG_HW_AA_HIRES))) {
 
-			// Retry without antialiasing — emulators and some devices don't support MSAA
+			// Retry without antialiasing - emulators and some devices don't support MSAA
 			SDL_GL_SetAttribute (SDL_GL_MULTISAMPLEBUFFERS, 0);
 			SDL_GL_SetAttribute (SDL_GL_MULTISAMPLESAMPLES, 0);
 			sdlWindow = SDL_CreateWindow (title, SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, sdlWindowFlags);
@@ -1147,7 +1147,7 @@ namespace lime {
 		if (!window->sdlWindow) {
 
 			delete window;
-			return NULL;
+			return nullptr;
 
 		}
 
--- a/src/lime/_internal/backend/native/NativeWindow.hx
+++ b/src/lime/_internal/backend/native/NativeWindow.hx
@@ -117,6 +117,7 @@ class NativeWindow
 
 		if (handle == null)
 		{
+			frameRate = resolvedFrameRate;
 			return;
 		}
 
--- a/src/lime/_internal/backend/native/NativeWindow.hx
+++ b/src/lime/_internal/backend/native/NativeWindow.hx
@@ -109,15 +109,17 @@ class NativeWindow
 		#if (!macro && lime_cffi)
 		handle = NativeCFFI.lime_window_create(parent.application.__backend.handle, width, height, flags, title);
 
-		if (handle != null)
+		if (handle == null)
 		{
-			parent.__width = NativeCFFI.lime_window_get_width(handle);
-			parent.__height = NativeCFFI.lime_window_get_height(handle);
-			parent.__x = NativeCFFI.lime_window_get_x(handle);
-			parent.__y = NativeCFFI.lime_window_get_y(handle);
-			parent.__hidden = (Reflect.hasField(attributes, "hidden") && attributes.hidden);
-			parent.id = NativeCFFI.lime_window_get_id(handle);
+			return;
 		}
+
+		parent.__width = NativeCFFI.lime_window_get_width(handle);
+		parent.__height = NativeCFFI.lime_window_get_height(handle);
+		parent.__x = NativeCFFI.lime_window_get_x(handle);
+		parent.__y = NativeCFFI.lime_window_get_y(handle);
+		parent.__hidden = (Reflect.hasField(attributes, "hidden") && attributes.hidden);
+		parent.id = NativeCFFI.lime_window_get_id(handle);
 
 		parent.__scale = NativeCFFI.lime_window_get_scale(handle);
 


project/src/backend/sdl/SDLWindow.cpp
```diff
@@ -160,6 +160,15 @@ namespace lime {
 		}
 		#endif
 
+		if (!sdlWindow && (flags & (WINDOW_FLAG_HW_AA | WINDOW_FLAG_HW_AA_HIRES))) {
+
+			// Retry without antialiasing — emulators and some devices don't support MSAA
```
tobil4sk: Maybe best to remove the emdash as it may cause issues in some editors

project/src/backend/sdl/SDLWindow.cpp
```diff
@@ -1133,7 +1142,16 @@ namespace lime {
...
+		SDLWindow* window = new SDLWindow (application, width, height, flags, title);
+
+		if (!window->sdlWindow) {
+
+			delete window;
+			return NULL;
+
+		}
+
+		return window;
```
tobil4sk, line 1458: We have to make sure all callers of `CreateWindow` check for null with this change. We should also return `nullptr` rather than `NULL` since it is c++ code.

project/src/backend/sdl/SDLWindow.cpp
```diff
@@ -584,8 +593,8 @@ namespace lime {
 
 	int SDLWindow::GetHeight () {
 
-		int width;
-		int height;
+		int width = 0;
+		int height = 0;
```
tobil4sk, line 809: Can SDL_GetWindowSize actually exit without writing these values? Sdl2 docs don't mention that case: https://wiki.libsdl.org/SDL2/SDL_GetWindowSize

project/src/backend/sdl/SDLWindow.cpp
AxGord: Good catch — replaced with an ASCII hyphen in cbc833f7.

project/src/backend/sdl/SDLWindow.cpp
AxGord: Done — `nullptr` in cbc833f7. On the caller side: both C++ callers (`lime_window_create` / `hl_window_create` in `ExternalInterface.cpp`) pass the result straight into `CFFIPointer`/`HLCFFIPointer`, which already null-check (`CFFIPointer.cpp` returns `alloc_null()` for a null pointer; `HLCFFIPointer` returns `0`), so the Haxe side gets a real null. `NativeWindow.create()` *did* guard the width/height/x/y block with `if (handle != null)`, but then called `lime_window_get_scale(handle)` and `lime_window_get_context_type(handle)` unconditionally — those deref a null handle. Fixed in the same commit by replacing the partial guard with an early `return` right after `lime_window_create`.

project/src/backend/sdl/SDLWindow.cpp
AxGord: It can, via `CHECK_WINDOW_MAGIC`. In SDL2 (`src/video/SDL_video.c`) `SDL_GetWindowSize` starts with `CHECK_WINDOW_MAGIC(window, );`, which early-returns *before* writing `*w`/`*h` when the video subsystem is uninitialized, `window` is NULL, or `window->magic` doesn't match (a destroyed/stale window — the teardown race this guards). The wiki doesn't mention it but the source confirms it, so the `= 0` init is defensive against exactly that path; the happy path is unaffected.

tobil4sk: Just to provide the source:
https://github.com/libsdl-org/SDL/blob/f0d1101920864995c0c265128cfd7b96eb959ac5/src/video/SDL_video.c#L2399

https://github.com/libsdl-org/SDL/blob/f0d1101920864995c0c265128cfd7b96eb959ac5/src/video/SDL_video.c#L161

The docs also say:
> NULL can safely be passed as the w or h parameter if the width or height value is not desired.

Which can make the code cleaner and avoid setting the one we don't need to 0.

Also as an aside, it looks like in SDL3 we will be able to check the return value of SDL_GetWindowSize to see if it failed: https://wiki.libsdl.org/SDL3/SDL_GetWindowSize

src/lime/_internal/backend/native/NativeWindow.hx
```diff
@@ -115,17 +115,19 @@ class NativeWindow
 		#if (!macro && lime_cffi)
 		handle = NativeCFFI.lime_window_create(parent.application.__backend.handle, width, height, flags, title);
 
-		if (handle != null)
+		if (handle == null)
 		{
-			parent.__width = NativeCFFI.lime_window_get_width(handle);
+			return;
```
tobil4sk, line 121: It seems this early return causes `frameRate` to not be set. Not sure what the expected state is for a window that failed to launch.

project/src/backend/sdl/SDLWindow.cpp
AxGord: Good call — adopted in 16a34b66. `GetWidth`/`GetHeight` now pass `NULL` for the dimension they don't return and only init the one they do (still `= 0` so the `CHECK_WINDOW_MAGIC` early-return path stays defensive). Thanks for the SDL source links and the SDL3 return-value note.

src/lime/_internal/backend/native/NativeWindow.hx
AxGord: Fixed in c6680070 — `frameRate = resolvedFrameRate;` now runs *before* the early `return`, so a window that failed to launch still has `frameRate` populated for `getFrameRate()`'s field fallback. It's a plain field write (no handle deref) and `resolvedFrameRate` is already computed above, so it's safe on the failure tail. `setTextInputEnabled(false)` is still skipped, but that one only configures SDL state via the (now null) handle and is internally handle-guarded, so there's nothing meaningful to set for a window that never opened.

PR #2064, opened 2026-05-14
merge 8.4.0

dimensionscape:


player-03:
Whoops! Looks like #2005 overwrote some changes from #1980. I'll check if that can be resolved.

PR #2068, opened 2026-05-30
Fix: Haxe 5 compatibility (bump Haxe version from 4.3.6 to 4.3.7 + remove the old hxcpp compile cache when rebuilding Linux ARM ndlls)

MRK2009:
For some reason, It looks like Haxe 5 broke lime because some of the methods got removed from `MainLoop` class (including `progress` from the `Thread` class).
So I had to fix it by using a wrapper of `MainLoop` class and use `loopOnce` to make it work.
I tested it on C++ targets and it looks like it works fine(except it giving me warnings still)
But I'm not sure how to fix the errors for Flash and AIR targets though.


MRK2009:
Maybe I'll rework this pr someday, but rn I'll try make another where I remove the old hxcpp compile cache and bumping the haxe version 4.3.6 to 4.3.7.

player-03:
Thanks for pointing out that this also occurs in _Promise.hx_, I've added that fix to #2007. That's where we're tracking the event loop issue, so check it out and let us know if there's anything it's missing.

MRK2009:
Alright, I'll look into it.

PR #2069, opened 2026-06-01
Remove the old hxcpp compile cache when rebuilding ndlls for Linux ARM + bump Haxe from 4.3.6 to 4.3.7

MRK2009:
I don't think this is necessary to have it for Linux ARM part, while having `HXCPP_COMPILE_CACHE: ${{ github.workspace }}/.hxcpp_cache` as a `env`(For C++ targets) in same time in the workflow, also I updated Haxe version from 4.3.6 to 4.3.7, since it works fine.


tobil4sk (approved):
LGTM, it seems the compile cache step being left in was a merge artefact

MRK2009:
I hope my pr going to merged.

PR #2085, opened 2026-07-14
[tools] Support wine for running windows programs

tobil4sk:
When a windows build is cross compiled from a different host platform, it can be run via the WINE compatibility tool


moxie-coder:
honestly would like to see this

Fancy2209:
This'd be really useful for me as I use Linux but I'm working on an app who needs transparent windows.

PR #2086, opened 2026-07-19
Fix XRandR dynamic library name on Linux

Tutez64:
This fix the case of the XRandR shared library name in the custom Linux SDL configuration.

Linux filenames are case-sensitive, and the `libxrandr2` package provides `libXrandr.so.2`:

https://packages.debian.org/bookworm/amd64/libxrandr2/filelist

SDL previously failed to load XRandR dynamically. As a result, display enumeration could fall back to a generic X11 display.


joshtynjala:
Thanks!

PR #2087, opened 2026-07-22
Fix IOS build for hxcpp cert patch

tobil4sk:
A proposed hxcpp patch (https://github.com/HaxeFoundation/hxcpp/pull/1331) fixes certificate handling on ios, but it requires the Security framework to be linked. This adds that framework to the xcode project file to avoid linker errors with the patched hxcpp.


MRK2009:
Oh, so that's what the Security framework is for huh.

PR #2089, opened 2026-07-31
CI: Update macOS from 14 to 15

MRK2009:
Reason: macOS 14 will be removed from Github in 2nd November 2026.

According to this:
https://github.com/actions/runner-images/issues/13518

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -232,7 +232,7 @@ jobs:
           SDL_AUDIODRIVER=dummy xvfb-run -a Export/linux/bin/LimeVulkanSmoke
 
   macos:
-    runs-on: macos-14
+    runs-on: macos-15
     steps:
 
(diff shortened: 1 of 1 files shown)


MRK2009:
Wait, should I update hxcpp, since there some warings when using 15?
If so, which version from git releases?

joshtynjala:
Thanks! We probably won’t merge this until closer to the November deadline, though. Generally, we wait to update both macOS and Linux on GitHub Actions until as late as possible. Usually, updating the OS version also changes the minimum version that the compiled binary will run on. And our releases are made from CI artifacts. So we prefer to keep the CI version as old as possible to keep it more widely compatible.

MRK2009:
Ah I see huh, in that case I'll just wait for it to be merged.

Also yeah it can be a compatibility issue, if someone tries to build a project using the older os version with compiled binary made from newer one.

PR #2091, opened 2026-08-16
Add fromInt method on doc_gen

Geokureli:
Flixel-doc was getting the following error on [this line](https://github.com/openfl/lime/blob/develop/src/lime/_internal/backend/native/NativeOpenGLRenderContext.hx#L1500):
> 8,3,2/src/lime/_internal/backend/native/NativeOpenGLRenderContext.hx:1500: characters 26-33 : Abstract<lime.graphics.opengl.GLShader> has no field fromInt

I was gonna make an issue and fix later but I saw your new 12 part questionnaire to make a bug and decided I'll probably never ever want to go through that, especially for something like this that takes 2 sentences to properly convey.

### DIFF
--- a/src/lime/graphics/opengl/GLShader.hx
+++ b/src/lime/graphics/opengl/GLShader.hx
@@ -21,7 +21,7 @@ import lime.utils.Log;
 }
 #end
 
-public static function fromSource(gl:Dynamic, source:String, type:Int):GLShader
+public static function fromSource(gl:WebGLRenderContext, source:String, type:Int):GLShader
 {
 	var shader = gl.createShader(type);
 	gl.shaderSource(shader, source);
@@ -55,13 +55,13 @@ public static function fromSource(gl:Dynamic, source:String, type:Int):GLShader
 @:forward abstract GLShader(Dynamic) from Dynamic to Dynamic
 {
 	#if !lime_webgl
-	@:from private static function fromInt(id:Int):GLShader
+	@:noDoc @:from private static function fromInt(id:Int):GLShader
 	{
 		return 0;
 	}
 	#end
 	
-	public static function fromSource(gl:WebGLRenderContext, source:String, type:Int):GLShader
+	public static function fromSource(gl:Dynamic, source:String, type:Int):GLShader
 	{
 		return null;
 	}
--- a/src/lime/graphics/opengl/GLShader.hx
+++ b/src/lime/graphics/opengl/GLShader.hx
@@ -21,7 +21,7 @@ import lime.utils.Log;
 }
 #end
 
-public static function fromSource(gl:WebGLRenderContext, source:String, type:Int):GLShader
+public static function fromSource(gl:Dynamic, source:String, type:Int):GLShader
 {
 	var shader = gl.createShader(type);
 	gl.shaderSource(shader, source);
--- a/src/lime/graphics/opengl/GLShader.hx
+++ b/src/lime/graphics/opengl/GLShader.hx
@@ -54,6 +54,11 @@ public static function fromSource(gl:WebGLRenderContext, source:String, type:Int
 #else
 @:forward abstract GLShader(Dynamic) from Dynamic to Dynamic
 {
+	@:from private static function fromInt(id:Int):GLShader
+	{
+		return 0;
+	}
+	
 	public static function fromSources(gl:Dynamic, source:String, type:Int):GLShader
 	{
 		return null;
--- a/src/lime/graphics/opengl/GLShader.hx
+++ b/src/lime/graphics/opengl/GLShader.hx
@@ -54,12 +54,14 @@ public static function fromSource(gl:WebGLRenderContext, source:String, type:Int
 #else
 @:forward abstract GLShader(Dynamic) from Dynamic to Dynamic
 {
+	#if !lime_webgl
 	@:from private static function fromInt(id:Int):GLShader
 	{
 		return 0;
 	}
+	#end
 	
-	public static function fromSources(gl:Dynamic, source:String, type:Int):GLShader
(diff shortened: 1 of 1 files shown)


src/lime/graphics/opengl/GLShader.hx
```diff
@@ -54,6 +54,11 @@ public static function fromSource(gl:WebGLRenderContext, source:String, type:Int
 #else
 @:forward abstract GLShader(Dynamic) from Dynamic to Dynamic
 {
+	@:from private static function fromInt(id:Int):GLShader
```
tobil4sk: Maybe this should also be guarded with:
```haxe
#if !lime_webgl
```
like the other imlpementation

src/lime/graphics/opengl/GLShader.hx
```diff
@@ -54,6 +54,11 @@ public static function fromSource(gl:WebGLRenderContext, source:String, type:Int
...
+	{
+		return 0;
+	}
+	
 	public static function fromSources(gl:Dynamic, source:String, type:Int):GLShader
```
tobil4sk: While we're here, I think `fromSource` should be `fromSources` so the doc signature matches the real one. It seems it was a typo from c9d522da6bc9103fb1647c94a03bfe91eead3f2b. `GLShader.fromSource` takes one source, unlike `GLProgram.fromSources` which takes two sources.

Geokureli: Change the signature, now CI is failing with
> Error: Incompatibilities between lime.graphics.opengl.GLShader in Android and HTML5 (could not merge definition)

should I change it back to dynamic or something?

tobil4sk: > should I change it back to dynamic or something?

Yeah, that could be the reason

tobil4sk: Hm, I think there was probably a reason it is Dynamic anyway, so it should probably be kept like that. However, it seems that's not why it was failing, and on second thought it's probably adding `#if !lime_webgl` that introduced this error. Maybe we actually want to mark fromInt as noDoc or something.

Geokureli: Can you be more specific? I'm not familiar with noDoc, I assume it's a meta? would I add it to the new method, or both

tobil4sk: Yes, we can add `@:noDoc` just to the stub fromInt implementation.

src/lime/graphics/opengl/GLShader.hx
```diff
@@ -21,7 +21,7 @@ import lime.utils.Log;
 }
 #end
 
-public static function fromSource(gl:WebGLRenderContext, source:String, type:Int):GLShader
+public static function fromSource(gl:Dynamic, source:String, type:Int):GLShader
```
tobil4sk: I think the wrong one got reverted, only the stub should have Dynamic here

Geokureli: whoops, sorry I'm pretty scatterbrained today

src/lime/graphics/opengl/GLShader.hx
tobil4sk: I think ci is still failing with `@:noDoc` because it is running haxe 4.2.5, which is missing this fix: https://github.com/HaxeFoundation/haxe/commit/0d62f64e232ca7aabb498d57dcd0164e28e16093

So maybe we can try `@:dox(hide)` instead for better compatibility.

EDIT: I just tried it locally, and it isn't sufficient to stop the error. Maybe the docs will just require haxe 4.3+ to generate.

Geokureli: If fromInt is the issue then why did the first commit succeed? Are we sure the issue isn't renaming fromSources?

tobil4sk: The first commit succeeded because it didn't have `#if !lime_webgl` yet. Now the webgl targets can't be merged with the other targets, because they don't agree about whether the fromInt method should exist.
`@:noDoc` is meant to exclude it from being in the doc output, but that bug in haxe 4.2 stops it from working on `@:from` methods.

One way to work around it is to just use haxe 4.3 for the doc generation job.

Geokureli: Could also wrap this in ```#if haxe > version("4.3")```, I suppose. I don't get why you would generate docs in old haxe versions, openfl.org doesn't seem to show apis for older versions

tobil4sk: I think adding that check around fromInt would just bring back the original error: `has no field fromInt`

To get the ci passing, we can change the haxe version here:
https://github.com/Geokureli/lime/blob/f8ba43823ed773c04f88b12603c79edbd739cefa/.github/workflows/main.yml#L733

Maybe we can just update the global version to haxe 4.3, unless there was a reason lime is sticking to 4.2 for now. Perhaps @joshtynjala might know?

Geokureli: > I think adding that check around fromInt would just bring back the original error: `has no field fromInt`

It would fix my issue with flixel-docs and anyone else using haxe 4.3. but yeah, let's find out

joshtynjala: I don't recall any particular reason for us to strictly stick to Haxe 4.2 instead of 4.3 for our default on CI. I think it's been working fine, so we haven't bothered updating. 4.3 has been out for a while now, and is also very stable, so it shouldn't cause issues.

tobil4sk: In that case I think we can change the global haxe version to 4.3.7 in this PR, which should make the docs ci job pass here.

PR #2092, opened 2026-08-23
Stop forcing StageAlign.TOP_LEFT on Flash/AIR

Tutez64:
Flash/AIR already default `Stage.align` to `""` (centered). Haxe's `flash.Boot` overwrites that with `TOP_LEFT` when the value is empty, and Lime then forced `TOP_LEFT` again in `FlashWindow`.
Restore `""` after boot so AIR matches Flash Player.


joshtynjala:
I've always assumed that this was chosen purposefully for consistency with other targets, which default to `TOP_LEFT` and `NO_SCALE` (and the other targets didn't even support other `align`/`scaleMode` values until recently).

PR #2096, opened 2026-09-04
Android Gradle 9 support

IriySoft:
This PR allows using Gradle version 9+ and changes the way the .apk and .aab files are renamed during the build.

**Details:**

- `classpath 'org.apache.groovy:groovy-swing:+'` added to top-level `build.gradle` because it's not included by default in Gradle 9+
- for the builds that use Gradle versions lower than 9 the renaming code is replaced with `setProperty("archivesBaseName", "::APP_FILE::")` that also enables renaming .aab in addition to .apk in the app-level `build.gradle`
- for the builds that use Gradle 9+ the renaming is done with the following code in the app-level `build.gradle`
```
base {
	archivesName = "::APP_FILE::"
}
```

**Warning!**
- When I switched to Gradle 9+, I had to increase Java heap size with `org.gradle.jvmargs=-Xmx2048m` in `gradle.properties`, otherwise the build started to crash with "Out of memory" exception. The reason can be in the Appodeal SDK though that includes dozens of packages.
- Additionally I had to rebuild Lime / Android, before that the apps crashed on startup, but most probably it's because I'm using the dev branch and didn't rebuild it for quite some time...

### DIFF
--- a/templates/android/template/app/build.gradle
+++ b/templates/android/template/app/build.gradle
@@ -13,6 +13,10 @@ System.setProperty('java.awt.headless','false')
 	options.compilerArgs << "-Xlint:unchecked" << "-Xlint:deprecation"
 } */
 
+::if (ANDROID_GRADLE_PLUGIN>="9.0")::base {
+	archivesName = "::APP_FILE::"
+}::end::
+
 android {
 	namespace "::APP_PACKAGE::"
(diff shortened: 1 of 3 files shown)


MRK2009:
OH YES PLEASE!!!

PR #2101, opened 2026-09-19
Apply extraParams.hxml for path-overridden haxelibs

Tutez64:
`haxelib path` prepends extraParams.hxml. A `<haxelib path="..."/>` override never calls it, so --macro and other flags from that file were dropped.
The file is read from the library root, before classPath is appended.
Repro: [lime-path-haxelib-extraparams.zip](https://github.com/user-attachments/files/32416654/lime-path-haxelib-extraparams.zip)


player-03:
Duplicate of #2010, and some discussion already happened there. How about you submit your additions to [that branch](https://github.com/player-03/lime/tree/Haxelib_pathOverrides_extraParams) so we can keep everything in one place?

Tutez64:
Thanks — I missed #2010. Closing this as a duplicate.

The extra pieces versus your branch are: trim each line, skip `-L`, and treat `-D name` (no `=`) the same way the `haxelib path` branch already does. Sent against your branch: https://github.com/player-03/lime/pull/10

Tutez64:
Duplicate of #2010.

